PR #6, opened 2020-10-16
Modernize autoupdate

didrocks:
- Use create-pull-request v3
- Change env variable handling due to Github change
- Create pot bootstrap


didrocks:
Everything passed, let’s merge it now,

PR #46, opened 2021-01-20
DE-8 - Cache GOCACHE content for multiple runs

didrocks:
This will speed up builds, tests and other CI checks by reusing and
single cache per job.
Note that the main cache is shared with new branches on PR. Then, the
local branch case is preferred.

Co-authored-by: Jean-Baptiste Lallement <jean-baptiste@ubuntu.com>


didrocks:
This is not going to work, due to the cache being per worflow, let’s discare it for now.

PR #50, opened 2021-01-27
Convert policy

jibel:
Handle dconf settings as Windows policies instead preferences. It means that settings are managed a bit differently on Windows' side, in particular they are removed from the registry.pol files when the key is not configured and settings are not persistent anymore.


didrocks:
Looks good!

PR #60, opened 2021-02-10
Offline mode support

didrocks:
Reload the gpo list and rules from cache if the first ad connection fails
(error 2) and that we have a local cache available.
Those are the rules which will be enforced for the object.
Add tests.


jibel (approved):
+1. Thanks.

PR #65, opened 2021-02-15
CI: Only set go version once

didrocks:
Do not use matrix job as we are only running one version of go.


didrocks:
This runs with Go 1.15, let’s merge.

PR #66, opened 2021-02-16
Run CI to generate admx and adml files automatically

didrocks:
Multiple jobs are ran to collect and merged all policy rules to generate
admx/adml files:
- statically build a single binary without cgo
- run multiple containers in parallel, one for each release, and collect
  supported keys on all of them (dconf, apparmor, scripts…)
- collect all of that in a single job and merge them to generate
  admx/adml files.


didrocks:
Let’s merge the PR manually

PR #75, opened 2021-02-23
Embed admx and adml for multiple flavors and retrieve it thanks to the admx command

didrocks:


didrocks:
Let’s merge the new command before FF

PR #82, opened 2021-02-24
Reorganize policies related command under a single subcommand

didrocks:
Create aliases for shortcuts on some of them (applied and update), but
put them all under the single "policy" command category.


didrocks:
Shell completion works as expected with the new reorg.

PR #93, opened 2021-03-04
Renamed meta:default to meta:empty for clarity.

jibel:
Co-authored-by: Didier Roche <didrocks@ubuntu.com>


didrocks:
Good riddance, confusing "default"

PR #121, opened 2021-04-24
Replace warthogs.biz domain by example.com

upils:
`warthogs.biz` is the example domain used in the configuration example as long as test suite. I suspect developers and users of adsys may use the configuration example "as is" and try to connect to this domain to test the tool. To prevent a malicious actor to setup an Active Directory listening on this domain and exposing malicious GPO, I propose to replace with `example.com` (or any other neutral domain if you prefer).

I have made the changes but as the integration tests seems to use an private private ubuntu image I cannot pass them. I hope the CI here will execute these tests.

I have done a dumb search-and-replace, maybe we should do more.

In the meantime I have bought the domain to prevent someone else to do it.

Signed-off-by: Paul Mars <upils@eodras.com>


internal/testutils/admock/ldb/__init__.py
```diff
@@ -15,58 +15,58 @@ def binary_encode(s):
 accounts = {}
 
 ##############################
-# OU=RnD,OU=IT Dept,DC=warthogs,DC=biz
+# OU=RnD,OU=IT Dept,DC=example,DC=biz
```
didrocks, line 18: This should be DC=com, (please check if there is another I missed)

internal/testutils/admock/ldb/__init__.py
```diff
@@ -175,123 +175,123 @@ def getuserWithoutDomain():
...
 o.addAccount("UserIntegrationTest")
 
-o = OU("/warthogs/IntegrationTests/UserDep/Dep1")
+o = OU("/example/IntegrationTests/UserDep/Dep1")
 o.addGPO(GPO("{5EC4DF8F-FF4E-41DE-846B-52AA6FFAF242}", display_name="GPO1 for current User"))
 o.addGPO(GPO("{073AA7FC-5C1A-4A12-9AFC-42EC9C5CAF04}", display_name="GPO2 for current User"))
 o.addAccount(getuserWithoutDomain())
 
-# [b'[LDAP://cn={83A5BD5B-1D5D-472D-827F-DE0E6F714300},cn=policies,cn=system,DC=warthogs,DC=biz;0][LDAP://cn={5EC4DF8F-FF4E-41DE-846B-52AA6FFAF242},cn=policies,cn=system,DC=warthogs,DC=biz;0]'
+# [b'[LDAP://cn={83A5BD5B-1D5D-472D-827F-DE0E6F714300},cn=policies,cn=system,DC=example,DC=biz;0][LDAP://cn={5EC4DF8F-FF4E-41DE-846B-52AA6FFAF242},cn=policies,cn=system,DC=example,DC=biz;0]'
```
didrocks, line 276: Same here (and below)

didrocks (changes requested):
Thanks for the suggestion! This is indeed relevant and a good move.
Some small changes are needed, but overall good.

The test docker container for integration tests should be available on github though? It’s here: https://github.com/ubuntu/adsys/packages/712875.

I don’t find any option to make it more "public", but it seems github requests to be logged in on docker. If you have any hint or can’t still access it, tell us so that we can try to figure it out.

(if you want to build it locally, the docker file is in this folder: https://github.com/ubuntu/adsys/tree/main/cmd/adsysd/integration_tests/polkitd and you can run the build script).

I’ll let the CI run as a first pass even and we can see the first set of results.

didrocks:
Two tests tests are failing, like `can't get policies for "fv-az180-137": failed to retrieve the list of GPO: exit status 1`.
This is probably some replacements missing.
(**TestGetPolicies/Standard_policy,_computer_object**, **TestGetPolicies/Computer_ignored_CCBaseName**). Do you mind having a look?

You can ignore the following warning (we will fix it in a separate PR) as this is only for tests and will only be a blocker in a future protobuf release.
```
            	            	2021/04/26 07:39:11 WARNING: proto: file "log_test.proto" has a name conflict over Empty
            	            		previously from: "github.com/ubuntu/adsys"
            	            		currently from:  "github.com/ubuntu/adsys/internal/grpc/logstreamer/test"
            	            	A future release will panic on registration conflicts. See:
            	            	https://developers.google.com/protocol-buffers/docs/reference/go/faq#namespace-conflict
```

didrocks:
Evertyhing looks good from my POV, thanks a lot for working on that, great work! :) (and tests pass)

didrocks (approved):

upils:
Thanks ! This is a nice project. I will probably keep digging on it. Do not hesitate to ping me if you need some help. Not sure I would be available but maybe I could help.

PR #204, opened 2021-09-08
Test impish checkout

didrocks:
This should fix some packaging breakage not declared explicitely.


didrocks:
let’s remove that one and do another one with the real test fix

PR #206, opened 2021-09-14
TestServiceStatus: changed settings path

didrocks:
The behaviour of Test.Tempdir changed in go 1.17 and now include the
name of the test and the name of the subtest. As a consequence the
golden files are different between 1.16 and 1.17 and TestServiceStatus
cannot run on both. We now strip entirely the name of the test from the
output file.

Co-authored-by: Jean-Baptiste Lallement <jean-baptiste@ubuntu.com>


jibel (approved):
LGTM.

PR #210, opened 2021-09-15
Enable golangci-lint in CI and fix linted issues

didrocks:


didrocks:
Only the admx/adml generation failing due to libc6. The rest pass with all the code cleanup! Merging.

PR #214, opened 2021-09-27
Workaround CI adm generation failing on impish

didrocks:
ibus can’t be setup in the container, let’s ignore install failure.
We will review the PR anyway if anything is generated and it not
expected.


didrocks:
We can’t run an impish container on a non impish host due to libc. There is no workaround available as the fix needs to be on the host:
- https://github.com/moby/moby/pull/42681
- https://bugs.launchpad.net/cloud-images/+bug/1943049

PR #223, opened 2021-11-08
Use native grpc unix socket handler

didrocks:
grpc can now use unix socket directly. Remove our context dialer wrapper
and switch to it.


jibel (approved):
LGTM

PR #285, opened 2022-02-21
pam_adsys: Fix memory leak and identation

didrocks:
Both minors issues reported by Seth Arnold. Thanks!
https://bugs.launchpad.net/ubuntu/+source/adsys/+bug/1961459


jibel (approved):
lgtm

PR #287, opened 2022-02-25
Fix chown scripts

didrocks:
Fix potential security race with user/gid reset

Only chown to uid and gid when creating directories and any fd before closing we are in control of.
This impacted the API for passing desired uid and gid.
Remove the final recursive reset.
Rework chowning when needed. Set -1 to be API compliant for no change.


jibel (approved):
Great fix. Thanks!

setharnold:
Nice fix indeed :) thanks!

PR #289, opened 2022-02-27
internal/testutils: invoke smbd with option accepted by 4.15+

mwhudson:
Because obviously an "Improved command line user experience" (from the
Samba 4.15.0 release notes) means incompatible changes, adsys currently
ftbfs becaus smbd no longer takes -S / --log-stdout as an option, it's
now --debug-stdout.

### DIFF
--- a/internal/testutils/samba.go
+++ b/internal/testutils/samba.go
@@ -18,7 +18,7 @@ func SetupSmb(port int, sysvolDir string) func() {
 	dir, cleanup := mkSmbDirWithConf(smbPort, sysvolDir)
 
 	// #nosec:G204 - we control the directory we run smbd on (on tests)
-	cmd := exec.Command("smbd", "-FS", "-s", filepath.Join(dir, "smbd.conf"))
+	cmd := exec.Command("smbd", "-F", "--debug-stdout", "-s", filepath.Join(dir, "smbd.conf"))
 	stderr, err := cmd.StderrPipe()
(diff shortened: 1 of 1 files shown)


mwhudson:
If you want to backport new adsys versions to LTS releases then this fix isn't the right thing to do of course. But it's blocking stuff in proposed so doing *something* fairly quickly is quite important.

(as an aside, it would be neat if failing because of the waitForPortReady(smbPort) in SetupSmb failing, the stderr of the smbd process got dumped out somehow)

mwhudson:
FWIW I've uploaded this change to jammy to get some transitions unstuck.

didrocks:
The issue is that GitHub CI will continue running the old version. Do you mind making the patch a little bit more general and have a conditional command line, based on smbd version, if such a switch exists?

mwhudson:
Something like this? My Go is a bit rusty sadly

didrocks:
Thanks! I will probably rewrite it a little bit to capture stderr on our and simplify one or two things, but meanwhile can you either allow me to push to this branch or fix the CI linting?

mwhudson:
"allow edits by maintainers" is already checked, not sure what that issue is. But please, take this code and make it nice however you want, no need to merge the stuff I've produced.

didrocks:
I slightly rewrote it, tests pass and it’s backward compatible, Thanks a lot Michael!

PR #292, opened 2022-03-07
Try reducing race conditions when testing configuration change

didrocks:
Ensure the files are committed on disk for the refresh to be taken into
account.
This is in hope that on VM tests, we get less time-sensitive failures.


jibel (approved):
LTGM. Sync() should be enough to garantee that files are written to disk on Linux. If it is not successful then we'll swtich to fsync().

PR #294, opened 2022-03-08
Fix scripts refresh without user session

didrocks:
Only update scripts user directory when session is not running

Use a separate flag file that we add and remove depending on user being in session.
The issue on only depending on the ready flag and removing the whole scripts directory is the following:
- user enters a session, the scripts/ user directory is done and ready flag file is done.
- when exists the session, the scripts/ user directory is removed
- however, next refresh will list all active users based on krb5 file presence, and will prepare again the scripts/ user directory, with ready flag.
- any following updates will thus not refresh the scripts/ user directory as we have a directory existing and flagged, and we don’t know if the user entered the session. Only logon and then logoff of the user will then allow to refresh our scripts/ user directory.


jibel (approved):
Looks good. Thanks!

PR #295, opened 2022-03-08
Workaround tests hanging on ppcel64

didrocks:
When killing smbd process, on ppcel64, stderr was not closed, causing
the tests to hang forever.


jibel (approved):
nice catch!

PR #297, opened 2022-03-09
Try fixing flaky "pick up config changes" tests

didrocks:
This test is still failing randomly in the archive or in autopkgtests on
armhf and arm64.
Change thus the approach to define phases:
- First phase is initial configuration set
- Second optional phase is a configuration change for some tests
- Third phase is the "done" one.

This way, we control the amount of callbacks we are interested in and
picking, discaring the unwanted multiple inotify event for the same
write on disk.
It seems to make the flaky failure (which only impacts tests, as in a
real system, we are interested even in no-op changes) go away on arm64
after multiple attempts.

### DIFF
--- a/internal/config/config_test.go
+++ b/internal/config/config_test.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"sync"
 	"syscall"
 	"testing"
 	"time"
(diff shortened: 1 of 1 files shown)


jibel (approved):
Sounds like the right fix. Thanks!

PR #308, opened 2022-04-04
Rename markdown wiki files so it follows a numeric sorting

Adrast:
The filenames of the wiki markdown files are not properly **_sortable_** as seen in the picture below:

image

I've renamed the files (1.-welcome -> 01--welcome) to be **_sortable_**.


didrocks:
Hey, thanks for contributing to ADSys, do you mind signing the CLA and tell me once done so that I can rerun the bot?
Also, I like the "." syntax if we can keep it if possible. So something like:
`01.-` would be possible?

Thanks again!

didrocks:
Thanks for signing the CLA! I have adapted the integration test on documentation which was using the second chapter to use 02. instead of 2.. With this, all tests pass. Merging.

Thanks for your contribution :)

PR #330, opened 2022-05-13
Align setting verbosity between admxgen and adsys*

didrocks:
We always set the verbosity the same way between the two binaries. This
ensure consistency.


jibel:
+1

PR #345, opened 2022-06-07
Fix build-failure on 32 bits on kinetic

didrocks:
Fix new build failures on 32 bits due to libsmbclient-dev no longer sets
the large file support cflags in libsmbclient.h.
Update to latest libsmbclient-go.


jibel (approved):
All good. Thanks!

PR #348, opened 2022-06-15
Add AD Watch Daemon

GabrielNagy:
This pull request integrates the Active Directory Watch Daemon in the adsys project.


internal/watchdhelpers/watchdhelpers.go
```diff
@@ -0,0 +1,102 @@
...
+	// Make sure all directories exist
+	for _, dir := range dirs {
+		if _, err := os.Stat(dir); errors.Is(err, os.ErrNotExist) {
+			return fmt.Errorf(i18n.G("directory %q does not exist"), dir)
+		}
+	}
+
+	// Empty input means using the default config file
+	if confFile == "" {
+		confFile = defaultConfFile
```
didrocks, line 65: I would simplify here and do that on the caller side. Basically the API is "write it there"

internal/watchdhelpers/watchdhelpers.go
```diff
@@ -0,0 +1,102 @@
...
+// GetConfigFileFromArgs returns the path to the config file extracted from the
+// command line arguments.
+//
+// This is not an exhaustive implementation of "parsing" the command line and
+// only covers the cases used by the service installer, which should be good
+// enough for us.
+func GetConfigFileFromArgs(args string) (string, error) {
+	_, configFile, found := strings.Cut(args, "-c")
+	if !found {
+		return "", fmt.Errorf(i18n.G("missing config file in CLI arguments"))
```
didrocks, line 95: Couldn’t there be the case for the service not running with an argument (for whatever reason), and in that case, it’s using the default config path happily? Shouldn’t we return that one in that case?

didrocks (commented):
I haven’t review more before the rest is merged so that we can rebase, but first food for thoughts :)

.github/workflows/qa.yaml
```diff
@@ -52,8 +54,9 @@ jobs:
         uses: actions/setup-go@v2
         with:
           go-version: ${{ env.goversion }}
-      # We add a new user and execute under it for file permissions to be taken into account
-      # as root can write in RO directories.
+      # Workaround for https://github.com/actions/setup-go/issues/107
+      - name: Replace default go symlink
+        run: cp -f `which go` /usr/bin/go
```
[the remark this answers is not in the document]
didrocks: waow, this is a big issue IMHO. I think copying works because GOROOT is specified in some way by the action, but I really wonder if we will need to put that everywhere (why isn’t it failing in go vet and other checks?)

internal/watchdhelpers/watchdhelpers.go
```diff
@@ -0,0 +1,102 @@
+package watchdhelpers
```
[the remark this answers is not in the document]
didrocks: basically, the package name is watchd then? (most of the time, they match the directory name, it’s not a requirement, but always puzzling when using it if not).

Then, in your import in cmd/ and the service, you can rename it like `watchdconfig`?

internal/watchdhelpers/watchdhelpers.go
didrocks: yeah, but can’t an admin of the machine mess with the parameters in some Windows service manager UI?

didrocks:
I guess it’s now time for a rebase? (I’m always bothered by the fact that github doesn’t autorefresh PR on destination branch changing)

doc/12.-Active-Directory-Watch-Daemon.md
```diff
@@ -0,0 +1,119 @@
+# Active Directory Watch Daemon
```
didrocks, line 1: I would move the Command line reference at the end, and so, this one is 11, the reference is 12.

installer/setup.iss
```diff
@@ -0,0 +1,56 @@
...
+WizardStyle=modern
+DefaultDirName={autopf}\Ubuntu\AD Watch Daemon
+; Since no icons will be created in "{group}", we don't need the wizard
+; to ask for a Start Menu folder name:
+DisableProgramGroupPage=yes
+UninstallDisplayIcon={app}\icon.ico
+Compression=lzma2
+SolidCompression=yes
+WizardImageFile=assets\Ubuntu-Symbol.bmp
+WizardSmallImageFile=assets\Ubuntu-Symbol-small.bmp
```
didrocks, line 26: bitmaps, seriously? No other format is accepted?

installer/setup.iss
```diff
@@ -0,0 +1,56 @@
...
+AppPublisher=Ubuntu
+AppPublisherURL=https://www.ubuntu.com/
+AppCopyright=Copyright (C) 2022 Canonical Ltd.
+WizardStyle=modern
+DefaultDirName={autopf}\Ubuntu\AD Watch Daemon
+; Since no icons will be created in "{group}", we don't need the wizard
+; to ask for a Start Menu folder name:
+DisableProgramGroupPage=yes
+UninstallDisplayIcon={app}\icon.ico
+Compression=lzma2
```
didrocks, line 23: …bitmap but you have lzma compression :)

installer/setup.iss
```diff
@@ -0,0 +1,56 @@
...
+SolidCompression=yes
+WizardImageFile=assets\Ubuntu-Symbol.bmp
+WizardSmallImageFile=assets\Ubuntu-Symbol-small.bmp
+DisableWelcomePage=no
+; SignTool=MsSign $f
+OutputBaseFilename=adwatchd_setup
+
+[Languages]
+Name: en; MessagesFile: "compiler:Default.isl"
+Name: fr; MessagesFile: "compiler:Languages\French.isl"
```
didrocks, line 33: you know what matters ;)

didrocks (commented):
Ok, I will already post this for now. I am far from done (20/99 files viewed :p), but we’ll get there!

internal/loghooks/eventlog_test.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+	for name, tc := range tests {
+		tc := tc
+		t.Run(name, func(t *testing.T) {
+			buf.Reset()
+			log := logrus.New()
+			log.AddHook(&loghooks.EventLog{mockServiceLogger{}})
+			log.SetLevel(tc.level)
+
+			log.Debug("Debug msg")
```
didrocks, line 41: Shouldn’t that be `msgs["debug"]`?

didrocks (commented):
33/99! I think I will continue on Monday now :)

internal/config/watchd/watchd_test.go
```diff
@@ -0,0 +1,209 @@
...
+	tests := map[string]struct {
+		args string
+
+		want    string
+		wantErr bool
+	}{
+		"empty args":                        {wantErr: true},
+		"no config argument":                {args: "adwatchd.exe", wantErr: true},
+		"short config argument":             {args: `adwatchd.exe -c C:\path\to\adwatchd.yml`, want: `C:\path\to\adwatchd.yml`},
+		"short config argument with quotes": {args: `adwatchd.exe -c "C:\path\to\adwatchd.yml"`, want: `C:\path\to\adwatchd.yml`},
```
didrocks, line 27: Suggestion: try `-c` with no path (to protect us against bugs in the service manager) :)

internal/config/watchd/watchd_test.go
```diff
@@ -0,0 +1,209 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	watchdconfig "github.com/ubuntu/adsys/internal/config/watchd"
+	"gopkg.in/yaml.v2"
+)
+
+var update bool
+
+func TestGetConfigFileFromArgs(t *testing.T) {
```
didrocks, line 17: t.Parallel()

internal/config/watchd/watchd_test.go
```diff
@@ -0,0 +1,209 @@
...
+func TestFilterAbsentDirs(t *testing.T) {
+	tests := map[string]struct {
+		inputDirs    []string
+		existingDirs []string
+
+		wantDirs []string
+	}{
+		"no existing dirs":   {inputDirs: []string{"dir1", "dir2"}},
+		"some existing dirs": {inputDirs: []string{"dir1", "dir2"}, existingDirs: []string{"dir1"}, wantDirs: []string{"dir1"}},
+		"all existing dirs":  {inputDirs: []string{"dir1", "dir2"}, existingDirs: []string{"dir1", "dir2"}, wantDirs: []string{"dir1", "dir2"}},
```
didrocks, line 108: In your code, you check that the dest is a dir. I suggest that you have a case where it’s a file.

internal/config/watchd/watchd_test.go
```diff
@@ -0,0 +1,209 @@
...
+
+			var data []byte
+
+			if !tc.absentDirs {
+				for _, dir := range tc.dirs {
+					require.NoError(t, os.MkdirAll(dir, 0750), "failed to create dirs")
+				}
+			}
+
+			if update {
```
didrocks, line 157: Updating the golden files should be just before the comparison at the end, after running the tests and the golden should be the consequence of writing the config here, not preparing manually by another mean.

In some other words, the golden is the result of the code under tests as it’s what you are comparing to.

internal/config/watchd/watchd_test.go
```diff
@@ -0,0 +1,209 @@
...
+			}
+
+			configPath := "adwatchd.yml"
+			if tc.nestedConfigPath {
+				configPath = filepath.Join("path", "to", "adwatchd.yml")
+			}
+
+			error := watchdconfig.WriteConfig(configPath, tc.dirs)
+			if tc.wantErr {
+				require.Error(t, error, "expected error")
```
didrocks, line 177: return + remove the else, align left, as we won’t compare golden here.

didrocks (commented):
The next batch :p

internal/watchdservice/watchdservice.go
```diff
@@ -0,0 +1,373 @@
...
+
+	if err := s.waitForStatus(ctx, service.StatusRunning); err != nil {
+		// Try to start it (not all platforms try to start it after installing)
+		return s.service.Start()
+	}
+
+	return nil
+}
+
+// Uninstall uninstalls the watcher service. If the service is not installed it
```
didrocks, line 344: The comment would be more readable like this IMHO:
```go
// Uninstall uninstalls the watcher service.
// If the service is not installed, it logs a message and returns.
// If the service is running, it attempts to stop it first.
```

internal/loghooks/eventlog_test.go
didrocks: No, this is fine with a comment IMHO on the definition above :)

didrocks (commented):
Next batch with some answers for above

internal/watcher/watcher.go
```diff
@@ -0,0 +1,387 @@
...
+	"github.com/fsnotify/fsnotify"
+	"github.com/kardianos/service"
+	"github.com/ubuntu/adsys/internal/decorate"
+	log "github.com/ubuntu/adsys/internal/grpc/logstreamer"
+	"github.com/ubuntu/adsys/internal/i18n"
+	"gopkg.in/ini.v1"
+)
+
+const (
+	gptFileName = "gpt.ini"
```
didrocks, line 22: Do you mind checking the case of this file on a real AD? I’m still pretty sure this was `GPT.ini`.

internal/watcher/watcher_test.go
```diff
@@ -0,0 +1,446 @@
...
+						testutils.Copy(t, filepath.Join("testdata", rootDir), filepath.Join(temp, rootDir))
+					}
+				}
+				dirs = append(dirs, filepath.Join(temp, dir))
+			}
+
+			// Instantiate the object
+			w, err := watcher.New(context.Background(), dirs)
+			if tc.wantErrNew {
+				require.Error(t, err, "Create should have failed but hasn't")
```
didrocks, line 129: It’s named `New()` :)

internal/watcher/watcher_test.go
```diff
@@ -0,0 +1,446 @@
...
+			err = w.Stop(mockService{})
+			require.NoError(t, err, "Can't stop watcher")
+
+			// compare GPT.ini version
+			testutils.WaitForWrites(t)
+
+			if len(tc.wantVersions) > 0 {
+				for i, dir := range tc.existingDirs {
+					if tc.wantErrBump {
+						assertGPTVersionError(t, filepath.Join(temp, dir))
```
didrocks, line 196: continue + remove of the else, align left

internal/watcher/watcher_test.go
```diff
@@ -0,0 +1,446 @@
...
+	require.Error(t, err, "Stop should have failed but hasn't")
+}
+
+func assertGPTVersionEquals(t *testing.T, path string, version int) {
+	t.Helper()
+
+	testutils.WaitForWrites(t)
+	gptfile := filepath.Join(path, "gpt.ini")
+
+	// Wait for GPT.ini to be created
```
didrocks, line 374: I think the comment doesn’t match, was that for `WaitForWrites` above?

internal/watcher/watcher_test.go
```diff
@@ -0,0 +1,446 @@
...
+	require.True(t, gptFileExists, "GPT.ini not created")
+
+	cfg, err := ini.Load(gptfile)
+	require.NoError(t, err, "Can't load GPT.ini")
+
+	_, err = cfg.Section("General").Key("Version").Int()
+	require.Error(t, err, "Version should be invalid")
+}
+
+func writeToFiles(t *testing.T, files []string) {
```
didrocks, line 415: maybe a better name would `updateFiles()`

(Almost tempted to name it `updateFilesWithNEWCONTENT` :p)

internal/watcher/watcher_test.go
```diff
@@ -0,0 +1,446 @@
...
+}
+func (mockService) SystemLogger(errs chan<- error) (service.Logger, error) {
+	return service.ConsoleLogger, nil
+}
+func (mockService) String() string                  { return "" }
+func (mockService) Platform() string                { return "" }
+func (mockService) Status() (service.Status, error) { return 0, nil }
+
+func TestMain(m *testing.M) {
+	config.SetVerboseMode(2)
```
didrocks, line 444: is that on purpose?

didrocks (commented):
ok, it seems I know just have watchdtui to review \o/

didrocks (changes requested):
Excellent work! In particular on the tui front! Publishing the latest comments (phew!) so that the first pass is fully done :)

Some comments are really copy/paste on multiple places to not forget to changes all needed instances.

cmd/adwatchd/commands/main.go
```diff
@@ -0,0 +1,226 @@
...
+
+			// If we have a config file, pass it as an argument to the service.
+			var configFile []string
+			if len(a.viper.ConfigFileUsed()) > 0 {
+				absPath, err := filepath.Abs(a.viper.ConfigFileUsed())
+				if err != nil {
+					close(a.ready)
+					return err
+				}
+				configFile = []string{"-c", absPath}
```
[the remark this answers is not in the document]
didrocks: good point, I agree with you :)

cmd/adwatchd/integration_tests/adwatchd_run_test.go
```diff
@@ -0,0 +1,158 @@
...
+
+	// Give time for the watcher itself to start
+	time.Sleep(time.Millisecond * 100)
+
+	err = app.Quit(syscall.SIGTERM)
+	require.NoError(t, err, "Quitting should succeed")
+	select {
+	case <-done:
+	case <-time.After(1 * time.Second):
+		// TODO: fix quitting on windows
```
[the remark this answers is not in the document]
didrocks: Rather a FIXME then? Ok, this test doesn’t explicitly rely on quitting behavior. I suggest that we keep t.Fatal() conditioned to Linux?

cmd/adwatchd/integration_tests/adwatchd_service_test.go
```diff
@@ -0,0 +1,196 @@
...
+		"uninstall and restart":    {sequence: []string{"uninstall", "stop"}, wantErrAt: []int{1}},
+	}
+	for name, tc := range tests {
+		tc := tc
+		name := name
+		t.Run(name, func(t *testing.T) {
+			// Parallelization is not supported on Windows due to Service
+			// Control Manager reasons.
+			if runtime.GOOS != "windows" {
+				t.Parallel()
```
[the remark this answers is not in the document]
didrocks: you are correct. We only call getStatus very few times this is why there is no race until we have 2 `wantStopped` tests. So yes, you need to remove the parallel part.

cmd/adwatchd/integration_tests/adwatchd_service_test.go
```diff
@@ -0,0 +1,196 @@
...
+	t.Cleanup(func() {
+		uninstallService(t, configPath, app)
+	})
+
+	installService(t, configPath, app)
+
+	// Wait for service to be running
+	time.Sleep(time.Second)
+
+	want := fmt.Sprintf(`Service status: running
```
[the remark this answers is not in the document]
didrocks: Ack then!

internal/watcher/watcher.go
didrocks: True, but as we generate those files, it’s better to be in line with the case Microsoft is using when generating GPOs

internal/watcher/watcher.go
```diff
@@ -0,0 +1,387 @@
...
+// bumpVersion does the actual bumping of the version in the given GPT.ini file.
+func bumpVersion(ctx context.Context, path string) (err error) {
+	decorate.OnError(&err, i18n.G("can't bump version for %s"), path)
+	log.Infof(ctx, i18n.G("Bumping version for %s"), path)
+
+	cfg, err := ini.Load(path)
+
+	// If the file doesn't exist, create it and initialize the key to be updated.
+	if err != nil {
+		log.Warningf(ctx, i18n.G("error loading ini contents: %v, creating a new file"), err)
```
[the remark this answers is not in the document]
didrocks: (I’m thinking about first installation :p)

internal/watcher/watcher_test.go
didrocks: sounds good to me :) Just checking.

internal/watchdtui/watchdtui.go
```diff
@@ -0,0 +1,507 @@
...
+
+		// Sort the directories to avoid nondeterministic behavior
+		slices.Sort(dirs)
+
+		// Empty input means using the default config file
+		if confFile == "" {
+			confFile = m.defaultConfig
+		}
+
+		if err := watchdconfig.WriteConfig(confFile, dirs); err != nil {
```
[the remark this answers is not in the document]
didrocks: But we already have some other `watchd` between watcher and so on, I think the `adwatchdservice` is another layer.

README.md
```diff
@@ -886,3 +1316,8 @@ adsysd runscripts ORDER_FILE [flags]
   -v, --verbose count                     issue INFO (-v), DEBUG (-vv) or DEBUG with caller (-vvv) output
 ```
 
+## Installing development versions
+
+For every commit on the `main` branch of the `adsys` repository, the GitHub Actions CI builds a development version of the `adwatchd` project. This is *NOT* a stable version of the application and should not be used for production purposes. However, it may prove useful to preview features or bugfixes not yet available as part of a stable release.
+
+The latest development builds are available at: TODO
```
didrocks, line 1323: You still have a TODO here :p

cmd/adwatchd/integration_tests/adwatchd_service_test.go
```diff
@@ -194,3 +194,59 @@ Watched directories:
...
+			r, w, err := os.Pipe()
+			require.NoError(t, err, "Setup: pipe shouldn't fail")
+
+			app := commands.New()
+
+			changeAppArgs(t, app, "", "service", name, "-c", "badconf")
+
+			err = app.Run()
+			if tc.wantConfig {
+				require.ErrorContains(t, err, "invalid configuration file")
```
didrocks, line 224: Here, you probably want the test to continue in case of error to check for the second require below. I would use then assert for both (it’s basically 2 tests one after another, but I understand that you want to keep both in the same place)

internal/config/watchd/watchd.go
```diff
@@ -70,31 +83,36 @@ func WriteConfig(confFile string, dirs []string) error {
...
 	if !found {
-		return "", fmt.Errorf(i18n.G("missing config file in CLI arguments"))
+		return configFile, err
 	}
 
 	// Remove trailing quotes and spaces (quotes are added if the path to the
 	// config file contains spaces)
 	configFile = strings.Trim(configFile, `" `)
+	if configFile == "" {
+		return configFile, err
```
didrocks, line 114: I find: `return "", err` more readable (classic usage of returning the default value, instead of wondering why we return configFile here to have to read the line above… Ah it’s empty). Less cognitive load :p

internal/config/watchd/watchd_test.go
didrocks: Sorry, I was probably unclear, but I meant an additional test (one where both are dirs, as you had, one or more tests where one is a file, and maybe a mix of both?)

internal/watcher/watcher.go
```diff
@@ -80,10 +80,11 @@ func New(ctx context.Context, initialDirs []string, opts ...option) (*Watcher, e
 			case startCmd:
 				// Start from service doesn't pass a context explicitly
 				parentCtx := c.ctx
-				if parentCtx == context.TODO() {
-					parentCtx = ctx
+				if parentCtx == nil {
```
didrocks, line 83: Hum, I prefer the previous flow (in the quest of killing as much `else` as possible:
```golang
if parentCtx == nil {
    *parentCtx = ctx
}
ctx, cancel = context.WithCancel(*parentCtx)
```

internal/watcher/watcher_test.go
```diff
@@ -126,7 +126,7 @@ func TestWatchDirectory(t *testing.T) {
 			// Instantiate the object
 			w, err := watcher.New(context.Background(), dirs)
 			if tc.wantErrNew {
-				require.Error(t, err, "Create should have failed but hasn't")
+				require.Error(t, err, "watcher.New should have failed but hasn't")
```
didrocks, line 129: It’s just New(), we are testing `watcher`, which is implicit.

didrocks (changes requested):
And fully done with the review!

Some new comments, but very well done. Most of the issues have been nicely tackled :)

Another note: I saw that you successfully factored in `Chdir` in testutils and changed the adsys policy tests for it. However, there is some `os.Chdir` in tests in `internal/config`. Mind changing it?

Also, I haven’t found the changes in the documentation (apart from renaming the files) that I suggested. Could be a Github rendering issue or not clicking on the correct commit. I left the relevant comment unresolved then :)

didrocks:
EDIT: I opened directly the doc files and they are indeed updated. Mark the relevant conversations are resolved.

cmd/adwatchd/commands/service.go
```diff
@@ -5,14 +5,27 @@ import (
...
+		Use:   "service COMMAND",
+		Short: fmt.Sprintf(i18n.G("Manages the %s service"), watchdconfig.CmdName),
+		Long:  fmt.Sprintf(i18n.G(`The service command allows the user to interact with the %s service. It can manage and query the service status, and also install and uninstall the service.`), watchdconfig.CmdName),
+		Args:  cmdhandler.SubcommandsRequiredWithSuggestions,
+		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+			if a.rootCmd.Flags().Changed("config") {
+				return fmt.Errorf(i18n.G("cannot use --config with this subcommand"))
+			}
+
+			return a.rootCmd.PersistentPreRunE(cmd, args)
```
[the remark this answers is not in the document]
didrocks: and this is not needed anymore.

internal/watcher/watcher.go
didrocks: ah, indeed, if parentCtx is nil! :p

internal/config/watchd/watchd_test.go
didrocks: No more tests for this as no more functions :)

cmd/adwatchd/integration_tests/adwatchd_run_test.go
```diff
@@ -114,31 +114,3 @@ func TestRunReactsToConfigUpdates(t *testing.T) {
 	}
 	require.NoError(t, appErr, "App should exit without error")
 }
-
-func TestRunCanQuitWithCtrlC(t *testing.T) {
```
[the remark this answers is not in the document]
didrocks: ack

cmd/adwatchd/commands/main_windows.go
```diff
@@ -15,7 +16,8 @@ import (
 func (a *App) Quit(_ syscall.Signal) error {
 	a.WaitReady()
 	if !service.Interactive() {
-		return fmt.Errorf(i18n.G("not running in interactive mode"))
+		log.Debug(context.Background(), i18n.G("Calling quit on a non-interactive service is a no-op"))
```
didrocks, line 19: s/no-op/useless/ maybe?

cmd/adwatchd/commands/main_linux.go
```diff
@@ -15,7 +16,8 @@ import (
 func (a *App) Quit(sig syscall.Signal) error {
 	a.WaitReady()
 	if !service.Interactive() {
-		return fmt.Errorf(i18n.G("not running in interactive mode"))
+		log.Debug(context.Background(), i18n.G("Calling quit on a non-interactive service is a no-op"))
```
didrocks, line 19: ditto

internal/watcher/watcher.go
```diff
@@ -356,7 +355,7 @@ func bumpVersion(ctx context.Context, path string) (err error) {
 	log.Infof(ctx, i18n.G("Bumping version for %s"), path)
 
 	// Windows-generated files do not have spaces around the equals sign.
-	if !ini.PrettyFormat {
+	if ini.PrettyFormat {
```
didrocks, line 358: hum, basically, you always sert it from True to False, why not doing it unconditionnally then? (the if is useless?)

internal/watchdservice/watchdservice.go
```diff
@@ -202,6 +201,28 @@ func (s *WatchdService) waitForStatus(ctx context.Context, status service.Status
 	return nil
 }
 
+func (s *WatchdService) waitForError(ctx context.Context, err error) error {
```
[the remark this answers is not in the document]
didrocks: I am unsure to understand this function (well, I do from the code but I wonder why we abstract it)? Do we expect to reuse it? I think it makes more sense to inline is directly in `uninstall` with a comment telling we are waiting for the Status to report an error as the service is uninstalled. WDYT?

didrocks (commented):
TODO: check ADWATCHD_SKIP_INTEGRATION_TESTS consequence on package build and autopkgtests (but we can do that while merging).

I haven’t checked yet the new interactive code, the rest has some few comments but as you force-pushed I can’t comment anymore, hence publishing this :p

didrocks:
The reload/reconfigure existing config logic looks excellent with the additional tests!
So, I guess the few remaining comments above are the last ones to address :)

I noticed you had several golden files unused that you removed without tests being removed (yeah, leftovers can happen a lot). What you can do it to rm the golden directories and rerun with -update, then you just git status :)

Another note: have you checked the coverage report and see if there is any easy target?

didrocks:
Ok, so basically, what’s remaining is, if I’m correct (I didn’t see the diff here nor when I did reset locally in case it was part of the rebase):
- You haven’t changed in config.Init definition the variable name, as it’s not the rootCmd anymore, but the cmd with config flag attached.
- removal of `if !ini.PrettyFormat {`

Then, we can try to upload it together to a ppa and check the autokpgtests.

didrocks:
> Mmm as I mentioned in the review comment if I remove the if I get a race condition when running the watcher tests. Since we run those in parallel multiple goroutines go through the bumpVersion function at the same time and write the ini.PrettyFormat global. This is the only way I managed to fix it, but I'm open to other ways of doing it thinking. Let me know if there's a better way of tackling this.

This doesn’t really fix the race in case of concurrent executions:
- 2 goroutines can read the value, see it’s true
- then the 2 set write to it concurrently


BTW, `TestRefreshGracePeriod` is failing for me, even if the guard, I wonder if it’s not another show of this. Let me think of a way to fix this race and see.

didrocks (approved):

didrocks:
Ok, I think we will have a look for tests during package build and our autopkgtests infra, but at least, for now, it passes :) Let’s keep an eye on it.

Very well done on this work! Merging :)

PR #349, opened 2022-06-15
Adjust golden paths for Windows

GabrielNagy:
Some characters, such as `<>:#/|`, cannot be used for Windows path names. Attempting to check out the repository on Windows will fail in this case.

Update the affected filenames to remove/replace the illegal characters.

Sample error: https://github.com/ubuntu/adsys/runs/6780605093?check_suite_focus=true
```
  Error: error: invalid path 'cmd/adsysd/integration_tests/testdata/PolicyUpdate/golden/Current user with default domain completion/run/users/<CURRENT_UID>/scripts/.ready'
  Error: The process 'C:\Program Files\Git\bin\git.exe' failed with exit code 128
```


internal/ad/adsys-gpolist_test.go
```diff
@@ -201,6 +202,9 @@ func TestAdsysGPOList(t *testing.T) {
 				tc.url = "ldap://ldap_url"
 			}
 
+			// Replace non-Windows friendly characters in the test name
+			name = strings.Replace(name, ":", "", -1)
```
didrocks, line 206: Same call

cmd/adsysd/integration_tests/adsysctl_policy_test.go
```diff
@@ -128,6 +128,9 @@ func TestPolicyApplied(t *testing.T) {
 			// Reset color that we disable on client when we request --no-color
 			color.NoColor = false
 
+			// Replace non-Windows friendly characters in the test name
+			name = strings.Replace(name, `\`, "_", -1)
```
didrocks, line 132: I suggest you place that in testutils, which replacing ":", "\" and others in chain with a function name like normalize_golden() or something around that, (and only call it for the golden filename, don’t replace the name itself)

didrocks, line 132: Also, you can replace "Replace" with ReplaceAll (since Go 1.17 IIRC?).

didrocks (changes requested):
Thanks for separating this! Do you mind looking at factoring out as mentioned above, and only impact the golden filename?

didrocks (approved):
That’s perfect, thanks for the work there! Merging :)

PR #350, opened 2022-06-15
Add the current executable dir to the config load path

GabrielNagy:
In adwatchd we would like to use the executable directory as the default config path.

Update the config API to add the executable path to the config search path, after all the other paths.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"os"
+	"path/filepath"
 
 	"github.com/fsnotify/fsnotify"
 	"github.com/sirupsen/logrus"
@@ -54,6 +56,12 @@ func Init(name string, rootCmd cobra.Command, vip *viper.Viper, configChanged fu
 		vip.AddConfigPath("./")
 		vip.AddConfigPath("$HOME/")
 		vip.AddConfigPath("/etc/")
+		// Add the executable path to the config search path.
+		if binPath, err := os.Executable(); err != nil {
+			log.Warningf(context.Background(), i18n.G("Failed to get current executable path, not adding it as a config dir: %v"), err)
+		} else {
+			vip.AddConfigPath(filepath.Dir(binPath))
+		}
 	}
 
 	if err := vip.ReadInConfig(); err != nil {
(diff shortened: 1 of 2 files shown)


didrocks:
Yeah, I’m thinking again about that one. Do we really need it?

Like if you pass:
adwatchd --config foo

For interactive configuration, I would expect foo to exist, then, you prefill the first field entry with it (as it’s the config you want to change). I don’t think we should use that just to prefill the first field entry without the configuration existing beforehand.

Am I missing anything obvious?

didrocks:
> @didrocks the issue is that we set a default value for the config flag in adwatchd which causes config.Init to fail hard if the file doesn't exist:

Ah, I do remember now :)

< your other proposal >

> Does this seem more reasonable?

It really does to me. This is in line with the path search of local/home/global. Just a note: I would add it after "/etc" to me.  ("." is first)

> If the service is installed, either interactively or through adwatchd service install, the service arguments will always contain the -c argument.

Yes, until an admin mess with the windows service manager UI :)

internal/config/config.go
```diff
@@ -54,6 +56,14 @@ func Init(name string, rootCmd cobra.Command, vip *viper.Viper, configChanged fu
 		vip.AddConfigPath("./")
 		vip.AddConfigPath("$HOME/")
 		vip.AddConfigPath("/etc/")
+
```
didrocks, line 59: nitpick: I wouldn’t add an extra line here and keept it with the previous one as it’s the same "stenza"

internal/config/config.go
```diff
@@ -54,6 +56,14 @@ func Init(name string, rootCmd cobra.Command, vip *viper.Viper, configChanged fu
...
+		// Add the executable path to the config search path.
+		binPath, err := os.Executable()
+		if err != nil {
```
didrocks, line 62: Fun fact, you can (and I suggest ;) that you use:
`if binPath, err := os.Executable(); err != nil {`

And yes, before you ask it, `binPath` scope will be available in the `else` close.

That was the go trick of the day ;)

internal/config/config_test.go
```diff
@@ -108,6 +109,12 @@ func TestInit(t *testing.T) {
 			configFileContent: "value: filecontentvalue",
 			want:              "filecontentvalue", wantCallbackCalled: 1,
 		},
+		"Load configuration from executable dir": {
```
didrocks, line 112: With a test without asking! You got the spirit :) \o/

didrocks (changes requested):
One small change now that the formatting pass (and I’m sure the test will pass ;)

didrocks (approved):
I will surely enjoy the "new content" :)

PR #351, opened 2022-06-16
Add the ability to merge multiple coverages

GabrielNagy:
For adwatchd we will have a number of packages where tests must be run with sudo on Linux, as indicated by the `ADSYS_SKIP_SUDO_TESTS` environment variable.

As we want coverage for both the sudo and non-sudo runs, update the coverage generation to concatenate the output of both runs.

We now write the "mode: set" line manually since it's only needed once in the final coverage file.

### DIFF
--- a/.github/workflows/qa.yaml
+++ b/.github/workflows/qa.yaml
@@ -46,6 +46,8 @@ jobs:
   tests:
     name: Tests
     runs-on: ubuntu-latest
+    env:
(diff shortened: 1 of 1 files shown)


didrocks:
> We now write the "mode: set" line manually since it's only needed once in the final coverage file.

Ahah, I’m sure you got the same issue than we had when merging previously coverage file :)

PR #356, opened 2022-06-21
CI: Use go setup v3 to cache and use version from go.mod

didrocks:
We will always be in sync with go.mod this way for our CI from GitHub Action.


didrocks:
So, with that change, we have the confirmation that indeed 1.18 is used (see https://github.com/ubuntu/adsys/runs/6985676741?check_suite_focus=true which was using 1.16) and there is no need for symlink @GabrielNagy. Do you mind reviewing and rebase on this?

PR #359, opened 2022-06-24
Allow config.Init to work on non-persistent flags

GabrielNagy:
Previously we made the assumption that the command passed to config.Init is always the root command. This works well when the config flag is persistent and defined on the root command.

Update the config initialization API to also work with cases where we don't want to define a persistent root flag and only have the `--config` flag on subcommands.

It is the responsibility of the caller to ensure that the correct Cobra command is passed to config.Init.


didrocks (approved):
Looking good! Thanks for separating this with the test. Let’s merge :)

PR #367, opened 2022-06-29
Add autopkgtest configuration for adwatchd

GabrielNagy:
More details in the individual commit messages

### DIFF
--- a/debian/tests/control
+++ b/debian/tests/control
@@ -1,4 +1,9 @@
-# Skip integration tests when building package: they need docker images.
-Test-Command: ADSYS_SKIP_INTEGRATION_TESTS=1 go test -v -mod=vendor ./...
+# Run regular tests without root
+Test-Command: ./debian/tests/test no-sudo
 Restrictions: allow-stderr
+Depends: @builddeps@
+
+# Run only tests that require root
+Test-Command: ./debian/tests/test sudo
+Restrictions: allow-stderr, needs-root
 Depends: @builddeps@
--- a/debian/tests/test
+++ b/debian/tests/test
@@ -1,0 +1,31 @@
+#!/bin/sh
+
+set -e
+
+HELP_MSG="Script argument must be one of: no-sudo, sudo"
+
+if [ "$#" -ne 1 ]; then
+    echo "$HELP_MSG"
+    exit 1
+fi
+
+case $1 in
+    no-sudo)
+        echo "Running non-root tests..."
+        export ADSYS_SKIP_INTEGRATION_TESTS=1
+        export ADSYS_SKIP_SUDO_TESTS=1
+        PACKAGES_TO_TEST=./...
+        ;;
+    sudo)
+        echo "Running root tests..."
+        DIR_NAME=$(dirname "$0")
+        PACKAGES_TO_TEST=$(cat "$DIR_NAME"/.sudo-packages)
+        ;;
+    *)
+        echo "$HELP_MSG"
+        exit 1
+        ;;
+esac
+
+# shellcheck disable=SC2086 # Splitting is intentional
+go test -v -mod=vendor $PACKAGES_TO_TEST
--- a/internal/watcher/watcher.go
+++ b/internal/watcher/watcher.go
@@ -167,6 +167,10 @@ func (w *Watcher) UpdateDirs(ctx context.Context, dirs []string) (err error) {
 	defer decorate.OnError(&err, i18n.G("can't update directories to watch"))
 	log.Debugf(ctx, i18n.G("Updating directories to %v"), dirs)
 
+	if len(dirs) == 0 {
+		return errors.New(i18n.G("need at least one directory to watch"))
+	}
+
 	for _, dir := range dirs {
 		if _, err := os.Stat(dir); os.IsNotExist(err) {
 			return fmt.Errorf(i18n.G("directory %q does not exist"), dir)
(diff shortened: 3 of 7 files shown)


internal/watcher/watcher.go
```diff
@@ -167,14 +167,18 @@ func (w *Watcher) UpdateDirs(ctx context.Context, dirs []string) (err error) {
...
+
 	for _, dir := range dirs {
 		if _, err := os.Stat(dir); os.IsNotExist(err) {
 			return fmt.Errorf(i18n.G("directory %q does not exist"), dir)
 		}
 	}
 
 	if err := w.send(&ctx, stopCmd, nil); err != nil {
-		return err
+		log.Debugf(ctx, i18n.G("Error stopping watcher: %v"), err)
```
didrocks, line 181: I would use a warning for this.

internal/watcher/watcher.go
```diff
@@ -167,14 +167,18 @@ func (w *Watcher) UpdateDirs(ctx context.Context, dirs []string) (err error) {
 	defer decorate.OnError(&err, i18n.G("can't update directories to watch"))
 	log.Debugf(ctx, i18n.G("Updating directories to %v"), dirs)
 
+	if len(dirs) == 0 {
+		return errors.New(i18n.G("need at least one directory to watch"))
```
didrocks, line 171: Can we cover this error case with new tests?

debian/tests/control
```diff
@@ -1,4 +1,9 @@
 # Skip integration tests when building package: they need docker images.
-Test-Command: ADSYS_SKIP_INTEGRATION_TESTS=1 go test -v -mod=vendor ./...
+Test-Command: ADSYS_SKIP_INTEGRATION_TESTS=1 ADSYS_SKIP_SUDO_TESTS=1 go test -v -mod=vendor ./...
 Restrictions: allow-stderr
 Depends: @builddeps@
+
+# Run tests that require root.
+Test-Command: go test -v -mod=vendor ./cmd/adwatchd/integration_tests ./internal/watchdtui
```
didrocks, line 7: I wonder if we can use the sudo_packages list exported somewhere so that we source it here and in CI (to avoid repetition and forgetting to update one place if the list grows). Not a blocker though.

didrocks (changes requested):
This looks good to me :) Nice work and hunt on debugging those!

I would request an additional test for the new code path and some adjustement in logging, otherwise, this will be ready to merge :)

Thanks!

.env
```diff
@@ -0,0 +1 @@
+SUDO_PACKAGES="./cmd/adwatchd/integration_tests ./internal/watchdtui"
```
didrocks, line 1: nitpick: .testenv? :p

debian/tests/test
```diff
@@ -0,0 +1,23 @@
...
+
+set -ex
+
+DIR_NAME=$(dirname "$0")
+SCRIPT_NAME=$(basename "$0")
+
+# Source the environment variables from the .env file
+. "$DIR_NAME"/../../.env
+
+if [ "$SCRIPT_NAME" = "test" ]; then
```
didrocks, line 11: A little bit too much black magic I think with the 2 test fails. Maybe it will be simpler to have one test file with some arguments? like with-sudo/no-sudo?

That way we keep the one single script to run them all, but it’s a little bit more explicit.

internal/watcher/watcher_test.go
```diff
@@ -341,6 +341,60 @@ func TestUpdateDirsFailing(t *testing.T) {
 	assertGPTVersionEquals(t, destRemove, 3)
 }
 
+func TestUpdateDirsWithEmptyDirSlice(t *testing.T) {
```
didrocks, line 344: we couldn’t have it added to the previous table testing with a WantErr?

internal/watcher/watcher_test.go
```diff
@@ -341,6 +341,60 @@ func TestUpdateDirsFailing(t *testing.T) {
...
+	err = w.Start(mockService{})
+	require.NoError(t, err, "Setup: Can't start watcher")
+	defer w.Stop(mockService{})
+
+	// Update the watched directories with an empty slice
+	err = w.UpdateDirs(context.Background(), []string{})
+	require.ErrorContains(t, err, "need at least one directory to watch", "Updating directories should have failed")
+}
+
+func TestUpdateDirsOnStoppedWatcher(t *testing.T) {
```
didrocks, line 365: same, can’t be factor in table testing?

debian/tests/test
```diff
@@ -0,0 +1,31 @@
+#!/bin/sh
+
+set -e
+
+HELP_MSG="The script requires exactly 1 parameter"
```
didrocks, line 5: Which ones :p (yes I know, but as we only have 2 options, better to write them :p)

debian/tests/test
```diff
@@ -0,0 +1,31 @@
...
+        DIR_NAME=$(dirname "$0")
+        PACKAGES_TO_TEST=$(cat "$DIR_NAME"/.sudo-packages)
+        ;;
+    *)
+        echo "$HELP_MSG"
+        exit 1
+        ;;
+esac
+
+# shellcheck disable=SC2086 # Splitting is intentional
```
didrocks, line 30: Sorry, what split?

internal/watcher/watcher_test.go
didrocks: ack, let’s have a look in a separate PR if this could be refactored. If you feel there is not enough commonality, feel free to discare.

didrocks (commented):
one small thingy and let’s merge afterwards!

debian/tests/test
didrocks: ahah, nice! Thanks for the explanation :)

didrocks (approved):

didrocks:
Excellent! Let’s merge before preparing a release :)

PR #370, opened 2022-07-01
Update packaging to work across architectures

GabrielNagy:
Opening this for early discussion as we still have to fix the failing `armhf` [build](https://launchpadlibrarian.net/610641368/buildlog_ubuntu-kinetic-armhf.adsys_0.8.6ubuntu12_BUILDING.txt.gz):
```
# github.com/ubuntu/adsys/vendor/github.com/mvo5/libsmbclient-go
In file included from src/github.com/ubuntu/adsys/vendor/github.com/mvo5/libsmbclient-go/libsmbclient.go:17:
/usr/include/samba-4.0/libsmbclient.h:84:13: error: size of array 'smbc_off_t_should_be_at_least_64bits_use_LFS_CFLAGS' is too large
   84 | typedef int smbc_off_t_should_be_at_least_64bits_use_LFS_CFLAGS[sizeof(off_t)-7];
      |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
------------------------------------
## Update packaging to work across architectures
- build/ship adwatchd on amd64 and arm64, other architectures are not supported by go
- build adwatchd with PIE on amd64 as other architectures are not supported
- run sudo autopkgtests on supported architectures (amd64 and arm64), skip them otherwise
- unify the rules comment format to be uppercase with no period at the end

## Force a sync when running generators
There's an edge case on the LP builders where due to some race condition we can get an error like the following after running the `go generate` command:

    cp: error reading 'debian/tmp/usr/share/polkit-1': Is a directory

It usually happens with different directories from the install file. From local/LP testing it looks like forcing a sync makes the issue disappear.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -39,7 +39,7 @@ Description: ${source:Synopsis}
  ${source:Extended-Description}
 
 Package: adsys-windows
-Architecture: any
+Architecture: amd64 arm64
 Built-Using: ${misc:Built-Using},
 Depends: ${shlibs:Depends},
          ${misc:Depends},
--- a/debian/rules
+++ b/debian/rules
@@ -2,15 +2,28 @@
 #export DH_VERBOSE = 1
 
 export GOCACHE=$(shell mktemp -d /tmp/gocache-XXXX)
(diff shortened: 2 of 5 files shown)


didrocks (approved):
Apart from the samba part, this is looking good to me. Let’s merge.

PR #371, opened 2022-07-04
Fix LP builder race conditions

GabrielNagy:
Looks like the sync didn't fully solve our issues. The failures are more sporadic now but they still show up from time to time.

A way of working around would have been to do something similar to 280196d0, but in our case the failing paths differ (I've seen `locale`, `polkit-1`, `zsh`, `bash-completion` until now). So instead of force-creating those directories beforehand, add a measly `time.Sleep` before exiting the generator.

I've tested this fix by triggering 3 sets of builds on LP, each one without errors.

### DIFF
--- a/internal/generators/copy.go
+++ b/internal/generators/copy.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"path/filepath"
 	"syscall"
+	"time"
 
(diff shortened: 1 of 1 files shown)


didrocks (approved):
looks good! Thanks for doing this!

PR #372, opened 2022-07-04
Increase copy generator sleep to 1 second

GabrielNagy:
We still got a failure on ppc64el (that went away on rebuild).

As we don't run this generator outside of packaging and CI it should be fairly non-intrusive to increase the sleep time to a second. Also it might be better for the sync to happen prior to the sleep.

We'll monitor this and see how it goes before spending more time on the issue.


didrocks (approved):
+1 feel free to merge once the tests pass :)

PR #379, opened 2022-07-08
Add CODEOWNERS to adsys

GabrielNagy:
This will request the review of @ubuntu/enterprise-desktop-team for every new PR that is opened which is not a draft.


didrocks:
LGTM! Thanks :)

didrocks (approved):

PR #381, opened 2022-07-08
Bump runner version for the Jira sync workflow

GabrielNagy:
The ubuntu/sync-issues-github-jira action can now convert GitHub Markdown to JIRA Markdown. However, this is only supported starting with Ubuntu 22.04 which is not yet `ubuntu-latest`.

Explicitly request `ubuntu-22.04`. We can revert this change when GitHub changes `ubuntu-latest` to point at `ubuntu-22.04`.


didrocks (approved):
Ack! :)

PR #405, opened 2022-08-04
Clean up GitHub Actions workflows

GabrielNagy:
- update used actions to latest
- remove git safe path workaround
- install curl in the same step as the other dependencies
- do not use containers unless explicitly needed

### DIFF
--- a/.github/workflows/auto-updates.yaml
+++ b/.github/workflows/auto-updates.yaml
@@ -11,17 +11,13 @@ jobs:
   update-po:
     name: Update po files
     runs-on: ubuntu-latest
-    container: ubuntu:rolling
     steps:
-      - name: Install dependencies, including git for checkout
+      - name: Install dependencies
         run: |
-          export DEBIAN_FRONTEND=noninteractive
-          apt update
-          apt install -y gettext git
-      - name: work around permission issue with git vulnerability (we are local here). TO REMOVE
-        run: git config --global --add safe.directory /__w/adsys/adsys
+          sudo DEBIAN_FRONTEND=noninteractive apt update
+          sudo DEBIAN_FRONTEND=noninteractive apt install -y gettext
       # Checkout code with git
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
         with:
           ref: main
       # Install go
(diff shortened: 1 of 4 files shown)


.github/workflows/auto-updates.yaml
```diff
@@ -63,17 +59,13 @@ jobs:
     # This should just be "after", but we don't want the 2 jobs to push at the same time
     needs: update-po
     runs-on: ubuntu-latest
-    container: ubuntu:rolling
```
didrocks, line 66: Just a remark: the containers were needed at the time to get some fixes we needed in libsmclient-dev to generate what we needed. Indeed, not needed anymore, thanks for cleaning it up!

.github/workflows/qa.yaml
```diff
@@ -11,19 +11,18 @@ jobs:
   quality:
     name: Code quality and mod files up to date
     runs-on: ubuntu-latest
-    container: ubuntu:latest
```
didrocks, line 14: I’m more biased on this one: we want to check our CI against latest release of ubuntu. ubuntu:latest is not ubuntu-latest. It’s the last stable released version of ubuntu. So, I would keep it.

.github/workflows/qa.yaml
```diff
@@ -144,17 +139,13 @@ jobs:
   generators:
     name: Generated files are up to date
     runs-on: ubuntu-latest
-    container: ubuntu:latest
```
didrocks, line 147: ditto for the generators.

didrocks (changes requested):
Thanks for cleaning this up! I do have some remarks though that we really want to run the tests with the latest release of ubuntu to detect issues early :)

didrocks:
ack and sad :p Thanks for playing ping-pong with the runners! Tell me once it’s ready for rereview

.github/workflows/qa.yaml
```diff
@@ -13,17 +13,19 @@ jobs:
...
+          DEBIAN_FRONTEND=noninteractive apt update
+          DEBIAN_FRONTEND=noninteractive apt install -y ca-certificates git gcc pkg-config libsmbclient-dev
       - name: work around permission issue with git vulnerability (we are local here). TO REMOVE
         run: git config --global --add safe.directory /__w/adsys/adsys
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
+      - uses: actions/setup-go@v3
+        with:
+          go-version-file: go.mod
+          cache: true
```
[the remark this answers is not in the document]
didrocks: or pulled by golangci-lint? Weird indeed, as we use a container, so not the go from the host.

didrocks (approved):
LGTM, thanks for cleaning up! You can merge whenever you wish :)

PR #415, opened 2022-09-12
Deeng 420 reduce dependencies

didrocks:
Reduce dependencies by excluding tools
As we vendor our packages for adsys build on ubuntu, move tooling in a separate go.mod file.
This way, we don’t embeed uneeded CI-related dependencies in the final source package.

Ensure we install the version specified in the tools/ subdirectory for both code generation and linting.

This reduce the size of the source package by 2.2Mb.


didrocks:
As you can see, it did run with:
- grpc/protobuf version in tools/go.mod
- NEW (fixed!) golangci-lint 1.48.0 instead of latest 1.49.0, matching go.mod.

PR #424, opened 2022-09-27
Implementation for adsys mount manager to handle sharing policies for users.

denisonbarbosa:
The policy manager is supposed to parse the entries and its values, ignore the duplicate values listed and write the mounts file in the specified directory.

DEENG-457
DEENG-459


internal/policies/mount/mount.go
```diff
@@ -0,0 +1,162 @@
...
+	defer decorate.OnError(&err, i18n.G("can't apply mount policy to %s"), objectName)
+
+	log.Debugf(ctx, "Applying mount policy to %s", objectName)
+
+	if isComputer {
+		return fmt.Errorf(i18n.G("computer mounts are currently not supported"))
+	}
+
+	if len(entries) == 0 {
+		return nil
```
didrocks, line 90: Just reading the tests, it seems this use case is not covered.

internal/policies/mount/mount_test.go
```diff
@@ -0,0 +1,106 @@
...
+func TestApplyPolicy(t *testing.T) {
+	t.Parallel()
+
+	entries := GetEntries(5, 0, 1, 0, []string{"\n"})
+
+	tests := map[string]struct {
+		entries    []entry.Entry
+		objectName string
+		computer   bool
+		uid        string
```
didrocks, line 24: I saw that on this one, you got inspired by the script part, which is great :)

However, I suggest that you keep the same name for the familiarity when reading tests between packages. This way, you can see things are similar, This includes:
- a blank line between the arguments of the `ApplyPolicy` function and the side effects, like returned uid/gid.
- `userReturnedUID` instead of `uid`, to clearly say this is not an input, but that we are mocking some side-effect returned value.

internal/policies/mount/mount_test.go
```diff
@@ -0,0 +1,106 @@
...
+		gid        string
+
+		wantErr bool
+	}{
+		"successfully generates mounts file":        {entries: entries},
+		"generates no file if there are no entries": {entries: []entry.Entry{}},
+
+		"error when user is not found":    {entries: entries, objectName: "dont exist", wantErr: true},
```
didrocks, line 32: With all the other tests, you can see that we comment them beforehand with `// Error cases`. Please follow the same convention.

internal/consts/consts.go
```diff
@@ -65,4 +65,11 @@ const (
 	SubscriptionDbusObjectPath = "/com/canonical/UbuntuAdvantage/Manager"
 	// SubscriptionDbusInterface is the interface we are using for access dbus properties.
 	SubscriptionDbusInterface = "com.canonical.UbuntuAdvantage.Manager"
+
+	// Mount related properties.
+
+	// DefaultMountsFilePath is the path in which the mount entries for the user will be / %s should be formatted with the full username.
+	DefaultMountsFilePath string = "/run/adsys/users/%s/mounts"
+	//DefaultUserMountServicePath is the path in which the mount service for the user will be.
```
didrocks, line 73: This is going to be removed, but another example on why rereading is really important: you don’t have any space between `//` and the start of the comment, where all the other have. Consistency is important :)

internal/policies/mount/internal_test.go
```diff
@@ -0,0 +1,70 @@
...
+		nEntries    int
+		nErrored    int
+		nValues     int
+		nDuplicated int
+		separators  []string
+
+		wantErr bool
+	}{
+		"write mounts file with multiple entries":                          {nEntries: 5, nValues: 1},
+		"write mounts file with values separated by ','":                   {nEntries: 1, nValues: 5, separators: []string{","}},
```
didrocks, line 27: I would add some tests with one entry and one value. Also, maybe one entry, no value.

internal/policies/mount/mount_test.go
```diff
@@ -0,0 +1,106 @@
...
+		entries    []entry.Entry
+		objectName string
+		computer   bool
+		uid        string
+		gid        string
+
+		wantErr bool
+	}{
+		"successfully generates mounts file":        {entries: entries},
+		"generates no file if there are no entries": {entries: []entry.Entry{}},
```
didrocks, line 30: Even if WriteMountsFile is well tested, as here, we are in the package tests, I would add some more entries cases without going to all details, but at least, to illustrate the most common functionalities:
- some duplicated entries, not following each other
- one entry with one value only.

internal/policies/mount/helper_test.go
```diff
@@ -0,0 +1,37 @@
...
+	for i := 0; i < nEntries; i++ {
+		if i < nErrored {
+			e := entry.Entry{Err: fmt.Errorf("this entry has an error")}
+			entries = append(entries, e)
+			continue
+		}
+
+		e := entry.Entry{Key: fmt.Sprintf("mount/%d", i)}
+		dup := 0
+		value := "protocol://domain.com/mount-%d-path"
```
didrocks, line 22: I would put a mix of protocols and fictive subpath to illustrate a real entry type. That way, by looking at the golden files, we know that we support multiple paths hierachy. (Not all needs to be like that, but some).

internal/policies/mount/helper_test.go
```diff
@@ -0,0 +1,37 @@
+package mount
+
+import (
+	"fmt"
+
+	"github.com/ubuntu/adsys/internal/policies/entry"
+)
+
+// GetEntries generates entries to be used in tests.
+func GetEntries(nEntries, nErrored, nValues, nDuplicated int, sep []string) []entry.Entry {
```
didrocks, line 10: `sep` is a byte, that way, you don’t need the `sep[n%len(sep)]` dance.

internal/policies/mount/internal_test.go
```diff
@@ -0,0 +1,70 @@
...
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+
+			mountsPath := filepath.Join(t.TempDir(), "mounts")
+
+			if len(tc.separators) == 0 {
+				tc.separators = append(tc.separators, "\n")
+			}
+
+			err := writeMountsFile(GetEntries(tc.nEntries, tc.nErrored, tc.nValues, tc.nDuplicated, tc.separators), WithMountsFilePath(mountsPath))
```
didrocks, line 47: I would keep GetEntries in a separate line, that way, you have:

- Setup, with clear inputs you are preparing
- The call itself, which is what you want to test
- Then the assertions.

internal/policies/mount/helper_test.go
```diff
@@ -0,0 +1,37 @@
...
+			continue
+		}
+
+		e := entry.Entry{Key: fmt.Sprintf("mount/%d", i)}
+		dup := 0
+		value := "protocol://domain.com/mount-%d-path"
+		for n := 0; n < nValues; n++ {
+			v := fmt.Sprintf(value, c)
+			c++
+			if dup < nDuplicated {
```
didrocks, line 26: I haven’t checked it, but to me, the duplicates are following each other in the entries. We want, for a more realistic use case, having them not following each other:

* within the same entry
* and between entries

didrocks (changes requested):
This is a start, but some works are still needed.

So, I’m not doing a full review, as you can see there are some things to change already. I will once again highlight the following:
- take inspiration of the scripts manager which is really similar to this one.
- look at dead code/useless assignement you might have in your code, like the service file. Also, if you need something here which is not in the script manager, question it twice: how is the user mount different from the user script running? Why do I need to add an additional const where there is none for scripts? Why would I need a mutex for it?
- Similarly, if you copy some code between modules, try to not renaming the variables, or maybe we should rename both? But don’t create inconsistency in the code.
- There are some spaces/blank lines in the original code with comment convention (like `// Error cases`) to clearly identify the error stenza in the tests. Again, it’s a question of concistency :)
- pay attention to details by rereading you (comments at the correct places, like separating tests, follow the same naming convention, look at unintentional spaces or missing ones…)
- the golden files are not in a directory under "golden" like the rest of the code. This is the way to separate fixtures (see below) from golden files. Again, look at the user scripts manager for inspiration.
- Look at the tests in the scripts manager too, you will see that there is some examples with pre-existing directory (`run_dir`), all of them, or none. Also, there are some tests with unwritable directories to cover the can’t write test. Try to keep the same structure. You will see there that the test fixtures are only for machine because this is the same logic for both machine and users (so no need to retest them), but here, we are tackling the user first.
- also, as stated multiple times, please run the coverage tool in your tests. If I’m correct, some branches are not covered (empty != nil, error writing files).

Once you redo your branch, and reread once more, please reassign it to me. Feel free to ping me if anything is not clear enough :)

internal/policies/mount/testdata/TestWriteMountsFile/write_mounts_file_with_values_separated_by_','/mounts
```diff
@@ -0,0 +1,5 @@
+protocol://domain.com/mount-0-path
```
didrocks, line 1: Also, the issue with that generation is that they are always ascii ordered,which results in ascii order in the golden files and might give the false impression we are ordering entries. So the generation needs to be a little bit smarter to avoid this.

internal/policies/mount/mount_test.go
```diff
@@ -0,0 +1,106 @@
...
+			if len(tc.entries) == 0 {
+				require.Error(t, err, "Expected no mounts file")
+				return
+			}
+			require.NoError(t, err, "Expected to read generated mounts file")
+
+			testdata := filepath.Join("testdata", t.Name())
+			wantPath := filepath.Join(testdata, "mounts")
+			if update {
+				err = os.MkdirAll(testdata, 0750)
```
didrocks, line 86: Never write directly to testdata please:
- Here, you don’t clean old files (so any files that we remove in the future during the golden generation will still stay).
- If there is a middle in the middle of the refresh, you end up with mixed parsed data.

Also, there are some functions in testutils (again, look at the script manager to get inspiration), using `CompareTreesWithFiltering` for instance. This allow you then to ensure you keep and tests properly the whole hierarchy that the manager should creates in addition to the file itself.

internal/policies/mount/internal_test.go
```diff
@@ -0,0 +1,73 @@
...
+			require.NoError(t, err, "Expected to be able to change permissions but got an error instead.")
+
+			err = writeMountsFile(gotPath+"/mounts", EntriesForTests[tc.entries])
+			if tc.wantErr {
+				require.Error(t, err, "Expected an error when writing mounts file but got none")
+				return
+			}
+			require.NoError(t, err, "Expected no error when writing mounts file but got one")
+
+			goldenPath := filepath.Join("testdata", t.Name(), "golden", "mounts")
```
didrocks, line 63: hum, we end up with `golden/mounts/mounts`. Mind looking if we only put `golden` in the destination directory to have `golden/mounts`?

internal/policies/mount/mount_test.go
```diff
@@ -0,0 +1,172 @@
...
+
+		secondCall      string
+		userReturnedUID string
+		userReturnedGID string
+
+		wantErr bool
+	}{
+		"successfully generates mounts file for one entry with one value": {},
+
+		"successfully generates mounts file for multiple entries with one value":       {entries: "multiple entries with one value"},
```
didrocks, line 67: If we change the order in the unit tests, please sync the changes here :)

internal/policies/mount/mount_test.go
```diff
@@ -0,0 +1,172 @@
...
+
+func TestApplyPolicy(t *testing.T) {
+	t.Parallel()
+
+	tests := map[string]struct {
+		entries    string
+		objectName string
+		computer   bool
+
+		usrDirPerm os.FileMode
```
didrocks, line 57: See my comment on the unit tests. Only one case change the permission, so I would go for something which explains more the intention of "not able to write in userDir) (see also `usrDir` vs `userDir`)

internal/policies/mount/mount_test.go
```diff
@@ -0,0 +1,172 @@
...
+
+			goldenPath := filepath.Join("testdata", t.Name(), "golden")
+
+			makeIndependentOfCurrentUID(t, testRunDir, usr.Uid)
+			testutils.CompareTreesWithFiltering(t, testRunDir, goldenPath, mount.Update)
+		})
+	}
+}
+
+// makeCurrentUIDIndepmakeIndependentOfCurrentUIDendent renames any file or directory which exactly match uid in path and replace it with 4242.
```
didrocks, line 149: Comment title doesn’t match the function. (I guess incorrect double copy/paste and not enough reread from my side :p).

Feel free to fix it in a separate comment (but same PR is ok) in `scripts/scripts_test.go` too.

internal/policies/mount/mount_test.go
```diff
@@ -0,0 +1,172 @@
...
+			if tc.entries != "" {
+				entries = mount.EntriesForTests[tc.entries]
+			}
+
+			if tc.usrDirPerm != 0 {
+				opts = append(opts, mount.WithPerm(0100))
+			}
+
+			m, err := mount.New(opts...)
+			require.NoError(t, err, "Expected no error but got one.")
```
didrocks, line 127: Same, this is a `Setup:` at this level as it’s not what we are testing exactly.

TODO: we should all come with a convention to ensure this is done everywhere, because it’s so easy to miss one…

didrocks (changes requested):
Despite the amount of comments, this is really really good! Thanks for making those changes :)

I find the new version way more readable, and thus, maintainable on the long term. Also, the testing plan is good.

Comments are generally nitpicks (apart from the chown that you forgot on the mounts file itself, which we can’t really test but we know it needs to be there + changing the permissions on it).

If you see some comments that applies to the scripts manager too, feel free to change them in a separate commit, but same part of the PR for easyness (or do another PR if you prefer this, which is even better).

Nice work! Let me know if anything that I tell doesn’t make sense, but I hope I am clear enough on them!

Last thing: I haven’t checked manually the code coverage yet (codecov is really unhelpful as it doesn’t show up the file changes…). I will look if there are some error cases in file permission/creation that we can easily cover and we did not once you did the other changes.

internal/policies/mount/entries_test.go
```diff
@@ -0,0 +1,58 @@
+package mount
+
+import (
+	"fmt"
+
+	"github.com/ubuntu/adsys/internal/policies/entry"
+)
+
+// EntriesForTests is a map with multiple []entries combinations to be used in tests.
+var EntriesForTests = map[string][]entry.Entry{
```
didrocks, line 10: I would add one entry with a tag (anonymous) btw (could be a separate test case too for both unit and package test). That way, we show we don’t do anything with them and it’s a common case we will get.

internal/policies/mount/mount.go
```diff
@@ -0,0 +1,192 @@
...
+	systemCtlCmd []string
+}
+
+// New creates a Manager to handle mount policies.
+func New(opts ...Option) (m *Manager, err error) {
+	defer decorate.OnError(&err, i18n.G("failed to create new mount manager"))
+	o := options{
+		runDir:     consts.DefaultRunDir,
+		userLookup: user.Lookup,
+		perm:       0750,
```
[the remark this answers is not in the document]
denisonbarbosa: I've just used the permission that was being used in the scripts manager. Should we change that as well?

internal/policies/mount/mount_test.go
denisonbarbosa: Indeed. I agree that this should be standardized.

internal/policies/mount/mount.go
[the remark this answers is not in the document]
didrocks: Oh, I thought I removed this comment, sorry. (I was on the file itself which should be 0600, the directory itself is fine as 0750).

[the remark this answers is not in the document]
denisonbarbosa: I've tested it and it seems 0600 it's not enough.

internal/policies/mount/mount.go
```diff
@@ -0,0 +1,192 @@
...
+			if _, ok := seen[v]; ok || v == "" {
+				continue
+			}
+			p = append(p, v)
+			seen[v] = struct{}{}
+		}
+	}
+
+	// #nosec G306. This should be world-readable.
+	if err := os.WriteFile(mountsPath+".new", []byte(strings.Join(p, "\n")+"\n"), 0644); err != nil {
```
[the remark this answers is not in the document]
denisonbarbosa: The permissions are handled right after the file is written (look at ApplyPolicy). Should we already handle the permissions in the writeFile function?

internal/policies/mount/mount.go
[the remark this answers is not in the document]
didrocks: you mean, for the file or the directory? the file should be, not the directory ofc.

internal/policies/mount/mount.go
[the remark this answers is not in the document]
didrocks: I would expect this to be in writeFile TBH. (so 0600 + known user), as it’s where we are dumping it.

internal/policies/mount/internal_test.go
```diff
@@ -0,0 +1,88 @@
...
+		"write mounts file with one entry with repeatead values": {entries: "one entry with repeatead values"},
+
+		// Multiple entries cases.
+		"write mounts file with multiple entries with one value":       {entries: "multiple entries with one value"},
+		"write mounts file with multiple entries with multiple values": {entries: "multiple entries with multiple values"},
+		"write mounts file with multiple entries with the same value":  {entries: "multiple entries with the same value"},
+		"write mounts file with multiple entries with repeated values": {entries: "multiple entries with repeated values"},
+
+		// Badly formatted entries.
+		"write mounts file trimming whitespaces":           {entries: "entry with linebreaks and spaces"},
```
didrocks, line 38: I wouldn’t mix linebreaks and space (to minimize the amount of concern in this test).

internal/policies/mount/mount_test.go
```diff
@@ -0,0 +1,188 @@
...
+			if tc.entries != "" {
+				entries = mount.EntriesForTests[tc.entries]
+			}
+
+			if tc.readOnlyUserDir {
+				opts = append(opts, mount.WithPerm(0100))
+
+				// Ensures that the test dir will be cleaned after the test.
+				defer func(p string) {
+					//nolint:errcheck,gosec
```
didrocks, line 135: Always better to add a rationale on why we are ignoring the error :)
I would still do `_ = os.Chmod()` though

internal/policies/mount/mount.go
```diff
@@ -0,0 +1,195 @@
...
+	// Removes the current mounts file, if it exists, before continuing applying the policy.
+	if err = os.Remove(mountsPath); err != nil && !errors.Is(err, os.ErrNotExist) {
+		return err
+	}
+
+	if len(entries) == 0 {
+		return nil
+	}
+
+	// This creates usrDir directory.
```
didrocks, line 117: there is no more usrDir :p

internal/policies/mount/mount_test.go
```diff
@@ -0,0 +1,188 @@
...
+
+	for name, tc := range tests {
+		tc := tc
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+
+			runDir := t.TempDir()
+			if tc.invalidPerm {
+				err := os.Chmod(runDir, 0100)
+				require.NoError(t, err, "Expected to be able to change permissions but got an error instead.")
```
didrocks, line 36: Missing Setup.

internal/policies/mount/mount.go
```diff
@@ -0,0 +1,195 @@
+// Package mount implements the manager responsible to handle the file sharing
+// policy of adsys, parsing the GPO rules and setting up the mount process for
+// the requested drives. User mounts will be handled by a systemd user unit and
```
didrocks, line 3: systemd user service

didrocks:
(FTR: pending the one vs multiple entries refresh on the input)

internal/policies/mount/internal_test.go
```diff
@@ -0,0 +1,153 @@
...
+		"fails when writing on a dir with invalid permissions": {readOnlyDir: true, wantErr: true},
+		// "error when path already exists and has different ownership": {pathAlreadyExists: true, wantErr: true},
+	}
+
+	for name, tc := range tests {
+		tc := tc
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+
+			u, err := user.Current()
```
didrocks, line 85: This should be done before the for loop, as this will not change :p

internal/policies/mount/internal_test.go
```diff
@@ -0,0 +1,153 @@
...
+			iGID, err := strconv.Atoi(gid)
+			require.NoError(t, err, "Setup: Failed to convert gid to int")
+
+			filePath := filepath.Join(path, "test_write")
+
+			if tc.pathAlreadyExists {
+				err = os.WriteFile(filePath, []byte("file already existed"), 0600)
+				require.NoError(t, err, "Setup: Failed to set up pre existent file for testing")
+
+				err = os.Chown(filePath, iUID+1, iGID+1)
```
didrocks, line 115: Ah, ok, this is why it’s commented. Yeah, we can’t do that. Let’s work on this together.

internal/policies/mount/internal_test.go
```diff
@@ -0,0 +1,153 @@
...
+			s, err := os.Stat(filePath)
+			require.NoError(t, err, "Failed when fetching info of the written file")
+
+			//nolint:forcetypeassert // This happens in a controlled environment
+			gotUID := s.Sys().(*syscall.Stat_t).Uid
+			//nolint:forcetypeassert // This happens in a controlled environment
+			gotGID := s.Sys().(*syscall.Stat_t).Gid
+
+			require.Equal(t, iUID, int(gotUID), "Expected UID to be the same")
+			require.Equal(t, iGID, int(gotGID), "Expected GID to be the same")
```
didrocks, line 142: I think we already mentioned that on MM last week, but this is useless in the tests: we are running the tests under a certain user and ofc, those will be under the current user (as they are not running as root user).

internal/policies/mount/mount_test.go
```diff
@@ -0,0 +1,247 @@
...
+		key        string
+		isComputer bool
+
+		noEntries bool
+	}{
+		"successfully applies policy with valid key": {key: "user-mounts"},
+
+		"does nothing when trying to apply policy with user key and isComputer is set": {key: "user-mounts", isComputer: true},
+		"does nothing when trying to policy with unsupported key":                      {key: "not-supported-yet"},
+		"does nothing when trying to apply policy with no entries":                     {noEntries: true},
```
didrocks, line 64: I would put all users and (future computer) tests under the same umbrella here: users don’t care that this is internally 2 separate functions, but there is only one API call. So, having them all in the same table test makes more sense and make the API more discoverable.

internal/policies/mount/mount_test.go
```diff
@@ -0,0 +1,247 @@
...
+			require.NoError(t, err, "Expected no error but got one")
+
+			if tc.secondCall != "" {
+				if tc.wantErr {
+					iUID, err := strconv.Atoi(tc.userReturnedUID)
+					require.NoError(t, err, "Setup: Failed to convert uid to int")
+					iGID, err := strconv.Atoi(tc.userReturnedGID)
+					require.NoError(t, err, "Setup: Failed to convert gid to int")
+
+					err = os.Chown(filepath.Join(testRunDir, "users", tc.userReturnedGID, "mounts"), iUID+1, iGID+1)
```
didrocks, line 200: ditto, this indeed will not be possible

internal/policies/mount/mount_test.go
```diff
@@ -0,0 +1,212 @@
...
+		"successfully apply policy for entry with multiple values":  {entries: []string{"entry with multiple values"}},
+		"successfully apply policy for entry with repeatead values": {entries: []string{"entry with repeatead values"}},
+
+		// USER // Special cases.
+		"successfully apply policy with anonymous values":                                   {entries: []string{"entry with anonymous tags"}},
+		"generates no file for errored entry":                                               {entries: []string{"errored entry"}},
+		"generates no file when trying to apply policy with user key and isComputer is set": {isComputer: true},
+
+		// USER // Badly formatted entries.
+		"successfully apply policy trimming whitespaces":           {entries: []string{"entry with linebreaks and spaces"}},
```
didrocks, line 80: This case has an empty golden file result, which is due to `entry with linebreaks and spaces` not existing in entries, so being nil.

I suggest you do a second round of looking at all the goden files content and that they makes sense compared to the input/test case :)

internal/policies/mount/internal_test.go
```diff
@@ -0,0 +1,135 @@
...
+		readOnlyDir       bool
+		pathAlreadyExists bool
+
+		wantErr bool
+	}{
+		"write file with current user ownership": {},
+
+		"error when invalid uid":                               {uid: "-150", wantErr: true},
+		"error when invalid gid":                               {gid: "-150", wantErr: true},
+		"fails when writing on a dir with invalid permissions": {readOnlyDir: true, wantErr: true},
```
didrocks, line 75: I would prefix again with `error` here, so that we can run all error cases with `go test -run TestWriteFileWithUIDGID/error`

internal/policies/mount/internal_test.go
```diff
@@ -0,0 +1,135 @@
...
+			require.NoError(t, err, "Setup: Failed to convert gid to int")
+
+			filePath := filepath.Join(path, "test_write")
+
+			if tc.pathAlreadyExists {
+				err = os.MkdirAll(filePath, 0750)
+				require.NoError(t, err, "Setup: Failed to set up pre existent directory for testing")
+
+				t.Cleanup(func() {
+					//nolint:errcheck // This happens in a controlled environment
```
didrocks, line 115: I think the comment is not the one you wanted here?

internal/policies/mount/mount.go
```diff
@@ -0,0 +1,215 @@
...
+	objectPath := filepath.Join(m.runDir, "users", usr.Uid)
+	mountsPath := filepath.Join(objectPath, "mounts")
+
+	// Removes the current mounts file, if it exists, before continuing applying the policy.
+	if err = os.Remove(mountsPath); err != nil && !errors.Is(err, os.ErrNotExist) {
+		return err
+	}
+
+	// This creates userDir directory.
+	// We chown userDir to uid:gid of the user. Nothing is done for the machine
```
didrocks, line 133: As there is no more userDir variable, you can write:
`This creates user directory and chown it to uid:gid of the user`

Also, there is no need anymore for the "Nothing is done for the machine", now that you split it in a deddicated user function.

internal/policies/mount/mount.go
```diff
@@ -0,0 +1,215 @@
...
+	}
+	if gid, err = strconv.Atoi(usr.Gid); err != nil {
+		return fmt.Errorf(i18n.G("couldn't convert %q to a valid gid for %q"), usr.Gid, username)
+	}
+
+	objectPath := filepath.Join(m.runDir, "users", usr.Uid)
+	mountsPath := filepath.Join(objectPath, "mounts")
+
+	// Removes the current mounts file, if it exists, before continuing applying the policy.
+	if err = os.Remove(mountsPath); err != nil && !errors.Is(err, os.ErrNotExist) {
```
didrocks, line 128: Here, we create a race with no file available (I know, we have it in the script directory too, but let’s not have it here as it’s a single file).

Remember that we rename the file from .new to final name. My suggestion then is to only remove the file if: there is no entry. Btw, it seems that it’s not the case anymore as you won’t enter this function if there are no more entries. -> I suggest, in addition to the fix, to create a new test then (`pathAlreadyExists=true`, apply empty entries -> golden shuld have no file)

For instance, here, you remove the file, a parsing fails -> no more mount file (not even the previous one).

didrocks (changes requested):
There are some regressions compared to previous version that needs to be fixed (due to the split user/machine). It prooves there is the need for more test cases :)
That should be easy to fix anyway.

However, as stated in some comments, please have a second, deep, look at every golden files that are produced as at list one doesn’t have the expected content. Especially the one where the old value should be overriden (Second Apply or overwrite existing start state).

Finally, some minor cosmetic changes, but we are getting there!

internal/policies/mount/mount_test.go
```diff
@@ -0,0 +1,212 @@
...
+			if tc.readOnlyUsersDir {
+				err := os.MkdirAll(filepath.Join(runDir, "users"), 0750)
+				require.NoError(t, err, "Setup: Expected no error when creating users dir for tests.")
+				testutils.MakeReadOnly(t, filepath.Join(runDir, "users"))
+			}
+
+			if tc.pathAlreadyExists {
+				err := os.MkdirAll(filepath.Join(runDir, "users", u.Uid, "mounts"), 0750)
+				require.NoError(t, err, "Setup: Expected no error when creating mounts dir for tests.")
+				err = os.WriteFile(filepath.Join(runDir, "users", u.Uid, "mounts", "exists"), []byte("already exists."), 0600)
```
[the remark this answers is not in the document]
denisonbarbosa: os.Remove() does not target a non empty directory, so in order to keep the os.Remove call in applyUserPolicy and test the pathAlreadyExists case, the directory must not be empty.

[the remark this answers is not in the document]
didrocks: Unsure what you mean? `os.Remove` needs an empty directory, so that contradicts what you are writing here?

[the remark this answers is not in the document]
didrocks: Seems like you removed it :)

internal/policies/mount/mount.go
didrocks: Ack on the first part, it seems though that the additional test case is not added yet?

denisonbarbosa: It is, isn't it?

didrocks: Hum, I don’t see a test case with a pre-existing mounts file, which is then emptied/removed on ApplyPolicy call? What test case are you talking about?

denisonbarbosa: The test case "mount file is removed on refreshing policy with no entries" falls under that category if I'm not mistaken.

didrocks: ah, you did it with a second call, this is fine then! My last remaining part is checking the golden files, but this is just to double check on what you rereviewed :)

didrocks (approved):
Excellent! It looks good to me :)

Approving

PR #426, opened 2022-09-29
Expose ubuntu pro status

didrocks:
This PR exposes if the client is Ubuntu Pro enabled or not in the "status" command.
We now dynamically fetch the status instead of caching it to not rely on an update to be processed before getting to know our state.
There are some minor misc. cleaning too.

DEENG-455


denisonbarbosa (approved):
Everything as expected. LGTM!

PR #429, opened 2022-09-29
ADMX and ADML generation support for the sharing policy.

denisonbarbosa:
Added the required files to generate the ADM(X&L) templates for the sharing policy for the user.

DEENG-462

### DIFF
--- a/cmd/admxgen/defs/mount.yaml
+++ b/cmd/admxgen/defs/mount.yaml
@@ -1,0 +1,11 @@
+- key: "/mount"
+  displayname: "Mount points for the client"
+  explaintext: |
+    Define mount points that will be mounted for the client.
+    The order of the mounts will be according to the order the mount points were listed in the policy.
+    The mounts listed will be appended to the list of mounts referenced higher in the GPO hierarchy.
+  elementtype: "multiText"
+  release: "any"
+  type: "mounts"
+  meta:
+    strategy: "append"
--- a/cmd/admxgen/defs/mount.yaml
+++ b/cmd/admxgen/defs/mount.yaml
@@ -1,10 +1,9 @@
-- key: "/mount"
+- key: "/user-mounts"
   displayname: "Mount points for the client"
   explaintext: |
     Define mount points that will be mounted for the client.
-    The order of the mounts will be according to the order the mount points were listed in the policy.
+    We will add one mount per line, the order of the mounts will be according to the order the mount points were listed in the policy, removing duplicates.
     The mounts listed will be appended to the list of mounts referenced higher in the GPO hierarchy.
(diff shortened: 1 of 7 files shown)


denisonbarbosa:
Might still need some tweaks (and probably some renaming), but we can discuss it along the PR.

cmd/admxgen/defs/mount.yaml
```diff
@@ -0,0 +1,11 @@
+- key: "/mount"
+  displayname: "Mount points for the client"
+  explaintext: |
+    Define mount points that will be mounted for the client.
+    The order of the mounts will be according to the order the mount points were listed in the policy.
+    The mounts listed will be appended to the list of mounts referenced higher in the GPO hierarchy.
```
didrocks, line 6: Precise that the mounts will be deduplicated. Also that both anynomous mount and those needing kerberos credentials. are handled.

And also, maybe precise the format of the text mount, like protocol:///?
You can add some example like, the anonymous tag and one with a protocol, telling it’s going to try to be authenticated in comments.

didrocks (changes requested):
This is going to be interesting. We are going to have the same key twice which is handled by adsys without any issue as they are in separate `class`. However, the displayname "Drive Mapping" is not typed (I think you decided to have them with different names?). Not a biggie, we can revisit anyway when implementing the maching policy as we are not going to cut a release without it, so no beakage.

(I think we may end up with either different category name or different key name).

I made some suggestion on the text explanation itself, otherwise, this looks good to me and it’s indeed only what is needed.

You probably noticed the PR created automatically against your branch (https://github.com/ubuntu/adsys/pull/430). It looks OK to me :) What we will to do is:
- get the text explanation ok
- wait for https://github.com/ubuntu/adsys/pull/430 to be updated and if happy, merge it to your branch
- then, we can get this branch merged.

Nice work! :)

cmd/admxgen/defs/mount.yaml
```diff
@@ -0,0 +1,41 @@
...
+
+
+
+- key: "/computer-mounts"
+  displayname: "Mount points for the client"
+  explaintext: |
+    Define mount points that will be mounted for the client.
+    We will add one mount per line, the order of the mounts will be according to the order the mount points were listed in the policy, removing duplicates.
+    The mounts listed will be appended to the list of mounts referenced higher in the GPO hierarchy.
+    Anonymous and Kerberos validated mounts will be handled automatically.
```
didrocks, line 29: Are we sure we will be able to handle Kerberos for the computer?

didrocks (commented):
Let’s maybe only stage the user for now and do the computer afterwards?

didrocks:
The new text looks good otherwise :)

didrocks (approved):

PR #431, opened 2022-09-30
Implement apparmor policy manager for machines

GabrielNagy:
The manager currently works additively and only removes rules when the policy is disabled entirely.

Tests to follow.

### DIFF
--- a/internal/policies/apparmor/apparmor.go
+++ b/internal/policies/apparmor/apparmor.go
@@ -1,0 +1,298 @@
+// Package apparmor provides a manager to apply apparmor policies.
+package apparmor
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"sync"
+
+	"github.com/ubuntu/adsys/internal/decorate"
+	log "github.com/ubuntu/adsys/internal/grpc/logstreamer"
+	"github.com/ubuntu/adsys/internal/i18n"
+	"github.com/ubuntu/adsys/internal/policies/entry"
+	"golang.org/x/exp/slices"
+)
+
+// Manager prevents running multiple apparmor update processes in parallel while parsing policy in ApplyPolicy.
+type Manager struct {
+	apparmorDir       string
+	apparmorParserCmd []string
+
+	muMu    sync.Mutex             // protect pathsMu
+	pathsMu map[string]*sync.Mutex // mutex is per destination path (e.g. machine, users/user1, users/user2)
+}
+
+type options struct {
+	apparmorParserCmd []string
+}
+
+// Option reprents an optional function to change the apparmor manager.
+type Option func(*options)
+
+// New creates a manager with a specific apparmor directory.
+func New(apparmorDir string, opts ...Option) *Manager {
+	// defaults
+	args := options{
+		apparmorParserCmd: []string{"apparmor_parser"},
+	}
+	// applied options
+	for _, o := range opts {
+		o(&args)
+	}
+
+	return &Manager{
+		pathsMu:           make(map[string]*sync.Mutex),
+		apparmorDir:       apparmorDir,
+		apparmorParserCmd: args.apparmorParserCmd,
+	}
+}
+
+// AssetsDumper is a function which uncompress policies assets to a directory.
+type AssetsDumper func(ctx context.Context, relSrc, dest string, uid int, gid int) (err error)
+
+// ApplyPolicy generates an apparmor policy based on a list of entries.
+// Steps:
+// 1. Create /etc/apparmor.d/adsys/<object>.new with new policy
+// 2a. Move /etc/apparmor.d/adsys/<object> to /etc/apparmor.d/adsys/<object>.old
+// 2b. Move /etc/apparmor.d/adsys/<object>.new to /etc/apparmor.d/adsys/<object>
+// 3. Run apparmor_parser -r on all files in /etc/apparmor.d/adsys/<object>
+// 4a. If apparmor_parser fails, move /etc/apparmor.d/adsys/<object>.old to /etc/apparmor.d/adsys/<object>
+// 4b. If apparmor_parser succeeds, remove /etc/apparmor.d/adsys/<object>.old.
+func (m *Manager) ApplyPolicy(ctx context.Context, objectName string, isComputer bool, entries []entry.Entry, assetsDumper AssetsDumper) (err error) {
+	defer decorate.OnError(&err, i18n.G("can't apply apparmor policy to %s"), objectName)
+
+	objectDir := "machine"
+	// TODO: add user support
+	if !isComputer {
+		// objectDir = "users"
+		log.Warningf(ctx, i18n.G("apparmor policies are currently only supported for computers"))
+		return nil
+	}
+	apparmorPath := filepath.Join(m.apparmorDir, objectDir)
+
+	// Mutex is per destination path (e.g. machine, users/user1, users/user2)
+	m.muMu.Lock()
+	// if mutex does not exist for this destination, creates it
+	if _, exists := m.pathsMu[apparmorPath]; !exists {
+		m.pathsMu[apparmorPath] = &sync.Mutex{}
+	}
+	m.muMu.Unlock()
+	m.pathsMu[apparmorPath].Lock()
+	defer m.pathsMu[apparmorPath].Unlock()
+
+	// No point in continuing if apparmor isn't available
+	absPath, err := exec.LookPath(m.apparmorParserCmd[0])
+	if err != nil {
+		// If we do have entries to apply we should explicitly fail
+		if len(entries) > 0 {
+			return err
+		}
+		// Otherwise, just let the user know
+		log.Warningf(ctx, "Apparmor is not available on this system: %v", err)
+		return nil
+	}
+	m.apparmorParserCmd[0] = absPath
+
+	// If we have no entries, attempt to unload them and remove the apparmor directory
+	if len(entries) == 0 {
+		return unloadRules(ctx, m.apparmorParserCmd, apparmorPath)
+	}
+
+	log.Debugf(ctx, "Applying apparmor policy to %s", objectName)
+
+	// Return early if we cannot find a valid policy entry
+	idx := slices.IndexFunc(entries, func(e entry.Entry) bool { return e.Key == "apparmor-machine" })
+	if idx == -1 {
+		log.Warning(ctx, i18n.G("No valid entry found for the apparmor machine policy"))
+		return nil
+	}
+
+	if err := os.MkdirAll(apparmorPath, 0750); err != nil {
+		return fmt.Errorf(i18n.G("can't create apparmor directory %q: %v"), apparmorPath, err)
+	}
+
+	// Remove any existing stale directories
+	oldApparmorPath := apparmorPath + ".old"
+	if err := os.RemoveAll(oldApparmorPath); err != nil {
+		return fmt.Errorf(i18n.G("can't remove old apparmor directory %q: %v"), oldApparmorPath, err)
+	}
+	newApparmorPath := apparmorPath + ".new"
+	if err := os.RemoveAll(newApparmorPath); err != nil {
+		return fmt.Errorf(i18n.G("can't remove new apparmor directory %q: %v"), newApparmorPath, err)
+	}
+
+	// Dump assets to the adsys/machine.new/ subdirectory with correct
+	// ownership. If no assets is present while entries != nil, we want to
+	// return an error.
+	if err := assetsDumper(ctx, "apparmor/", newApparmorPath, -1, -1); err != nil {
+		return err
+	}
+
+	// Rename existing apparmor policy to .old
+	if err := os.Rename(apparmorPath, oldApparmorPath); err != nil {
+		return fmt.Errorf(i18n.G("can't rename apparmor directory %q to %q: %v"), apparmorPath, oldApparmorPath, err)
+	}
+	defer cleanupOldApparmorDir(ctx, oldApparmorPath, apparmorPath)
+
+	// Rename new apparmor policy to current
+	if err := os.Rename(newApparmorPath, apparmorPath); err != nil {
+		return fmt.Errorf(i18n.G("can't rename apparmor directory %q to %q: %v"), newApparmorPath, apparmorPath, err)
+	}
+
+	e := entries[idx]
+	// Get the list of files to run apparmor_parser on
+	filesToLoad, err := filesFromEntry(e, apparmorPath)
+	if err != nil {
+		return err
+	}
+
+	// Clean up dumped asset files that are not in the policy entry
+	if err := removeUnusedAssets(apparmorPath, filesToLoad); err != nil {
+		return err
+	}
+
+	if len(filesToLoad) > 0 {
+		// Run apparmor_parser on all profiles
+		apparmorParserCmd := append(m.apparmorParserCmd, "-r")
+		apparmorParserCmd = append(apparmorParserCmd, filesToLoad...)
+
+		// #nosec G204 - We are in control of the arguments
+		cmd := exec.CommandContext(ctx, apparmorParserCmd[0], apparmorParserCmd[1:]...)
+		if out, err := cmd.CombinedOutput(); err != nil {
+			return fmt.Errorf(i18n.G("failed to load apparmor rules: %w\n%s"), err, string(out))
+		}
+	}
+
+	// Loading rules succeeded, remove old apparmor policy dir
+	if err := os.RemoveAll(oldApparmorPath); err != nil {
+		return fmt.Errorf(i18n.G("can't remove old apparmor directory %q: %v"), oldApparmorPath, err)
+	}
+
+	return nil
+}
+
+// cleanupOldApparmorDir handles putting the old apparmor policy files back if the
+// new ones failed to apply.
+func cleanupOldApparmorDir(ctx context.Context, oldApparmorPath, apparmorPath string) {
+	// If the old apparmor directory is not present, it means we succeeded in
+	// applying the new policy
+	if _, err := os.Stat(oldApparmorPath); err != nil && os.IsNotExist(err) {
+		return
+	}
+
+	// Otherwise, we need to restore the old apparmor directory
+	if err := os.RemoveAll(apparmorPath); err != nil {
+		log.Warningf(ctx, "can't remove new apparmor directory: %v", err)
+	}
+
+	if err := os.Rename(oldApparmorPath, apparmorPath); err != nil {
+		log.Warningf(ctx, "can't restore previous apparmor directory: %v", err)
+	}
+}
+
+// unloadRules unloads the apparmor rules in the given directory and removes the directory.
+// No action is taken if the directory doesn't exist.
+func unloadRules(ctx context.Context, apparmorParserCmd []string, apparmorPath string) (err error) {
+	defer decorate.OnError(&err, i18n.G("can't unload apparmor rules"))
+
+	// Nothing to do if the directory doesn't exist
+	if _, err := os.Stat(apparmorPath); err != nil && os.IsNotExist(err) {
+		return nil
+	}
+
+	// Walk the directory and get all the files to unload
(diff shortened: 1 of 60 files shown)


internal/policies/manager.go
```diff
@@ -138,6 +156,7 @@ func NewManager(bus *dbus.Conn, opts ...Option) (m *Manager, err error) {
 		privilege: privilegeManager,
 		scripts:   scriptsManager,
 		gdm:       args.gdm,
+		apparmor:  apparmorManager,
```
didrocks, line 159: this should be before gdm to preserve the order :)

internal/policies/manager.go
```diff
@@ -161,7 +180,9 @@ func (m *Manager) ApplyPolicies(ctx context.Context, objectName string, isComput
 	g.Go(func() error {
 		return m.scripts.ApplyPolicy(ctx, objectName, isComputer, rules["scripts"], pols.SaveAssetsTo)
 	})
-	// TODO g.Go(func() error { return m.apparmor.ApplyPolicy(ctx, objectName, isComputer, rules["apparmor"]) })
+	g.Go(func() error {
```
didrocks, line 183: I think we need to filter the apparmor rule when there is no Ubuntu Pro subscription, though.

internal/policies/apparmor/apparmor.go
```diff
@@ -0,0 +1,262 @@
...
+	}
+	defer cleanupApparmorDir(ctx, oldApparmorPath, apparmorPath)
+
+	// Dump assets to the adsys/machine/ subdirectory with correct ownership. If no assets is present while entries != nil, we want to return an error.
+	if err := assetsDumper(ctx, "scripts/apparmor/", apparmorPath, -1, -1); err != nil {
+		return err
+	}
+
+	// Get the list of files to run apparmor_parser on
+	e := entries[0]
```
didrocks, line 121: I understand why you take the first entry, but I would maybe safeguard this on the key name. Imagine that someone messed with the admx/adml and then, the pol files creates 2 keys, we want to act at least on the well-known name.

internal/policies/apparmor/apparmor.go
```diff
@@ -0,0 +1,262 @@
...
+// removeUnusedAssets removes all files/directories in the given directory that
+// are not in the given list of files.
+func removeUnusedAssets(apparmorPath string, filesToKeep []string) error {
+	return filepath.WalkDir(apparmorPath, func(path string, d os.DirEntry, err error) error {
+		if !d.Type().IsRegular() {
+			return nil
+		}
+
+		// Remove files that are not in the policy entry
+		if !slices.Contains(filesToKeep, filepath.Clean(path)) {
```
didrocks, line 240: Let’s reverse the logic to align more left:
```
if <positive> {
  continue
}
// the rest of the code for removal
```

internal/policies/apparmor/apparmor.go
```diff
@@ -0,0 +1,262 @@
...
+		}
+
+		filesToLoad = append(filesToLoad, filepath.Clean(scriptFilePath))
+	}
+	return filesToLoad, nil
+}
+
+// removeUnusedAssets removes all files/directories in the given directory that
+// are not in the given list of files.
+func removeUnusedAssets(apparmorPath string, filesToKeep []string) error {
```
didrocks, line 233: I would decorate the error here and just return directly the err in each branch, as the error should be descriptive enough.

internal/policies/apparmor/apparmor.go
```diff
@@ -0,0 +1,262 @@
...
+			parentDir := filepath.Dir(path)
+			parentEntries, err := os.ReadDir(parentDir)
+			if err != nil {
+				return fmt.Errorf(i18n.G("can't read directory %q: %v"), parentDir, err)
+			}
+			if len(parentEntries) != 0 {
+				return nil
+			}
+
+			if err := os.Remove(parentDir); err != nil {
```
didrocks, line 255: and with this, you can just return os.Remove here.

internal/policies/apparmor/apparmor.go
```diff
@@ -0,0 +1,262 @@
...
+// It returns an error if the file does not exist or is a directory.
+func filesFromEntry(e entry.Entry, apparmorPath string) ([]string, error) {
+	var filesToLoad []string
+	for _, script := range strings.Split(e.Value, "\n") {
+		script = strings.TrimSpace(script)
+		if script == "" {
+			continue
+		}
+
+		scriptFilePath := filepath.Join(apparmorPath, script)
```
didrocks, line 217: is the script name intended (apparmor terminology for this is script?)

internal/policies/apparmor/apparmor.go
```diff
@@ -0,0 +1,262 @@
...
+
+	// Loading rules succeeded, remove old apparmor policy dir
+	if err := os.RemoveAll(oldApparmorPath); err != nil {
+		return fmt.Errorf(i18n.G("can't remove old apparmor directory %q: %v"), oldApparmorPath, err)
+	}
+
+	return nil
+}
+
+// cleanupApparmorDir handles putting the old apparmor policy files back if the
```
didrocks, line 150: `cleanupOldApparmorDir`?

didrocks (commented):
Some comments and food for thoughts, but overall, this is looking good! :)
Well done!

internal/policies/apparmor/apparmor.go
```diff
@@ -110,15 +110,21 @@ func (m *Manager) ApplyPolicy(ctx context.Context, objectName string, isComputer
...
-	defer cleanupApparmorDir(ctx, oldApparmorPath, apparmorPath)
+	defer cleanupOldApparmorDir(ctx, oldApparmorPath, apparmorPath)
 
 	// Dump assets to the adsys/machine/ subdirectory with correct ownership. If no assets is present while entries != nil, we want to return an error.
-	if err := assetsDumper(ctx, "scripts/apparmor/", apparmorPath, -1, -1); err != nil {
+	if err := assetsDumper(ctx, "apparmor/", apparmorPath, -1, -1); err != nil {
 		return err
 	}
 
+	idx := slices.IndexFunc(entries, func(e entry.Entry) bool { return e.Key == "apparmor-machine" })
```
didrocks, line 120: We reall have a fan of x/slices I see :p

internal/policies/apparmor/apparmor.go
```diff
@@ -110,15 +110,21 @@ func (m *Manager) ApplyPolicy(ctx context.Context, objectName string, isComputer
...
+	if idx == -1 {
+		log.Warning(ctx, i18n.G("No valid entry found for the apparmor machine policy"))
+		return nil
+	}
+
+	e := entries[idx]
```
didrocks, line 126: ok, I think we can get it that way for now and same for users. It may not be future-proof, but as we expect always one key for machine, and one for users, this is ok.

didrocks:
The changes looks good. It seems that you only missed the change to reduce the window without any valid profile or am I wrong?

internal/policies/apparmor/apparmor.go
```diff
@@ -0,0 +1,262 @@
...
+// AssetsDumper is a function which uncompress policies assets to a directory.
+type AssetsDumper func(ctx context.Context, relSrc, dest string, uid int, gid int) (err error)
+
+// ApplyPolicy generates an apparmor policy based on a list of entries.
+// Steps:
+// 1. Move /etc/apparmor.d/adsys/<object> to /etc/apparmor.d/adsys/<object>.old
+// 2. Create /etc/apparmor.d/adsys/<object> with new policy
+// 3. Run apparmor_parser -r on all files in /etc/apparmor.d/adsys/<object>
+// 4a. If apparmor_parser fails, move /etc/apparmor.d/adsys/<object>.old to /etc/apparmor.d/adsys/<object>
+// 4b. If apparmor_parser succeeds, remove /etc/apparmor.d/adsys/<object>.old.
```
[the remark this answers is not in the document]
didrocks: Exactly!

internal/policies/apparmor/apparmor.go
```diff
@@ -0,0 +1,268 @@
...
+
+type options struct {
+	apparmorParserCmd string
+}
+
+// Option reprents an optional function to change the apparmor manager.
+type Option func(*options)
+
+// New creates a manager with a specific apparmor directory.
+func New(apparmorDir string, opts ...Option) (m *Manager, err error) {
```
denisonbarbosa, line 38: Right now, there's no way for this function to return an error. Even though this keeps the same signature as the other managers.New funcs, I think it's best to take a look at it and, if the function can't fail in any way, it shouldn't expect to return an error.

didrocks: I have no strong opinion either way on this. Both approach works for me (being future proof and return an error, or no error being returned and be able to update both API and package manager tests as needed). As this is an internal API, I would leave that up to your decision :)

I wouldn’t check `apparmor_parser` executable there yet, as you say, we aren’t sure we have a policy to apply, and the executable may vanish between the manager creation and the `ApplyPolicy` call.

internal/policies/apparmor/apparmor.go
```diff
@@ -222,14 +245,18 @@ func filesFromEntry(e entry.Entry, apparmorPath string) ([]string, error) {
 
 		profileFilePath := filepath.Join(apparmorPath, profile)
 		info, err := os.Stat(profileFilePath)
-		if errors.Is(err, os.ErrNotExist) {
-			return nil, fmt.Errorf(i18n.G("apparmor profile %q doesn't exist in SYSVOL apparmor/ subdirectory"), profile)
+		if err != nil {
```
didrocks, line 248: hum, we can have other kinds of errors that triggers this behavior (can’t read/stat because of parent directory), maybe the returned error should say "no accessible" rather than does not exist?

internal/policies/apparmor/apparmor.go
```diff
@@ -222,14 +245,18 @@ func filesFromEntry(e entry.Entry, apparmorPath string) ([]string, error) {
...
+			return nil, fmt.Errorf(i18n.G("apparmor profile doesn't exist in SYSVOL apparmor/ subdirectory: %w"), err)
 		}
 		if info.IsDir() {
 			return nil, fmt.Errorf(i18n.G("apparmor profile %q is a directory and not a file"), profile)
 		}
 
-		filesToLoad = append(filesToLoad, filepath.Clean(profileFilePath))
+		cleanProfilePath := filepath.Clean(profileFilePath)
+		if slices.Contains(filesToLoad, cleanProfilePath) {
```
didrocks, line 256: I would add a comment telling how we are filtering the file.

internal/policies/apparmor/apparmor_test.go
```diff
@@ -0,0 +1,223 @@
...
+		"existing .old directory is removed": {entries: defaultMachineProfile, destAlreadyExists: "machine.old", noParserOutput: true},
+		"existing .new directory is removed": {entries: defaultMachineProfile, destAlreadyExists: "machine.new", noParserOutput: true},
+
+		// shared cases
+		"no profiles, existing rules are removed": {destAlreadyExists: "machine"},
+		"no profiles, apparmor directory absent":  {noParserOutput: true},
+		"unexpected entry key":                    {entries: []entry.Entry{{Key: "apparmor-foo", Value: "usr.bin.foo"}}, noParserOutput: true},
+
+		// user cases
+		"user, one profile": {entries: []entry.Entry{{Key: "apparmor-user", Value: "usr.bin.foo"}}, user: true, noParserOutput: true},
```
didrocks, line 55: hum, you are testing the user without the implementation? :p

internal/policies/apparmor/apparmor_test.go
```diff
@@ -0,0 +1,223 @@
...
+func mockApparmorParserCmd(t *testing.T, parserOutputFile string, args ...string) []string {
+	t.Helper()
+
+	cmdArgs := []string{"env", "GO_WANT_HELPER_PROCESS=1", os.Args[0], "-test.run=TestMockApparmorParser", "--", parserOutputFile}
+	cmdArgs = append(cmdArgs, args...)
+	return cmdArgs
+}
+
+func TestMockApparmorParser(t *testing.T) {
+	if os.Getenv("GO_WANT_HELPER_PROCESS") != "1" {
```
didrocks, line 170: Niceeeeeeeeeeeeeee, I think it’s the first time you are doing one, right? I remember having presented you this pattern, but this is great to see!

internal/policies/apparmor/apparmor_test.go
```diff
@@ -0,0 +1,223 @@
...
+			}
+			if tc.readOnlyApparmorDir != "" {
+				testutils.MakeReadOnly(t, filepath.Join(apparmorDir, tc.readOnlyApparmorDir))
+			}
+			sat := sat{err: tc.saveAssetsError}
+
+			m := apparmor.New(apparmorDir, apparmor.WithApparmorParserCmd(apparmorParserCmd))
+			err := m.ApplyPolicy(context.Background(), "ubuntu", !tc.user, tc.entries, sat.mockSaveAssetsTo)
+			if tc.wantErr {
+				// We don't return here as we want to check that the apparmor
```
didrocks, line 113: thanks for the comment :p

internal/policies/apparmor/apparmor_test.go
```diff
@@ -0,0 +1,223 @@
...
+				require.Error(t, err, "ApplyPolicy should have failed but didn't")
+			} else {
+				require.NoError(t, err, "ApplyPolicy failed but shouldn't have")
+			}
+
+			// Restore permissions to be able to correctly compare trees
+			if tc.readOnlyApparmorDir != "" {
+				// nolint:gosec //false positive, this is a directory
+				err = os.Chmod(filepath.Join(apparmorDir, tc.readOnlyApparmorDir), 0700)
+				require.NoError(t, err, "can't chmod apparmor dir")
```
didrocks, line 124: I would add a Setup: prefix here or Teardown, as this is not what we are really testing (same comments for other error checks below)

internal/policies/apparmor/apparmor_test.go
```diff
@@ -0,0 +1,223 @@
...
+					shutil.CopyTree(
+						filepath.Join("testdata", "ApplyPolicy", "apparmor_dir", object), filepath.Join(apparmorDir, tc.destAlreadyExists),
+						&shutil.CopyTreeOptions{Symlinks: true, CopyFunction: shutil.Copy}),
+					"Setup: can't create initial apparmor dir profiles content")
+				fmt.Println(filepath.Join(apparmorDir, tc.destAlreadyExists))
+			}
+			if tc.readOnlyApparmorDir != "" {
+				testutils.MakeReadOnly(t, filepath.Join(apparmorDir, tc.readOnlyApparmorDir))
+			}
+			sat := sat{err: tc.saveAssetsError}
```
didrocks, line 108: what does sat stands for? (I had to read it twice to understand that this is the asset dumper)

internal/policies/apparmor/apparmor_test.go
```diff
@@ -0,0 +1,223 @@
...
+		"computer, one profile":              {entries: defaultMachineProfile},
+		"computer, multiple profiles,":       {entries: []entry.Entry{{Key: "apparmor-machine", Value: "usr.bin.foo\nusr.bin.bar\nnested/usr.bin.baz"}}},
+		"computer, duplicated profiles":      {entries: []entry.Entry{{Key: "apparmor-machine", Value: "usr.bin.foo\nusr.bin.foo"}}},
+		"computer, blank line profiles":      {entries: []entry.Entry{{Key: "apparmor-machine", Value: "usr.bin.foo\n\nusr.bin.bar\n"}}},
+		"computer, only blank profiles":      {entries: []entry.Entry{{Key: "apparmor-machine", Value: "\n\n\n"}}},
+		"existing .old directory is removed": {entries: defaultMachineProfile, destAlreadyExists: "machine.old", noParserOutput: true},
+		"existing .new directory is removed": {entries: defaultMachineProfile, destAlreadyExists: "machine.new", noParserOutput: true},
+
+		// shared cases
+		"no profiles, existing rules are removed": {destAlreadyExists: "machine"},
```
didrocks, line 50: and I was going to write about the missing removal test with a pre-existing directory and I see that you did it. Well done :p (ofc, once we have the partial removal, you will handle this too)

internal/policies/apparmor/apparmor_test.go
```diff
@@ -0,0 +1,223 @@
...
+		noApparmorParser    bool
+
+		wantErr bool
+	}{
+		// computer cases
+		"computer, one profile":              {entries: defaultMachineProfile},
+		"computer, multiple profiles,":       {entries: []entry.Entry{{Key: "apparmor-machine", Value: "usr.bin.foo\nusr.bin.bar\nnested/usr.bin.baz"}}},
+		"computer, duplicated profiles":      {entries: []entry.Entry{{Key: "apparmor-machine", Value: "usr.bin.foo\nusr.bin.foo"}}},
+		"computer, blank line profiles":      {entries: []entry.Entry{{Key: "apparmor-machine", Value: "usr.bin.foo\n\nusr.bin.bar\n"}}},
+		"computer, only blank profiles":      {entries: []entry.Entry{{Key: "apparmor-machine", Value: "\n\n\n"}}},
```
didrocks, line 45: You are missing some cases where you have spaces to be trimmed inside the profile. (and that a blank line only with spaces is handled)

internal/policies/apparmor/apparmor_test.go
```diff
@@ -0,0 +1,223 @@
...
+			testutils.CompareTreesWithFiltering(t, apparmorDir, filepath.Join("testdata", "ApplyPolicy", "golden", "trees", testutils.NormalizeGoldenName(t, name)), update)
+
+			// Return early if we don't want to check apparmor_parser output for
+			// whatever reason (e.g. command did not execute, returned an error before etc.)
+			if tc.noParserOutput {
+				return
+			}
+
+			// Check that apparmor_parser was called with the expected arguments
+			goldPath := filepath.Join("testdata", "ApplyPolicy", "golden", testutils.NormalizeGoldenName(t, name))
```
didrocks, line 135: I would organize things a little bit differently:
- golden/<name of the test>
 and under it:
  - we have the output result from the parser command (with a relevant file name, always the same between tests)
  - then, same kind of fake `etc/apparmor.d` instead of "trees" so that we have an idea where, by default, we project it on the hard drive (if that makes sense), or just `apparmor.d` at least or whatever this is "IRL" :)

didrocks (commented):
Really nicely done! I like the testing part a lot (and made some useless comments just to say I noticed :p)

There are few remaining comments, mostly nitpick.

TBH, I have to trust you on the apparmor resulting part, because you clearly have more experience now that I have on this, so nested/not nested and various cases, it’s all on you :o) I’m happy to have a deep look if you prefer though, but that would take some investigation time.

Finally, have you tried to look at the coverage and see if there are no easy case of uncovered error returned that we could trigger?

Again, very well done. I’m impressed.

internal/policies/apparmor/apparmor_test.go
didrocks: yep!

internal/policies/apparmor/apparmor_test.go
didrocks: I was sure this was on me :p But yeah, probably SaveAssetsTo, mind renaming (and so, same in scripts, even if it’s not really part of that PR but does it worth its own PR?) to "mockAssetDumper" or anything similar?

didrocks: if we extract it somewhere, it will be specific test helper for managers (it doesn’t make sense out of the current manager context I think?)

didrocks (commented):
Just one remaining (minor, but it’s a question of preference and I think in test maintainability on the long term) comment!

I looked a the coverage, and indeed, the error cases that are left uncovered are ones we can’t easily trigger without pausing the program, changing things (permissions/layout) under it and resuming.

My only nitpick is that we don’t cover any error case at all of `removeUnusedAssets`. We may achieve it by controlling the asset dumper to chmod a particular directory (so that we then can’t remove files under it) in one test case, maybe? Worth exploring if you don’t mind. (one of the "good" rule is to, at least cover one error returned by function, so that we ensure we pick it in the calling function and not ignore it :p)

PR #436, opened 2022-10-07
Fixes for integration tests

denisonbarbosa:
A little ground work to help integrating the new managers into the integration tests and improve code consistency.


didrocks (changes requested):
Are you sure that the files are the ones that you expect? It seems that you updated the golden, but the existing state should be updated too to match what’s in the GPO, right?

Like the up to date state should match the already updated machine?
And old state should have other scripts, old?

I suggest that you update those, and look then very carefully that the golden files are reflecting what you expect.

The code change themselves (renaming the errors) is ok, but to me, this PR only does half of what we discussed…

didrocks:
Also, there are some discrepancies between old_state for user cache and the projected golden files, this is why digging and looking carefully at each test case for integration tests is needed. Matching the input with the expected output.

This is why I’m renewing my offering of taking this PR on me and do this work carefully.

didrocks (approved):
Test pass!

PR #439, opened 2022-10-07
Integration tests for the mount manager

denisonbarbosa:
Initial integration of the mount manager into the integration tests setup. More tests are still required, but those will be added after the mounts handler is implemented.
This branch was rebased on #436, so it should be dealt with before merging this branch.

DEENG-460


cmd/adsysd/integration_tests/adsysctl_policy_test.go
```diff
@@ -594,6 +595,25 @@ func TestPolicyUpdate(t *testing.T) {
 			},
 			wantErr: true,
 		},
+		// This is will make the scripts fail as well, sadly. But currently there's
```
denisonbarbosa, line 598: This is a little annoying. We could "fix" by restoring a run/users/uid directory with a mounts file and then make the file read only, rather than using localhost-uptodate or by changing the way we save the mounts file. 
Wdyt?

didrocks:
Here is the reason I wanted to have both PR separated: as this one is based on the other one, I can’t easily rereview it before the other one is merged.

I thought though that this wasn’t necessary? As it’s only a question of renaming tests (that we don’t have to do here) + old state/golden files, the files added don’t conflict between each other, and so, this branch can just branch main separately and the conflict should be minimal, if not existent at all.

Please really try to use separate branches as a separation of concern.

cmd/adsysd/integration_tests/adsysctl_policy_test.go
didrocks: Isn’t that the goal of `readOnlyDirs` that you are using? (Could be extended to files, but having the directory RO itself should still fail the mount manager AND scripts btw)

denisonbarbosa: Well, yes. But what I meant is that by making the run directory read only, we are making the scripts manager fail as well. Would be better if there was a way to make the mounts manager fail individually rather than always affecting both the mounts and the scripts managers.

didrocks: Could we try in one test to create the mounts file preventively and make it RO?

denisonbarbosa: I think we can change the way we go about readOnly dirs to make it work. We can try to create the specified path (if it doesn't exist) before making it read only. This would also give us more flexibility in the tests as well, I think.

denisonbarbosa:
PR will be redone in another branch

PR #440, opened 2022-10-10
Integration tests for mount manager

denisonbarbosa:
Updating integration tests to cover the recently added mount manager for users policies.


cmd/adsysd/integration_tests/adsysctl_policy_test.go
```diff
@@ -594,6 +594,14 @@ func TestPolicyUpdate(t *testing.T) {
 			},
 			wantErr: true,
 		},
+		"Error on mount apply failing": {
+			initState: "localhost-uptodate",
+			readOnlyDirs: []string{
+				// In order to make the mount manager fail, we need create a not empty mounts directory inside the user folder.
+				fmt.Sprintf("run/users/%s/mounts", currentUID),
```
denisonbarbosa, line 601: It might be a little weird to rely on readOnlyDirs to create the directory that we need for the failure.
I can add another field in the test case struct to handle it, if you think it's better.

denisonbarbosa: Well, not really. We already saw that, when writing in ext4 'safe mode', Go does not care if we replace a file with readOnly permissions and that the only way  to make the mount manager fail (when it comes to permissions) is to have a directory called mounts, rather than a file with readOnly permissions. So if we initial state already have a mounts file, we can't make the mount manager fail (at least for now), unless we remove the file and replace it with a non empty directory.

didrocks: So, the comment is not correct, it’s "we need to change the file by a non empty RO directory". Does it really need to be non empty btw?

I would still like if we can minimize the amount of parameters, if possible at all. However, the new logic may change the other failure, no?

cmd/adsysd/integration_tests/adsysctl_policy_test.go
```diff
@@ -830,6 +842,14 @@ func TestPolicyUpdate(t *testing.T) {
 				err := os.RemoveAll(filepath.Join(adsysDir, k))
 				require.NoError(t, err, "Setup: could not remove generate assets db")
 			}
+
+			// Some tests will need some additional paths to be created
+			for _, k := range tc.addPaths {
+				testutils.CreatePath(t, filepath.Join(adsysDir, k)+"/")
+				_, err := os.Stat(filepath.Join(adsysDir, k))
```
didrocks, line 849: hum, why so?  Doesn't CreatePath fails if we couldn’t create it?

cmd/adsysd/integration_tests/adsysctl_policy_test.go
```diff
@@ -596,6 +597,17 @@ func TestPolicyUpdate(t *testing.T) {
 			},
 			wantErr: true,
 		},
+		"Error on mount apply failing": {
+			initState: "old-data",
+			clearDirs: []string{
+				fmt.Sprintf("run/users/%s/mounts", currentUID),
+			},
+			// This generates an error when the used path already exists as a directory
```
didrocks, line 605: + "instead of a file" or something like that in the description

cmd/adsysd/integration_tests/adsysctl_policy_test.go
```diff
@@ -225,6 +225,7 @@ func TestPolicyUpdate(t *testing.T) {
 		krb5ccname            string
 		krb5ccNamesState      []krb5ccNamesWithState
 		clearDirs             []string // Removes already generated system files eg dconf db, apparmor profiles, ...
+		addPaths              []string
```
didrocks, line 228: shouldn’t that be `addDir` (as we always append / to the path name)

cmd/adsysd/integration_tests/testdata/PolicyUpdate/states/old-data/cache/policies/userintegrationtest@example.com/policies
```diff
@@ -28,6 +28,12 @@ gpos:
             old-dir/old-other-script
           disabled: false
           strategy: append
-  - id: "{31B2F340-016D-11D2-945F-00C04FB984F9}"
-    name: Default Domain Policy
-    rules: {}
+      mount:
+        - key: user-mounts
+          value: |
+            protocol://example.com/old/old-mount
```
didrocks, line 34: I would differentiate the 2 users mount to not be exactly the same.

didrocks (changes requested):
I haven’t checked all the test case yet as you will need to reupdate some.

But in addition, I spotted some inconsistency at least on one test (this is why you need to recheck all the test cases, carefully everytime):
"Current user, update old data" indeed, update the current user. However, the other user doesn’t have the old .mount file. This is inconsistent with the old policy cache for the other user which list one policy. Looking up, it seems that old_state doesn’t have a mount file for this other user, which is then the reason why.

Once those are fixed + the other comments, I will rereview again all test case, but please, do it preventively :) (and don’t start from the golden file, but rather from the test, to assert if a test doesn’t update anything that this is indeed the expected outcome).

cmd/adsysd/integration_tests/adsysctl_policy_test.go
denisonbarbosa: I think we can do both ways. I can either change the name to be addDirs or I could make it a []struct{path, isDir} and we append / to the path if isDir is true. Wdyt?
For some reason, filepath.Join trims the last / that it's on the string, so if we want to be able to handle directories and files, we must do something like the []struct[}

didrocks: I think addDir is the more consice and declares precisely what we do here.

denisonbarbosa: It is, indeed. But is less versatile as well...

didrocks: Right, but the question is: do we need versatility right now? Do we have other use cases matching another need where this field should be versatile? Is it part of any public API making this design decision set in stone forever?

If thhe answer to the 3 questions is "no" (spoiler alert: it is!), then let’s not plan for a future that may (and probably will) never happen. :)

cmd/adsysd/integration_tests/adsysctl_policy_test.go
didrocks: Hum, I see +1 from you here, but that wasn’t changed?

denisonbarbosa: Should be fixed now

didrocks (approved):
All golden are now good. You can merge the intergation tests!

PR #444, opened 2022-10-12
User mount handler

denisonbarbosa:
Implementation of the Rust binary responsible to read the mounts file and mount the parsed entries for the current user.

DEENG-463


internal/policies/mount/adsys_mount/src/error.rs
```diff
@@ -0,0 +1,17 @@
+use thiserror::Error;
+
+/// Our own error type
+///
+/// It's not good practice to return errors that are not from your own API
+/// (unless they are from the std library, i.e. io or fmt errors).
+#[non_exhaustive] // Blocks match expressions without wildcards (_) for this enum.
+#[derive(Debug, Error)]
+pub enum AdysMountError {
```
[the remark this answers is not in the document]
denisonbarbosa: Oh, yeah... I totally knew about it, I was just testing you guys...
😅

internal/policies/mount/adsys_mount/src/main.rs
```diff
@@ -0,0 +1,235 @@
+use clap::Parser;
+use gio::{
+    self,
+    traits::{FileExt, MountOperationExt},
+};
+use glib::ObjectExt;
```
didrocks, line 6: is it used anywhere?

internal/policies/mount/adsys_mount/src/main.rs
```diff
@@ -0,0 +1,235 @@
...
+
+    // Attaches the receiver to the main context, along with a closure that is called everytime there is a new message in the channel.
+    rx.attach(Some(&g_ctx), move |x| {
+        match x.status {
+            MountStatus::Done => debug!("Mounting of {} was successful", x.path),
+            MountStatus::Error(_) => {
+                warn!("Failed when mounting {}", x.path);
+                mu_clone.lock().unwrap().push(x);
+            }
+            _ => {}
```
didrocks, line 95: I would probably still catch the MountStatus::Ask status here to return an error that was not expected…

internal/policies/mount/adsys_mount/src/main.rs
```diff
@@ -0,0 +1,235 @@
...
+
+    if had_error {
+        return Err(AdsysMountError::MountError);
+    }
+    Ok(())
+}
+
+/// Reads the file and parses the mount points listed in it.
+fn parse_entries(path: &String) -> Result<Vec<MountEntry>, std::io::Error> {
+    debug!("Parsing file content");
```
didrocks, line 134: I would add some context to the debug, like the path of the file we are currently parsing

internal/policies/mount/adsys_mount/src/main.rs
```diff
@@ -0,0 +1,235 @@
...
+
+    let mut parsed_entries: Vec<MountEntry> = Vec::new();
+
+    // The ? operator tries to unwrap the result and, if there is an error, returns it to the caller of this function.
+    let content = fs::read_to_string(path)?;
+
+    for p in content.split_terminator('\n') {
```
didrocks, line 141: IIRC, there are other ways to iterate over lines, like `content.lines()`

internal/policies/mount/adsys_mount/src/main.rs
```diff
@@ -0,0 +1,235 @@
...
+/// Callback that is invoked by gio when prompted for password.
+fn ask_password_cb(
+    mount_op: &gio::MountOperation,
+    _: &str,
+    _: &str,
+    _: &str,
+    flags: gio::AskPasswordFlags,
+) {
+    if mount_op.is_anonymous() && flags.contains(gio::AskPasswordFlags::ANONYMOUS_SUPPORTED) {
+        unsafe {
```
didrocks, line 212: hum, why do we have this with no comment? :p

Anyway to avoid calling unsafe?

internal/policies/mount/adsys_mount/src/main.rs
```diff
@@ -0,0 +1,235 @@
...
+            if let Some(data) = mount_op.data("state") {
+                // Ensures that we only try anonymous access once.
+                if let MountStatus::Asked = *(data.as_ptr()) {
+                    warn!("Anonymous access denied.");
+                    mount_op.reply(gio::MountOperationResult::Aborted);
+                }
```
didrocks, line 218: can we return early and avoid too many identations?

internal/policies/mount/adsys_mount/src/main.rs
```diff
@@ -0,0 +1,235 @@
...
+            } else {
+                debug!("Anonymous is supported by the provider.");
+                mount_op.set_data("state", MountStatus::Asked);
+                mount_op.reply(gio::MountOperationResult::Handled);
+            }
+        }
+    // Checks if the machine has a kerberos ticket defined.
+    } else if std::env::var("KRB5CCNAME").is_ok() {
```
didrocks, line 226: yeah, ditto, if we can avoid the else if…

internal/policies/mount/adsys_mount/src/main.rs
```diff
@@ -0,0 +1,235 @@
...
+    }
+
+    if had_error {
+        return Err(AdsysMountError::MountError);
+    }
+    Ok(())
+}
+
+/// Reads the file and parses the mount points listed in it.
+fn parse_entries(path: &String) -> Result<Vec<MountEntry>, std::io::Error> {
```
didrocks, line 133: I think you can safely use a &str, thanks to implicit coerxion, you don’t even need to typecast.

It will make some change where you can pass directly the slice reference in the test.

internal/policies/mount/adsys_mount/src/test.rs
```diff
@@ -0,0 +1,30 @@
+#[cfg(test)]
+mod tests {
+    use crate::{parse_entries, MountEntry};
+
+    /// Tests parse_entries function.
+    #[test]
+    fn test_parse_entries() {
```
didrocks, line 7: I think there are multiple tests that we can do, like empty files, and so on… (also, file that doesn’t exist). That way, you can test the Err case which is not covered right now.

While I’m at it… Do you mind looking if we can transform those in some kind of table testing? (Honest request, I haven’t looked how to do that in rest, but at worst, we can have a for loop basically, without real subtest support, and relax the assert to a global error?)

didrocks (commented):
Here we go (finally!) I have some comments and some exploration that I think you can have a look to. Feel free to tell me if anything is not clear

internal/policies/mount/adsys_mount/src/main.rs
denisonbarbosa: Yes... There are some "methods" that require traits to be imported

internal/policies/mount/adsys_mount/src/main.rs
didrocks: I still see the split terminator functino here

didrocks (commented):
something to look at.

internal/policies/mount/adsys_mount/src/test.rs
```diff
@@ -44,7 +56,11 @@ mod tests {
             );
 
             match want {
-                Ok(w) => assert_eq!(w, format!("{:?}", got)),
+                Ok(w) => {
+                    for i in 0..w.len() {
+                        assert_eq!(w[i], got[i]);
```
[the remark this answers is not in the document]
schopin-pro: Presumably you could do `assert_eq!(w, got)` since there's this: https://doc.rust-lang.org/std/vec/struct.Vec.html#impl-Eq-for-Vec%3CT%2C%20A%3E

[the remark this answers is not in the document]
didrocks: funnily, we adressed it an hour ago in a HO :)

internal/policies/mount/adsys_mount/src/lib.rs
```diff
@@ -0,0 +1,220 @@
...
+fn ask_password_cb(
+    mount_op: &gio::MountOperation,
+    _: &str,
+    _: &str,
+    _: &str,
+    flags: gio::AskPasswordFlags,
+) {
+    if mount_op.is_anonymous() && flags.contains(gio::AskPasswordFlags::ANONYMOUS_SUPPORTED) {
+        // Unsafe block is needed for data and set_data implementations in glib.
+        unsafe {
```
schopin-pro, line 193: I tend to follow this policy regarding unsafe blocks:

https://std-dev-guide.rust-lang.org/documentation/safety-comments.html

Usually, `unsafe fn foo()` will have a comment describing why it is unsafe, i.e. which invariants need to be upheld by the calling code. Here, `set_data()` and `data()` have indeed a "Safety" section: https://docs.rs/glib/latest/glib/object/trait.ObjectExt.html#tymethod.set_data

> This function doesn’t store type information

> The returned pointer can become invalid by a call to set_qdata, steal_qdata, set_data or steal_data.
>
> The caller is responsible for ensuring the returned value is of a suitable type

A key principle in Rust is local reasoning: you should be able to figure stuff out at the function level. This is true for type inference (all types can be deduced from either local variable declarations or deducted from the function signature), but also for safety reasoning.

Here, the only way to make sure the invariant is upheld (we only store booleans in there, and nowhere else is `set_data` called) is to grep through the whole codebase, which means that your function `ask_password_cb` should be marked `unsafe`. The place where you can enforce that invariant is actually when you connect the callback, as there you have a grasp over the `MountOperation` lifecycle, know that it's only used locally (and assume that `mount_enclosing_volume` don't do nasty stuff with `set_data`, but that should be documented as well).

So, in short, this is what I'm proposing:

```rust
/// Handles the mount operation to mount the specified entry.
fn mount_entry(entry: Entry, tx: glib::Sender<Msg>) {
    debug!("Mounting entry {}", entry.path);

    let f = gio::File::for_uri(&entry.path);

    let mount_op = gio::MountOperation::new();

    if entry.is_anonymous {
        debug!("Anonymous mount requested for {}", entry.path);
        mount_op.set_anonymous(true);
    }

    mount_op.connect_ask_password(|mount_op, _, _, _, flags| {
        if mount_op.is_anonymous() && flags.contains(gio::AskPasswordFlags::ANONYMOUS_SUPPORTED) {
            // SAFETY: no use of `set_data` and friends on the local mount_op, assuming
            // `mount_enclosing_volume` also doesn't.
            unsafe { ensure_single_anonymous_access(mount_op) };
            return;
        }
        // Checks if the machine has a kerberos ticket defined.
        if std::env::var("KRB5CCNAME").is_ok() {
            debug!("Kerberos ticket found on the machine.");
            mount_op.reply(gio::MountOperationResult::Handled);
            return;
        }

        warn!("Kerberos ticket not available on the machine.");
        mount_op.reply(gio::MountOperationResult::Aborted);
    });

    // Callback invoked by gio after setting up the mount.
    let mount_handled_cb = move |r: Result<(), glib::Error>| {
        let msg = Msg {
            path: entry.path,
            status: r,
        };

        if let Err(e) = tx.send(msg) {
            error!("Failed to send message in the channel: {}", e);
        }
    };

    f.mount_enclosing_volume(
        gio::MountMountFlags::NONE,
        Some(&mount_op),
        gio::Cancellable::NONE,
        mount_handled_cb,
    );
}

/// # Safety
///
/// This function assumes it has exclusive use of `mount_op`'s `set_data()` and related methods.
unsafe fn ensure_single_anonymous_access(mount_op: &gio::MountOperation) {
    // SAFETY: per function invariant
    if let Some(data) = unsafe { mount_op.data::<bool>("state") } {
        // Ensures that we only try anonymous access once.
        // SAFETY: per function invariant
        if unsafe { *data.as_ptr() } { // could also have been `if *(unsafe { data.as_ref() })` but that's rather ugly
            warn!("Anonymous access denied.");
            mount_op.reply(gio::MountOperationResult::Aborted);
        }
    } else {
        debug!("Anonymous is supported by the provider.");
        // SAFETY: per function invariant
        unsafe { mount_op.set_data("state", true) };
        mount_op.reply(gio::MountOperationResult::Handled);
    }
}
```

Note that in Rust < 1.65 this will give you warnings, as the community practices around `unsafe` are still evolving. Previously, being in a `unsafe fn` not only meant that the function couldn't be called without `unsafe` but also that its contents could use unsafe operations without further `unsafe` blocks, and so an old lint considered them redundant. To disable these old warnings (and actually warn if *not* using an `unsafe` block), add `#[warn(unsafe_op_in_unsafe_fn)]` at the top of `lib.rs`.

schopin-pro (commented):
There's a whole thing regarding `unsafe`, and another smaller one regarding tests (I know, I know...)

didrocks:
ok, so apart from my comment on https://github.com/ubuntu/adsys/pull/549 + the fact that the package doesn’t build for me (missing `XS-Vendored-Sources-Rust` field in debian/control), this will be ready for merging.

Please fix the package build, we will assess for the other comment.

schopin-pro:
> (missing `XS-Vendored-Sources-Rust` field in debian/control)

That part is actually deliberate. I haven't found a satisfying way to
magically generate this for you in a way that's also enforceable at
build time, so I'm just prompting the packager with the actual line to
add to their d/control.

I did not add said line to d/control as it's fairly ugly (1500
characters) and depends on the state of the vendored tree, which itself
isn't vendored. With hindsight, that last argument is kinda moot since
the vendored tree is generated from the lockfile, which *is* versioned.

didrocks:
ok, I think I’m done with different PRs around this one. See the details on each one. In addition to those:

I think it’s worth exploring a way to not remove the cargo lock file (see the exchange on the packaging PR)
as we discussed yesterday, some CI to update the X- vendored deps rust will update to fix the FTBS in kinetic+ and having it always up to date.
and finally, update debian/copyright (yes, I know, this is exciting…)

Cargo.toml
```diff
@@ -0,0 +1,7 @@
+[workspace]
+
+members = [
+    "internal/policies/mount/adsys_mount",
+]
+
+exclude = [ "vendor_rust/" ]
```
liushuyu, line 7: I would suggest enabling LTO for the release profile by default so that the binary could be even more optimized and smaller in size.

```suggestion
exclude = [ "vendor_rust/" ]

[profile.release]
lto = "fat"
```

internal/policies/mount/adsys_mount/src/lib.rs
```diff
@@ -0,0 +1,225 @@
...
+                if *(data.as_ref()) {
+                    warn!("Anonymous access denied.");
+                    mount_op.reply(gio::MountOperationResult::Aborted);
+                }
+            } else {
+                debug!("Anonymous is supported by the provider.");
+                mount_op.set_data("state", true);
+                mount_op.reply(gio::MountOperationResult::Handled);
+            }
+        }
```
liushuyu, line 209: I highly recommend limiting the `unsafe` scope so that only the statements requiring `unsafe` are inside the `unsafe` block, making it more obvious which operations may be prone to issues.

```suggestion
        // SAFETY: The usage of set_data and data can result in type unsafety. Since we are
        // controlling which key is being set and by whom, this should be safe.
        if let Some(data) = unsafe { mount_op.data::<bool>("state") } {
            // Ensures that we only try anonymous access once.
            if *(data.as_ref()) {
                warn!("Anonymous access denied.");
                mount_op.reply(gio::MountOperationResult::Aborted);
            }
        } else {
            debug!("Anonymous is supported by the provider.");
            mount_op.set_data("state", true);
            mount_op.reply(gio::MountOperationResult::Handled);
        }
```

internal/policies/mount/adsys_mount/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+include = ["src/**/*"]
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+[lib]
+name = "lib"
+path = "src/lib.rs"
+
+[dependencies]
+anyhow = "1.0.65"
+chrono = "0.4.22"
```
liushuyu, line 15: `chrono` is unmaintained. I recommend switching to [`time`](https://lib.rs/crates/time) crate:

```suggestion
time = { version = "0.3", features = ["formatting", "macros"] }
```

internal/policies/mount/adsys_mount/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+gio = "0.16.2"
+glib = "0.16.2"
+log = "0.4.17"
+serde = { version = "1.0.147", features = ["derive"] }
+serde_json = "1.0.87"
+thiserror = "1.0.37"
```
liushuyu, line 21: I recommend referencing their major version for crates with stabilized versions (according to the semantic versioning). Like this:

```suggestion
serde = { version = "^1", features = ["derive"] }
serde_json = "^1"
thiserror = "^1"
```

PR #447, opened 2022-10-13
Switch to new fsnotify event check syntax

GabrielNagy:
This was added in v1.6.0 of the library which we just updated to, and it makes for a more readable way of checking for events.


didrocks (approved):
That makes it way easier to read. good work!

PR #461, opened 2022-10-18
Add apparmor tests to policies manager

GabrielNagy:
This involves a bit of argument restructuring because in order to have a passing apparmor manager run we need control over both the `apparmor_parser` command and the loaded policies file.

The current (and best) way to do it for now is to use functional options. This will also aid us when updating the integration test to
take into account the apparmor manager.

I've repurposed the `loadedPoliciesFile` to be configurable via the apparmorfs directory which is its parent directory and exposed the configurable options to the policies manager.

For tests, the `apparmor_parser` command is stubbed to `/bin/true` for convenience, and similar to the other managers we set the directory as read-only in order to trigger a failure.


didrocks:
@denisonbarbosa mind reviewing this first? Also, do you mind investigating Gabriel’s comment?

denisonbarbosa (changes requested):
Looks good to me, apart from some comments that we can iterate over (specially the consts one)

didrocks (approved):
This LGTM! On the permission change, I think the helper used to create as 0755 before we went to gosec and change them. Permission change is typically not taken into account by git add, so I’m surprised this was taken. However, as discussed, this will be decouple in another PR.

I agree about the bubbling up of options which is annoying. I have that in my mind for quite some time but didn’t think about an idiomatic solution yet. I capture it as https://github.com/ubuntu/adsys/issues/465 for now.

Nice work! Nothing change requested :)

PR #464, opened 2022-10-18
Add integration tests for apparmor computer policy manager

GabrielNagy:
I've opened this against https://github.com/ubuntu/adsys/pull/461 since it depends on some functional options magic from there. I separated the updates in easy to follow commits to make for an easier review.

### DIFF
--- a/cmd/adsysd/integration_tests/adsysctl_policy_test.go
+++ b/cmd/adsysd/integration_tests/adsysctl_policy_test.go
@@ -464,6 +464,7 @@ func TestPolicyUpdate(t *testing.T) {
 				"sudoers.d",
 				"polkit-1",
 				"run/machine",
+				"apparmor.d/adsys/machine",
 			},
 			krb5ccNamesState: []krb5ccNamesWithState{
 				{
@@ -484,6 +485,7 @@ func TestPolicyUpdate(t *testing.T) {
 				"sudoers.d",
 				"polkit-1",
 				"run/machine",
+				"apparmor.d/adsys/machine",
 				"cache/sysvol/Policies/{C4F393CA-AD9A-4595-AEBC-3FA6EE484285}",
 				"cache/sysvol/Policies/{31B2F340-016D-11D2-945F-00C04FB984F9}",
 			},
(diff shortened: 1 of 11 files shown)


didrocks:
Can you force the refresh on latest main? (I don’t know how to force GH doing this)

cmd/adsysd/integration_tests/adsysctl_policy_test.go
```diff
@@ -947,6 +962,7 @@ func TestPolicyUpdate(t *testing.T) {
 			testutils.CompareTreesWithFiltering(t, filepath.Join(adsysDir, "dconf"), filepath.Join("testdata", "PolicyUpdate", "golden", goldName, "dconf"), update)
 			testutils.CompareTreesWithFiltering(t, filepath.Join(adsysDir, "sudoers.d"), filepath.Join("testdata", "PolicyUpdate", "golden", goldName, "sudoers.d"), update)
 			testutils.CompareTreesWithFiltering(t, filepath.Join(adsysDir, "polkit-1"), filepath.Join("testdata", "PolicyUpdate", "golden", goldName, "polkit-1"), update)
+			testutils.CompareTreesWithFiltering(t, filepath.Join(adsysDir, "apparmor.d", "adsys"), filepath.Join("testdata", "PolicyUpdate", "golden", goldName, "apparmor.d", "adsys"), update)
```
denisonbarbosa, line 965: Maybe we can move the comparison one level higher? Rather than comparing apparmor.d/adsys, it would compare apparmor.d/ (along with everything nested in it).

denisonbarbosa (changes requested):

didrocks (changes requested):
Good work, how is the headache? :p

It was almost perfect, there are just some few changes needed, but this should not impact the golden files, which is the good news!

I have to write 3 things to change here, as it’s parts missing in the diff:
* the old-state should have the apparmor assets corresponding to the projected apparmor.d directory. Basically, it means you need to update `PolicyUpdate/states/old-data/cache/sysvol/assets/` too.
* in adsysctl_policy_test.go, if you look at the offline test cases, we ensure that we can regenerate from old state + policy cache the directories on the system. This is done by removing some directories before applying policies. I didn’t see the apparmor.d directories being changed.
* `localhost-uptodate/cache/sysvol/assets` should have the content corresponding to assets.db (for the sake of consistency)

If you have any question, feel free to reach out to me. Excellent work (as usual, I would say).

didrocks (approved):
perfect! Really great work again. Thanks!

denisonbarbosa (approved):

PR #468, opened 2022-10-19
DEENG-487 Update scripts manager creation and golden files consistency

didrocks:
Ensure we always have a machine directory created by the script manager (it should be similar to all managers using those directories).
Fix the localhost up to date state in integration tests to always have the empty user directory. This is consistent with the managers creation state. Refresh the golden files for them.


didrocks:
No, you are completely right, I just rebased and remove those changes. Forced push.

PR #474, opened 2022-10-20
Compare golden tree executable permissions

GabrielNagy:
The `testutils.CompareTreesWithFiltering` helper knows to write golden files with the expected permissions when the `-update` flag is given.

However, these permissions are not enforced in the comparison so we could get in the case where `go test` passes, and `go test -update` causes changes in the permissions of the golden files.

Because git only tracks executable bits, only check for this when doing the comparison, and represent it as a boolean to make it easier to read. This produces an output like the following:

    Diff:
    --- Expected
    +++ Actual
    @@ -119,3 +119,3 @@
       path: (string) (len=35) "/run/adsys/machine/scripts/.running",
    -  executable: (bool) true
    +  executable: (bool) false
      },
    Test:       	TestApplyPolicies/succeed
    Messages:   	got and expected content differs

Due to how the diff is shown the output might get noisy when files have large contents.


didrocks (approved):
One minor thing to change, and then, you can merge! I made some tests fails and indeed, the output looks good, excellent :)

PR #475, opened 2022-10-20
Rework skipping integration tests

### LINKED ISSUE #469: go test ./... should pass with non permissive user
As per our discussion, we should skip the tests that needs root, and write in the skip comment that it needs the env variable to run.

ofc, CI and package build (if we can run them: build and autopkgtests) needs to be updated.

GabrielNagy:
With the addition of the Windows codebase we needed to skip certain tests on Linux if we did not have root permissions. An `ADSYS_SKIP_SUDO_TESTS` flag was added for this.

Additionally, some tests proved to be problematic when ran with `-race` on Windows. Thus, we added `ADWATCHD_SKIP_INTEGRATION_TESTS` to account for this.

Besides this, adsys already had `ADSYS_SKIP_INTEGRATION_TESTS` before which was used to skip integration tests during Debian packaging and autopkgtests.

This led to, among other things, tests failing if one ran go test ./... without root privileges on Linux. To simplify the situation:
- automatically skip tests requiring root and log an appropriate message
- drop ADSYS_SKIP_SUDO_TESTS as it's accounted by the change above
- rename ADWATCHD_SKIP_INTEGRATION_TESTS to ADSYS_SKIP_INTEGRATION_TESTS in order to have a single environment variable controlling this

Closes #469 / DEENG-509

### DIFF
--- a/.github/workflows/qa.yaml
+++ b/.github/workflows/qa.yaml
@@ -74,7 +74,7 @@ jobs:
         run: |
           set -eu
 
-          ADSYS_SKIP_SUDO_TESTS=1 go test -coverpkg=./... -coverprofile=/tmp/coverage.out -covermode=count ./...
+          go test -coverpkg=./... -coverprofile=/tmp/coverage.out -covermode=count ./...
 
           # Run integration tests that need sudo
           # Use command substitution to preserve go binary path (sudo does not preserve path even with -E)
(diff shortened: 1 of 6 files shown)


didrocks (approved):
I was going to (heavily of course), complain that by removing `ADSYS_SKIP_SUDO_TESTS` you were removing the semantic by making all under the `INTEGRATION_TEST` umlbrella which was untrue… But that was just by reading the comment before looking at the diff.

Then I saw you added the `SkipUnlessRoot` and my life changed to a happy place :)

Very nice work, and I confirm this works both locally, building package and autopkgtests (I only read what we do for autopkgtests, didn’t battle test it, but if there is any issue, we can fix it).

So, we still have `ADSYS_SKIP_ROOT_CALLS` which is, unfortunately, still inside the code itself, but it’s an internal env variable. Maybe we should have a think about it at some point.

Approving!

PR #479, opened 2022-10-24
Allow running mount_handler tests as part of go test

didrocks:
Hook up running rust tests from go test command, with proper detection and cache handling.

### DIFF
--- a/internal/policies/mount/adsys_mount/run_test.go
+++ b/internal/policies/mount/adsys_mount/run_test.go
@@ -1,0 +1,36 @@
+// Package adsysmount_test runs rust tests from a go test global command
+package adsysmount_test
+
+import (
+	"os"
+	"os/exec"
+	"testing"
+
+	"github.com/ubuntu/adsys/internal/testutils"
+)
+
+func TestRust(t *testing.T) {
+	t.Parallel()
+
+	// properly inform Go about which files to use for cache by reading input files.
+	testutils.MarkRustFilesForTestCache(t)
+
+	cmd := exec.Command("cargo", "test")
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+
+	if err := cmd.Run(); err != nil {
+		t.Fail()
+	}
+}
+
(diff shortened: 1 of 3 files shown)


internal/policies/mount/adsys_mount/run_test.go
```diff
@@ -0,0 +1,36 @@
...
+	}
+}
+
+func TestMain(m *testing.M) {
+	canRun, withCoverage := testutils.CanRunRustTests()
+	if !canRun {
+		os.Exit(1)
+	}
+
+	_ = withCoverage
```
denisonbarbosa, line 33: I know this is being set up for later, but can't we just leave at ```canRun, _ := testutils.CanRunRustTests()``` plus a comment informing that coverage will be "coming soon"?

internal/policies/mount/adsys_mount/run_test.go
```diff
@@ -0,0 +1,36 @@
...
+
+	// properly inform Go about which files to use for cache by reading input files.
+	testutils.MarkRustFilesForTestCache(t)
+
+	cmd := exec.Command("cargo", "test")
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+
+	if err := cmd.Run(); err != nil {
+		t.Fail()
```
denisonbarbosa, line 23: Shouldn't we log the error here as well (even though we are already setting the cmd stds to the os ones)?

denisonbarbosa (changes requested):
Looks good to me, just a few questions about some stuff

internal/policies/mount/adsys_mount/run_test.go
didrocks: No really, I tried it with failing tests and then, the error message makes sense: you see that this test fails, and you get the old stdout/stderr printed above, which is what makes sense and is in line with pure Go tests.

didrocks:
Merged your suggestion and answered your comment

denisonbarbosa (approved):
LGTM! Feel free to merge.

PR #482, opened 2022-10-25
Add ADMX/ADML definitions for apparmor user policy

GabrielNagy:
DEENG-472

### DIFF
--- a/policies/Ubuntu/all/Ubuntu.adml
+++ b/policies/Ubuntu/all/Ubuntu.adml
@@ -27,6 +27,7 @@
       <string id="UbuntuDisplayPowerManagement">Power Management</string>
       <string id="UbuntuDisplaySessionManagement">Session management</string>
       <string id="UbuntuDisplayUserScripts">User Scripts</string>
+      <string id="UbuntuDisplayUserApplicationConfinement">User application confinement</string>
       <string id="UbuntuDisplayUserDriveMapping">User Drive Mapping</string>
       <string id="UbuntuExplainTextUserDconfOrgGnomeDesktopInterfaceToolkitAccessibility">Whether toolkits should load accessibility related modules.
 
@@ -821,7 +822,7 @@ Note: -
 An Ubuntu Pro subscription on the client is required to apply this policy.</string>
       <string id="UbuntuDisplayMachineAllPrivilegeAllowLocalAdmins">Allow local administrators</string>
       <string id="UbuntuExplainTextMachineScriptsStartup">Define scripts that are executed on machine boot, once the GPO is downloaded.
-Those scripts are ordered, one by line, and relative to SYSVOL/scripts/ubuntu/ directory.
+Those scripts are ordered, one by line, and relative to SYSVOL/ubuntu/scripts/ directory.
 Scripts from this GPO will be appended to the list of scripts referenced higher in the GPO hierarchy.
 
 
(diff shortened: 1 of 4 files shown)


cmd/admxgen/defs/apparmor.yaml
```diff
@@ -16,3 +16,37 @@
...
+  displayname: "AppArmor"
+  explaintext: |
+    Define an AppArmor user profile to be parsed and loaded on client machines.
+    The profile is specified as a file path relative to the SYSVOL/apparmor/ubuntu/ directory.
+    On the client machine, user profiles are stored in /etc/apparmor.d/adsys/users/<user-name>, thus the administrator can reference abstractions and tunables shipped with the client distribution of AppArmor.
+
+    The profile will ideally contain a mapping between a user and a role. Roles must be configured beforehand in the System-wide application confinement section.
+    Below is an example of a user profile declaration:
+
+      ^{{ADSYS_USER}} {
```
didrocks, line 29: As this will be copied by us on disk and always stats with the same, should we, to avoid errors, only request to put the content of the template itself, and we encapsulate it with `user { … }`?

cmd/admxgen/defs/apparmor.yaml
```diff
@@ -16,3 +16,37 @@
...
+
+    Where {{ADSYS_USER}} is a special string that will be replaced with the client username when adsys applies the policy, and default_user is a role declared beforehand in the Machine section. More details and examples can be found in the apparmor section of the adsys documentation.
+
+    The configured profile will override any profile referenced higher in the GPO hierarchy.
+  elementtype: "text"
+  release: "any"
+  note: |
+   -
+    * Enabled: The profile in the text entry is applied on the client machine.
+    * Disabled: The profile is removed from the target machine, and any related rules are unloaded.
```
didrocks, line 51: unconfigured vs Disabled, is there any change of behaviour? (seems that if Disabled and parent will configured something, we won’t have it compared to only unconfigured. Is it worth mentionning?)

cmd/admxgen/defs/apparmor.yaml
```diff
@@ -16,3 +16,37 @@
   type: "apparmor"
   meta:
     strategy: append
+- key: "/apparmor-users"
+  displayname: "AppArmor"
+  explaintext: |
+    Define an AppArmor user profile to be parsed and loaded on client machines.
+    The profile is specified as a file path relative to the SYSVOL/apparmor/ubuntu/ directory.
```
didrocks, line 23: `SYSVOL/ubuntu/apparmor` rather?

didrocks (commented):
some small comments waiting for your opinon! :)

cmd/admxgen/defs/apparmor.yaml
didrocks: Can we have some magic post-processing the file maybe? Like supporting {} maybe if the user wanted it? However, here, we really need to make something generic as you did with the templated name…

cmd/admxgen/defs/apparmor.yaml
didrocks: with inheritance, I’m really sure that if you have a "closer" GPO as unconfigured, it’s just ignored and the parent wins… Where disable will override the parent.

The unconfigured case works if all of them are unconfigured.

cmd/admxgen/defs/apparmor.yaml
didrocks: pfffffffffffffff details details :p (sure, feel free to fix!)

didrocks (approved):
LGTM now!

PR #487, opened 2022-10-27
Rust tests with coverage and CI

didrocks:
* Hook up running rust tests from go test command, with proper detection and cache handling.
* Add Rust -> Go coverage convertion support
 Similarly to python, add helpers for Rust code coverage.
 Refactor some helper functions to minimize the API surface and make it more idiomatic.
* Add coverage to Rust mount handler
 In addition to call the coverage file, we need to mark src/ as a Go package for the Go coverage tool picking it up.

Note that this has been tested with Rust modules in subdirectories of src/ too.

DEENG-513

### DIFF
--- a/.github/workflows/qa.yaml
+++ b/.github/workflows/qa.yaml
@@ -4,7 +4,7 @@ on:
     branches:
       - main
     tags:
-      - '*'
+      - "*"
   pull_request:
 
 jobs:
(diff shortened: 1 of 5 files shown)


denisonbarbosa (approved):
Looks like we're finally having Rust coverage... Very nice job!

didrocks:
Thanks to both of you for the testing and happy that the error messages were clear enough to get it setup :)

> I see that we are relying on TEST_RUST_TARGET to reuse an existing build directory. Why don't we let cargo decide the directory on its own? This way we'd benefit from incremental builds implicitly (I think). I'm sure you've already thought of this but wanted to be sure :P

Yeah, I’m happy to revisit the decision. By default, the target is target/ in the current directory, meaning that when we run `go test ./...` (or just run `cargo test` in the rust-code directory), we then ends up with some artifacts in target/, which may be committed and so on. I’m not a big fan of files being generated in the tree itself when simply running tests (the Go artifacts are generated in a temporary directory, the cache though is shared in the user's HOME for reuse). It doesn’t seem that Rust/Cargo as a user-centralized cache, but it’s project by project.

To avoid generating those artifacts, this is why I decided to rely on a target in a temporary directory. Still, this is not ideal in an development environment (even if I think in general, we will use `cargo test` directly instead of `go test`, hence this environment variable option to reuse something.

If you can come up with a better idea, my hears are all opened :) (this is not set in stone either, so we can merge and revisit)

PR #489, opened 2022-11-02
Session dbus mock

denisonbarbosa:
Setting up the session dbus mock to be used in the integration tests


didrocks (approved):
Excellent work! The live review was fun too :) Approving!

PR #499, opened 2022-11-16
Handle missing version key in GPT.INI

### LINKED ISSUE #492: Invalid GPT.INI: version not found (prevents login if adsys is enabled in PAM)
It would appear that at some point, AD group policy ini file can omit "Version=" tag. 
We have one of this kind of policy, domain default policy that has never been edited and is totally empty.
Windows doesn't care. Group policy editor shows 0 as version. But Ubuntu 22 with ADsys will not let users log in.

Error message:
ERROR Error from server: error while updating policy: can't get policies for "user@domain.tld": can't download all gpos and assets: one or more error while fetching GPOs and assets: can't download "Default Domain Policy": can't check if Default Domain Policy needs refreshing: invalid remote GPT.INI: version not found

using pam-auth-update and deselecting ADsys Authentication lets user in, but ofc without policies being applied.

GabrielNagy:
As per #492, having a GPT.INI file with no Version key seems to be allowed. Instead of failing hard in this scenario, log a warning message and return 0 for the version.

Additionally, rewrite the getGPOVersion function using the go-ini library.

Closes #492 / DEENG-527

### DIFF
--- a/internal/ad/download.go
+++ b/internal/ad/download.go
@@ -157,9 +157,8 @@ func (ad *AD) fetch(ctx context.Context, krb5Ticket string, downloadables map[st
 
 var errNoGPTINI = errors.New("no GPT.INI file")
 
-// needsDownload erturn if the downloadable should be refreshed.
+// needsDownload returns if the downloadable should be refreshed.
 // This is done by comparing GPT.INI Version= content.
-// allowMissing is used to allow the function to return false if the file is missing.
 func needsDownload(ctx context.Context, client *libsmbclient.Client, g *downloadable, localPath string) (updateNeeded bool, err error) {
 	defer decorate.OnError(&err, i18n.G("can't check if %s needs refreshing"), g.name)
 


internal/ad/download.go
```diff
@@ -173,7 +171,7 @@ func needsDownload(ctx context.Context, client *libsmbclient.Client, g *download
 	if f, err := os.Open(filepath.Clean(gptIniPath)); err == nil {
 		defer decorate.LogFuncOnErrorContext(ctx, f.Close)
 
-		if localVersion, err = getGPOVersion(f); err != nil {
+		if localVersion, err = getGPOVersion(ctx, f, g.name); err != nil {
 			log.Warningf(ctx, "Invalid local GPT.INI for %s: %v\nDownloading it again…", g.name, err)
```
didrocks, line 175: There is something I don’t get, we did error out from what we saw on the report, however, it seems we were only printing some warnings there?

didrocks: nicely spotted! :)

didrocks (approved):
Excellent work! Also, happy that we had an existing use-case with the error that you just needed to convert :)

Just one thought: do you think we should have an additional debug on getGPOVersion then?
Also, as we have a fallback assuming 0, I would downgrade the warning to 0 (meaning that anyway, we are going to download).

Coverage is lower, but this is only because the number of lines is less compared to uncovered.

internal/ad/download.go
```diff
@@ -196,22 +192,26 @@ func needsDownload(ctx context.Context, client *libsmbclient.Client, g *download
 	return true, nil
 }
 
-func getGPOVersion(r io.Reader) (version int, err error) {
+func getGPOVersion(ctx context.Context, r interface{}, downloadableName string) (version int, err error) {
```
didrocks, line 195: I’m not a big fan of changing this to an interface{} to support both filename or reader. It means that we don’t control the contract of what we accept in our function.

My suggestion would be then to always take a reader, readall the file in that function and call Load with the slice of bytes, which, reading from the source code, it seems to accept. That way, we don’t leak the lack of contract that ini.Load() imposes on us. wdyt?

didrocks:
All good then!

Issue #500, opened 2022-11-16
Stabilize integration test coverage


didrocks:
We have an issue when the coverage is called more than once in the same test. Prevent that from happening again and deal with the python case.

PR #501, opened 2022-11-16
Stabilize integration test coverage

didrocks:
* Prevent coverage tracking to be called twice in the same test
* Annotate python coverage file in a subdirectory
* Move python file tracking per test

DEENG-531
Fixes: #500

### DIFF
--- a/internal/testutils/pythoncoverage.go
+++ b/internal/testutils/pythoncoverage.go
@@ -88,7 +88,7 @@ exec python3-coverage run -a %s $@
 		coverDir := pythonCoverageFile + ".annotated"
 		// #nosec G204 - we have a const for coverageCmd
 		out, err := exec.Command(coverageCmd, "annotate", "-d", coverDir, "--include", tracedFile).CombinedOutput()
-		require.NoErrorf(t, err, "Teardown: can’t combine python coverage: %v", string(out))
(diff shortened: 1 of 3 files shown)


internal/testutils/pythoncoverage.go
```diff
@@ -84,8 +84,8 @@ exec python3-coverage run -a %s $@
 			return
 		}
 
-		// Convert to text format
-		coverDir := filepath.Dir(testGoCoverage)
+		// Convert to text format in a subdirectory named after the python coverage file.
+		coverDir := pythonCoverageFile + ".annotated"
 		// #nosec G204 - we have a const for coverageCmd
 		out, err := exec.Command(coverageCmd, "annotate", "-d", coverDir, "--include", tracedFile).CombinedOutput()
 		require.NoErrorf(t, err, "Teardown: can’t combine python coverage: %v", string(out))
```
denisonbarbosa, line 91: Weirdly enough, I only have a comment on a line that wasn't changed...
Can't we just do ```require.NoError(t, err, "Teardown: can’t combine python coverage: %s", out)```? It looks more idiomatic, I think.

didrocks: good one. Still using NoErrorf though.

denisonbarbosa (approved):
LGTM!

PR #504, opened 2022-11-17
Integration tests user mount handler

denisonbarbosa:
Integration tests for the user mount handler and their helper files

DEENG-464

### DIFF
--- a/cmd/adsysd/integration_tests/adsys_mount_test.go
+++ b/cmd/adsysd/integration_tests/adsys_mount_test.go
@@ -27,10 +27,14 @@ func TestUserMountHandler(t *testing.T) {
 		"mount successfully smb share": {mountsFile: "mounts_with_smb_entry"},
 
 		// Anonymous entries
-		"mount successfully one anonymous nfs entry": {mountsFile: "mounts_with_anonymous_nfs_entry"},
-		"mount successfully one anonymous smb entry": {mountsFile: "mounts_with_anonymous_smb_entry"},
+		"mount successfully anonymous nfs entry":                         {mountsFile: "mounts_with_anonymous_nfs_entry"},
+		"mount successfully anonymous smb entry":                         {mountsFile: "mounts_with_anonymous_smb_entry"},
+		"mount successfully anonymous nfs entry without kerberos ticket": {mountsFile: "mounts_with_anonymous_nfs_entry", noKrbTicket: true},
+		"mount successfully anonymous smb entry without kerberos ticket": {mountsFile: "mounts_with_anonymous_smb_entry", noKrbTicket: true},
 
 		// Many entries
+		"mount successfully many entries with nfs protocol":        {mountsFile: "mounts_with_many_nfs_entries"},
+		"mount successfully many entries with smb protocol":        {mountsFile: "mounts_with_many_smb_entries"},
 		"mount successfully many entries with different protocols": {mountsFile: "mounts_with_many_entries"},
 		"mount successfully many anonymous entries":                {mountsFile: "mounts_with_many_anonymous_entries"},
 
@@ -42,14 +46,14 @@ func TestUserMountHandler(t *testing.T) {
 		"error when file doesn't exist":              {mountsFile: "do_not_exist", wantStatus: 1},
 
 		// Authentication errors
-		"error when trying to mount smb without kerberos ticket":   {mountsFile: "mounts_with_smb_entry", noKrbTicket: true, wantStatus: 1},
 		"error when trying to mount nfs without kerberos ticket":   {mountsFile: "mounts_with_nfs_entry", noKrbTicket: true, wantStatus: 1},
-		"error when anonymous auth is not supported by the server": {mountsFile: "mounts_with_anonymous_smb_entry", sessionAnswer: "anonymous_error", noKrbTicket: true, wantStatus: 1},
+		"error when trying to mount smb without kerberos ticket":   {mountsFile: "mounts_with_smb_entry", noKrbTicket: true, wantStatus: 1},
+		"error when anonymous auth is not supported by the server": {mountsFile: "mounts_with_anonymous_smb_entry", sessionAnswer: "gvfs_anonymous_error", noKrbTicket: true, wantStatus: 1},
 
 		// Bus errors
-		"error when VFS bus is not available": {sessionAnswer: "no_vfs_bus", wantStatus: 1},
-		"error during ListMountableInfo step": {sessionAnswer: "list_info_fail", wantStatus: 1},
-		"error during MountLocation step":     {sessionAnswer: "mount_loc_fail", wantStatus: 1},
+		"error when VFS bus is not available": {sessionAnswer: "gvfs_no_vfs_bus", wantStatus: 1},
+		"error during ListMountableInfo step": {sessionAnswer: "gvfs_list_info_fail", wantStatus: 1},
+		"error during MountLocation step":     {sessionAnswer: "gvfs_mount_loc_fail", wantStatus: 1},
 
 		// Generic errors
 		"error when trying to mount unsupported protocol": {mountsFile: "mounts_with_unsupported_protocol", wantStatus: 1},
--- a/cmd/adsysd/integration_tests/adsys_mount_test.go
+++ b/cmd/adsysd/integration_tests/adsys_mount_test.go
@@ -42,8 +42,9 @@ func TestUserMountHandler(t *testing.T) {
 		"error when file doesn't exist":              {mountsFile: "do_not_exist", wantStatus: 1},
 
 		// Authentication errors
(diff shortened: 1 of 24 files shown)


cmd/adsysd/integration_tests/adsysctl_policy_test.go
```diff
@@ -821,8 +821,7 @@ func TestPolicyUpdate(t *testing.T) {
 			if tc.systemAnswer == "" {
 				tc.systemAnswer = "polkit_yes"
 			}
-			systemAnswer(t, tc.systemAnswer)
-			testutils.PythonCoverageToGoFormat(t, filepath.Join(rootProjectDir, "internal/ad/adsys-gpolist"), true)
```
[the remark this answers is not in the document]
denisonbarbosa: Not at all.

didrocks (changes requested):
Some first round of feedbacks.

Also, I think we should figure out the coverage part (see my hints on MM) to ensure we have coverage reported, even with a single binary build, but the many runs covers them. As told, I think this should be ok once you are using the correct hook. Please have some try!

Also, there are lacking quite some failure cases we discussed during the sprint (for instance, the callback failures on dbus mock, so that we check we behave accordingly in the daemon). This is the reason why I’m not going to refresh the docker image yet.

cmd/adsysd/integration_tests/adsys_mount_test.go
```diff
@@ -0,0 +1,91 @@
...
+
+	binDir := os.Getenv("TEST_RUST_TARGET")
+	if binDir == "" {
+		binDir = t.TempDir()
+	}
+	binPath := setupBinaryForTests(t, binDir)
+
+	tests := map[string]struct {
+		mountsFile    string
+		sessionAnswer string
```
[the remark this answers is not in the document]
denisonbarbosa: Yes, I'm just not sure how many more do we need. I could create a failure for each step of the process, but I'm not sure if this is overkill or it's needed.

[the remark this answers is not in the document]
didrocks: I think that we should have least of those with dbus call failed (we don’t have any tests for it), also, signal not sent but method returned.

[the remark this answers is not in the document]
denisonbarbosa: I went for it. Created options for no vfs bus, failures during list mount info and failures during mount location

cmd/adsysd/integration_tests/adsys_mount_test.go
```diff
@@ -0,0 +1,108 @@
...
+		// Single entries
+		"mount successfully nfs share": {mountsFile: "mounts_with_nfs_entry"},
+		"mount successfully smb share": {mountsFile: "mounts_with_smb_entry"},
+
+		// Anonymous entries
+		"mount successfully one anonymous nfs entry": {mountsFile: "mounts_with_anonymous_nfs_entry"},
+		"mount successfully one anonymous smb entry": {mountsFile: "mounts_with_anonymous_smb_entry"},
+
+		// Many entries
+		"mount successfully many entries with different protocols": {mountsFile: "mounts_with_many_entries"},
```
didrocks, line 34: I think the many entry with same protocols first is another use case so that we can see if the issue is due to many or just one protocol.

cmd/adsysd/integration_tests/adsys_mount_test.go
```diff
@@ -0,0 +1,108 @@
...
+		"exit code 0 when file is empty": {mountsFile: "mounts_with_no_entries"},
+
+		// File errors
+		"error when file has badly formated entries": {mountsFile: "mounts_with_bad_entries", wantStatus: 1},
+		"error when file doesn't exist":              {mountsFile: "do_not_exist", wantStatus: 1},
+
+		// Authentication errors
+		"error when trying to mount smb without kerberos ticket":   {mountsFile: "mounts_with_smb_entry", noKrbTicket: true, wantStatus: 1},
+		"error when trying to mount nfs without kerberos ticket":   {mountsFile: "mounts_with_nfs_entry", noKrbTicket: true, wantStatus: 1},
+		"error when anonymous auth is not supported by the server": {mountsFile: "mounts_with_anonymous_smb_entry", sessionAnswer: "anonymous_error", noKrbTicket: true, wantStatus: 1},
```
didrocks, line 47: I think we can have one more case (above), like "mount successfully with anynomous auth without kerberos ticket", so that this completes those tests.

cmd/adsysd/integration_tests/adsys_test.go
```diff
@@ -317,15 +317,21 @@ func changeAppArgs(t *testing.T, s setterArgs, conf string, args ...string) {
...
 		"polkit_yes",
 		"polkit_no",
 		"no_startup_time",
 		"invalid_startup_time",
 		"no_nextrefresh_time",
 		"invalid_nextrefresh_time",
 		"subscription_disabled",
+
+		// This values will only influence the session bus
+		"no_vfs_bus",
```
didrocks, line 331: can we prefix them all with `gfvs_`?

(I think the others not prefixed will need to be prefix with `systemd` in a separate PR to make more sense too now that this is bigger)

cmd/adsysd/integration_tests/systemdaemons/vfs_mocks.py
```diff
@@ -48,3 +81,20 @@ def Mounted(self, bus_name: str, obj_path: dbus.ObjectPath, display_name: str, s
...
+    def get_spec_values(self, protocol: str, mount_spec: tuple[list[bytes], dict[str, list[bytes]]]) -> tuple[str, str]:
+        server, share = '', ''
+
+        if protocol == 'smb-share':
+            server = ''.join([dbus.String(x) for x in mount_spec[1].get("server")])
+            share = ''.join([dbus.String(x) for x in mount_spec[1].get("share")])
+
+        elif protocol == 'nfs':
+            server = ''.join([dbus.String(x) for x in mount_spec[1].get("host")])
+            share = ''.join([dbus.String(x) for x in mount_spec[0]])
```
didrocks, line 94: Mind writing some comments for smb-share and nfs so that we know what mount_spec looks like? I’m puzzled that for smb-share we always take index 1, and for the share of nfs, it’s index 0, so writing them as comment would really help :)

cmd/adsysd/integration_tests/systemdaemons/vfs_mocks.py
```diff
@@ -48,3 +81,20 @@ def Mounted(self, bus_name: str, obj_path: dbus.ObjectPath, display_name: str, s
...
+
+        # Removes the null characters from the string
+        share = share.replace('\0', '')
```
didrocks, line 97: maybe `strip()` would work there?

didrocks (changes requested):
Nice work! Here and there, I think there are some room for improvements and places we can extend the tests, otherwise overall really good :)

cmd/adsysd/integration_tests/systemdaemons/vfs_mocks.py
```diff
@@ -85,16 +85,32 @@ def Mounted(self, bus_name: str, obj_path: dbus.ObjectPath, display_name: str, s
     def get_spec_values(self, protocol: str, mount_spec: tuple[list[bytes], dict[str, list[bytes]]]) -> tuple[str, str]:
         server, share = '', ''
 
+        # Mount spec for smb-share when mounting smb://example.com/mount/path
+        # mount_spec [
+        #  "/path" + \0,
+        #  dict {
+        #    "server": "example.com" + \0,
+        #    "share":  "mount" + \0,
+        #     "type":  "smb-share" + \0,
```
didrocks, line 94: nitpick: alignement

cmd/adsysd/integration_tests/systemdaemons/vfs_mocks.py
```diff
@@ -85,16 +85,32 @@ def Mounted(self, bus_name: str, obj_path: dbus.ObjectPath, display_name: str, s
...
+        #  }
         elif protocol == 'nfs':
             server = ''.join([dbus.String(x) for x in mount_spec[1].get("host")])
             share = ''.join([dbus.String(x) for x in mount_spec[0]])
 
-        # Removes the null characters from the string
-        share = share.replace('\0', '')
-        server = server.replace('\0', '')
+        # Removes the trailing null characters from the strings
+        share = share.strip('\0')
```
didrocks, line 113: No need for argument I think

denisonbarbosa: Yes, we do.The python documentation states "The chars argument is a string specifying the set of characters to be removed. If omitted or None, the chars argument defaults to removing whitespace."

didrocks (approved):
Good works!

PR #513, opened 2022-11-25
Retry flaky Windows test

### LINKED ISSUE #470: Windows tests are flacky, and we should do something about it
We have quite some failures in CI due to windows testing. We should tackle them to not let that slip through and starting having the habit of "rerunning tests"

GabrielNagy:
Wasn't able to pinpoint the exact reason why only this specific test would fail with "context deadline exceeded" from time to time. I increased the context timeout, added sleeps literally everywhere and it still occurred.

Extract it from the table test (also gives us a chance to clean up most of the variables there which were used for the sole purpose of this specific test) and reimplment it in a separate test that retries the `app.Run()` part a couple of times in case of errors.

To be even more on the safe side, if we get an unexpected error from starting the service we will surface it and still fail the test.

Closes #470 / DEENG-510

### DIFF
--- a/cmd/adwatchd/integration_tests/adwatchd_service_test.go
+++ b/cmd/adwatchd/integration_tests/adwatchd_service_test.go
@@ -2,6 +2,8 @@ package adwatchd_test
 
 import (
 	"bytes"
+	"context"
+	"errors"
 	"fmt"
 	"io"
 	"os"
@@ -21,13 +23,9 @@ import (
 
 func TestServiceStateChange(t *testing.T) {
 	tests := map[string]struct {
-		sequence   []string
-		invalidDir bool
+		sequence []string
 
-		skipUnlessWindows bool
-
-		wantErrAt   []int
-		wantStopped bool
+		wantErrAt []int
 	}{
 		// From stopped state
 		"stop multiple times": {sequence: []string{"stop"}},
@@ -37,8 +35,6 @@ func TestServiceStateChange(t *testing.T) {
 		"install":             {sequence: []string{"install"}, wantErrAt: []int{0}},
 
 		// From started state
-		// This should error on Windows but doesn't with systemd because of the auto-restart policy
-		"start with a bad dir": {sequence: []string{"start"}, invalidDir: true, wantStopped: true, skipUnlessWindows: true},
 		"start multiple times": {sequence: []string{"start", "start"}},
 		"start and stop":       {sequence: []string{"start", "stop"}},
 		"start and restart":    {sequence: []string{"start", "restart"}},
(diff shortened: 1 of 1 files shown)


cmd/adwatchd/integration_tests/adwatchd_service_test.go
```diff
@@ -97,15 +85,53 @@ func TestServiceStateChange(t *testing.T) {
...
+	tries := 0
+	for tries < 5 {
+		changeAppArgs(t, app, "", "service", "start")
+		err = app.Run()
+		if err == nil || !errors.Is(err, context.DeadlineExceeded) {
+			break
+		}
+		tries++
+	}
+	time.Sleep(time.Second)
```
didrocks, line 128: What is this sleep doing? Should it be for getStatus, and thus, after require.NoError?

cmd/adwatchd/integration_tests/adwatchd_service_test.go
```diff
@@ -97,15 +85,53 @@ func TestServiceStateChange(t *testing.T) {
...
+
+	// Begin with a stopped state
+	changeAppArgs(t, app, "", "service", "stop")
+	err = app.Run()
+	require.NoError(t, err, "Setup: Stopping the service failed but shouldn't")
+
+	// Remove watched dir to make the service not start
+	os.RemoveAll(watchDir)
+
```
didrocks, line 116: I think I remember this, but I just want to confirm: stop is blocking, correct? This is why we don’t have a timeout here?

cmd/adwatchd/integration_tests/adwatchd_service_test.go
```diff
@@ -97,15 +85,53 @@ func TestServiceStateChange(t *testing.T) {
...
 	}
 }
+
+func TestServiceStartWithABadDir(t *testing.T) {
+	// This should error on Windows but doesn't with systemd because of the auto-restart policy
+	if runtime.GOOS != "windows" {
+		t.Skip()
+	}
+
+	var err error
```
didrocks, line 99: I think we can skip this declaration

didrocks (commented):
Ok, I don’t really like retrying tests because we never know if it’s the test itself that is flaky or the code/platform itself having a flaky behaviour, but let’s do this for that bad use case.

I just need 2 clarifcations + one small improvements but overall, all the rest is excellent (including moving the comment, thx :))!

Thanks for fighting the CI! :)

didrocks (approved):
Thanks for the changes! Once the tests pass, feel free to merge. Hopefully, no more flakyness reported \o/

PR #517, opened 2022-11-28
Reuse our utility function for comparing trees.

### LINKED ISSUE #462: internal/ad/internal_test.go: replace usage of md5Tree
Mordernize with our comparetree function in testutils.

didrocks:
Remove usage of md5Tree in tests and replace it with our internal tool. The output is slightly different, but still readable in case of failure.

Fixes #462.
DEENG-507

### DIFF
--- a/internal/ad/internal_test.go
+++ b/internal/ad/internal_test.go
@@ -2,11 +2,8 @@ package ad
 
 import (
 	"context"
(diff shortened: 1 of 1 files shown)


denisonbarbosa (approved):
Looks good! It's definitely better to standardize the way we are comparing things.

PR #519, opened 2022-11-28
Yet more rust improvements

denisonbarbosa:
Addresses some PR comments and removes a big dependency that wasn't really needed.

### DIFF
--- a/cmd/adsysd/integration_tests/adsys_mount_test.go
+++ b/cmd/adsysd/integration_tests/adsys_mount_test.go
@@ -20,6 +20,8 @@ func TestUserMountHandler(t *testing.T) {
 		sessionAnswer string
 		noKrbTicket   bool
 
+		addArgs []string
+
 		wantStatus int
 	}{
 		// Single entries
@@ -55,6 +57,14 @@ func TestUserMountHandler(t *testing.T) {
 		// Generic errors
 		"error when trying to mount unsupported protocol": {mountsFile: "mounts_with_unsupported_protocol", wantStatus: 1},
 		"error during mount process":                      {mountsFile: "mounts_with_error", wantStatus: 1},
+
+		// Binary usage cases
+		"correctly prints the help message": {addArgs: []string{"--help"}},
+
+		// Binary usage errors
+		"errors out and prints usage message when executed with less than 2 arguments": {wantStatus: 2},
+		"errors out and prints usage message when executed with more than 2 arguments": {addArgs: []string{"more", "than", "two"}, wantStatus: 2},
+		"errors out and prints usage message even when --help is among the arguments":  {addArgs: []string{"i", "need", "--help"}, wantStatus: 2},
 	}
 	for name, tc := range tests {
 		tc := tc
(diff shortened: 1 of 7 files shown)


internal/policies/mount/adsys_mount/src/main.rs
```diff
@@ -1,22 +1,52 @@
...
     }
 
-    lib::mount(&args.mounts_file)
+    lib::mount(&mounts_file)
+}
+
+fn print_help_msg() {
+    print!(
+        "\
+Adsys helper binary to handle user mounts. This is not intended to be used manually.
```
didrocks, line 34: s/used/run/ ?

internal/policies/mount/adsys_mount/src/main.rs
```diff
@@ -1,22 +1,52 @@
...
-    let args = Args::parse();
+    if std::env::args().any(|p| &p == "--help") {
+        print_help_msg();
+        return Ok(());
+    }
+
+    let mut args = std::env::args();
+
+    // Ignores the first argument, which is the path of the executable.
+    args.next();
```
didrocks, line 13: I think you should check the len too.

internal/policies/mount/adsys_mount/src/main.rs
```diff
@@ -1,22 +1,52 @@
...
+
+    let mounts_file = match args.next() {
+        Some(arg) => arg,
+        None => {
+            print_invalid_msg();
+            return Ok(());
```
didrocks, line 19: This is going to exit 0, non? It should rather return an error of invalid syntax (exit code 2)

didrocks (changes requested):
Some changes needed around the exit code and checking return code, otherwise good!

internal/policies/mount/adsys_mount/src/main.rs
```diff
@@ -1,22 +1,58 @@
...
+        print_usage_msg();
+        return Err(AdsysMountError::ParseError);
+    }
+
+    // Ignores the first argument, which is the path of the executable.
+    args.next();
+
+    let mounts_file = match args.next() {
+        Some(arg) => arg,
+        None => {
```
didrocks, line 23: Now taht we check that the len is what we expected, we never can enter that brace, correct?

I wonder if we can then simplify with `if let Some(mounts_file) = args.next()…`

didrocks:
This looks good! One last thing: can you run the integration tests and see if we can’t trigger some error cases (the new one in parsing), so that we ensure we don't regress here? This should be easy to tackle `--help`, check no error, too many arguments and error for exit code.

internal/policies/mount/adsys_mount/src/main.rs
denisonbarbosa: Since we already ensured that we have two arguments, it's possible to just unwrap it and not worry about panicking out

cmd/adsysd/integration_tests/adsys_mount_test.go
```diff
@@ -41,7 +43,7 @@ func TestUserMountHandler(t *testing.T) {
 
 		// File errors
 		"error when file has badly formated entries": {mountsFile: "mounts_with_bad_entries", wantStatus: 1},
-		"error when file doesn't exist":              {mountsFile: "do_not_exist", wantStatus: 1},
+		"error when file doesn't exist":              {mountsFile: "do_not_exist", wantStatus: 2},
```
didrocks, line 46: This should be 1, it’s not a syntax issue.

denisonbarbosa: Should it? The os error code returned by Rust is 2.

denisonbarbosa: Oh, status 2 is reserved to misuse of shell builtins

didrocks: It should be 1, as it’s an internal process error. Unsure how Rust decide about exit process on internal error.

denisonbarbosa: From what I understood, everything that errors out of your code (i.e. an error from a main function), always counts as status 1. That's why I had to change the way we returned out of the main function in order to be able to get a status 2 when we have invalid args.

didrocks (approved):
Everything looks good now! Let’s merge this :)

PR #520, opened 2022-11-28
Add Winbind DC and online status detection

GabrielNagy:
Using the `wbclient` library we can determine the domain, DC and online status of a domain-joined host using the following C functions:

- `wbcInterfaceDetails`: get current domain name
- `wbcLookupDomainController`: lookup DC given a domain
- `wbcDomainInfo`: get online status given a domain

To test this I've opted to create a mock libwbclient library that behaves according to the value of the `ADSYS_WBCLIENT_BEHAVIOR` environment variable.

Kerberos cached credentials handling for the host still needs to be implemented but cannot be handled using this library, so it's outside the scope of this PR.

DEENG-490

### DIFF
--- a/internal/ad/backends/winbind/winbind_test.go
+++ b/internal/ad/backends/winbind/winbind_test.go
@@ -1,7 +1,6 @@
 package winbind_test
 
 import (
-	"bytes"
 	"context"
 	"flag"
 	"fmt"
@@ -70,29 +69,9 @@ func TestWinbind(t *testing.T) {
 				return
 			}
 
-			var got bytes.Buffer
-			got.WriteString(fmt.Sprintf("* Domain(): %s\n", backend.Domain()))
-
-			serverURL, err := backend.ServerURL(context.Background())
-			serverLine := fmt.Sprintf("* ServerURL(): %s\n", serverURL)
-			if err != nil {
-				serverLine = fmt.Sprintf("* ServerURL ERROR(): %s\n", err)
-			}
-			got.WriteString(serverLine)
-
-			isOnline, err := backend.IsOnline()
-			isOnlineLine := fmt.Sprintf("* IsOnline(): %t\n", isOnline)
-			if err != nil {
-				isOnlineLine = fmt.Sprintf("* IsOnline ERROR(): %s\n", err)
-			}
-			got.WriteString(isOnlineLine)
-
-			got.WriteString(fmt.Sprintf("* HostKrb5CCNAME(): %s\n", backend.HostKrb5CCNAME()))
-			got.WriteString(fmt.Sprintf("* DefaultDomainSuffix(): %s\n", backend.DefaultDomainSuffix()))
-			got.WriteString(fmt.Sprintf("* Config():\n%s\n", backend.Config()))
-
-			want := testutils.LoadWithUpdateFromGolden(t, got.String())
-			require.Equal(t, want, got.String(), "Got expected loaded values in winbind config object")
+			got := testutils.FormatBackendCalls(t, backend)
+			want := testutils.LoadWithUpdateFromGolden(t, got)
+			require.Equal(t, want, got, "Got expected loaded values in winbind config object")
 		})
 	}
 }
--- a/internal/ad/backends/sss/sss_test.go
+++ b/internal/ad/backends/sss/sss_test.go
@@ -1,7 +1,6 @@
 package sss_test
 
 import (
-	"bytes"
 	"context"
 	"errors"
 	"flag"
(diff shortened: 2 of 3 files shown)


internal/ad/backends/winbind/winbind_test.go
```diff
@@ -0,0 +1,187 @@
...
+
+	fmt.Println("Running subprocess with", subArgs)
+	// #nosec G204: this is only for tests, under controlled args
+	cmd := exec.Command(subArgs[0], subArgs[1:]...)
+
+	// Setup correct child environment, including LD_PRELOAD for wbclient mock
+	cmd.Env = append(os.Environ(),
+		"GO_WANT_HELPER_PROCESS=1",
+		// override system libwbclient
+		fmt.Sprintf("LD_PRELOAD=%s", mockLibPath),
```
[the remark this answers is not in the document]
didrocks: Interesting, no need to really push further anyway, `LD_PRELOAD` is syntatically correct too as we really want to point to that lib file.

internal/ad/backends/winbind/winbind.go
```diff
@@ -0,0 +1,172 @@
...
+bool is_online(char *domain) {
+  wbcErr wbc_status = WBC_ERR_UNKNOWN_FAILURE;
+  struct wbcDomainInfo *info = NULL;
+
+  wbc_status = wbcDomainInfo(domain, &info);
+  if (wbc_status != WBC_ERR_SUCCESS) {
+    return false;
+  }
+  return !(info->domain_flags & WBC_DOMINFO_DOMAIN_OFFLINE);
+}
```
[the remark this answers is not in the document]
didrocks: Yeah, I would try returning an error from the errno itself or set it ourself if errno is empty, Here is an example on getting back errno from the Go side with CGo: https://groups.google.com/g/golang-nuts/c/iCY2cjRtV8k/m/Lu7Hrvk_OAIJ

internal/ad/backends/winbind/winbind.go
```diff
@@ -0,0 +1,176 @@
...
+
+bool is_online(char *domain) {
+  wbcErr wbc_status = WBC_ERR_UNKNOWN_FAILURE;
+  struct wbcDomainInfo *info = NULL;
+
+  wbc_status = wbcDomainInfo(domain, &info);
+  if (wbc_status != WBC_ERR_SUCCESS) {
+    // Since there's no general purpose errno that we can use, set it to
+    // whatever wbc_status is and have the caller print the status code.
+	errno = wbc_status;
```
didrocks, line 46: I think we have a tab vs space issue here :)

internal/ad/backends/winbind/winbind.go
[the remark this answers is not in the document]
didrocks: heh, you couldn’t resist :) I agree their transformation of errno is nice!

internal/ad/backends/winbind/winbind.go
didrocks: ofc it should! Open a bug :p

internal/ad/backends/winbind/winbind.go
```diff
@@ -0,0 +1,176 @@
...
+	return "Current backend is Winbind"
+}
+
+// IsOnline refresh and returns if we are online.
+func (w Winbind) IsOnline() (bool, error) {
+	cDomain := C.CString(w.domain)
+	defer C.free(unsafe.Pointer(cDomain))
+	online, err := C.is_online(cDomain)
+	if err != nil {
+		err = fmt.Errorf(i18n.G("could not get online status for domain: status code %d"), err)
```
denisonbarbosa, line 153:
```suggestion
		err = fmt.Errorf(i18n.G("could not get online status for domain %q: status code %d"), w.domain, err)
```

Even though we are kind of scoped here, I feel like adding the domain name to the error message could be helpful to provide more context (redundancy is good sometimes).

internal/ad/backends/winbind/winbind.go
```diff
@@ -0,0 +1,176 @@
...
+	defer C.free(unsafe.Pointer(dc))
+	return C.GoString(dc), nil
+}
+
+func dcName(domain string) (string, error) {
+	cDomain := C.CString(domain)
+	defer C.free(unsafe.Pointer(cDomain))
+	dc := C.get_dc_name(cDomain)
+	if dc == nil {
+		return "", errors.New(i18n.G("could not get domain controller name"))
```
denisonbarbosa, line 172:
```suggestion
		return "", errors.New(fmt.Sprintf(i18n.G("could not get controller name for domain %q"), domain))
```

Same comment about providing more context...

denisonbarbosa (approved):
Apart from a few suggestions, looks good to me. Well done!

internal/ad/backends/winbind/mock/libwbclient_mock.c
```diff
@@ -0,0 +1,61 @@
...
+#include <stdio.h>
+
+#include <wbclient.h>
+
+char *get_mock_behavior()
+{
+    char *behavior = getenv("ADSYS_WBCLIENT_BEHAVIOR");
+    if (behavior == NULL) {
+        printf("ADSYS_WBCLIENT_BEHAVIOR not set, exiting...");
+        exit(1);
```
didrocks, line 14: I feel that if we forget to set it in some tests, it will just exit, correct? Is there a risk that if we come back to it in few months time, we will scratch our head on why the test program just exits?

internal/ad/backends/winbind/winbind.go
didrocks: `fmt.Errorf` then :)

internal/ad/backends/winbind/winbind_test.go
```diff
@@ -0,0 +1,187 @@
...
+
+			config := winbind.Config{}
+			if tc.staticADDomain != "" {
+				config.ADDomain = tc.staticADDomain
+			}
+			if tc.staticADServer != "" {
+				config.ADServer = tc.staticADServer
+			}
+
+			winbind, err := winbind.New(context.Background(), config)
```
didrocks, line 67: shadowing the package name to a variable could be weird. Maybe we can call it backend?

internal/ad/backends/winbind/winbind_test.go
```diff
@@ -0,0 +1,187 @@
...
+			isOnline, err := winbind.IsOnline()
+			isOnlineLine := fmt.Sprintf("* IsOnline(): %t\n", isOnline)
+			if err != nil {
+				isOnlineLine = fmt.Sprintf("* IsOnline ERROR(): %s\n", err)
+			}
+			got.WriteString(isOnlineLine)
+
+			got.WriteString(fmt.Sprintf("* HostKrb5CCNAME(): %s\n", winbind.HostKrb5CCNAME()))
+			got.WriteString(fmt.Sprintf("* DefaultDomainSuffix(): %s\n", winbind.DefaultDomainSuffix()))
+			got.WriteString(fmt.Sprintf("* Config():\n%s\n", winbind.Config()))
```
didrocks, line 92: Thid is a lot of repetition between the 2 backends. Maybe we can just create a backend test helper package, that are importing only by those 2 test files, which takes a writer + the result of all the calls and write/format it for us?

didrocks (changes requested):
Very well done! I have a few changes requested and some questions but overall, this is excellent, despite having some C code :p

internal/ad/backends/winbind/mock/libwbclient_mock.c
didrocks: this is fine for me

internal/ad/backends/winbind/winbind_test.go
didrocks: Excellent! I was almost in favor of doing the golden file update/comparison in my initial comment, but then, I think this is less flexible in the long term, so let’s keep it as you did, apart (chaining small helpers together).

internal/ad/backends/winbind/winbind.go
didrocks: LGTM (but I will let the comment reviewer closing it).

didrocks (approved):
Excellent! Everything is fixed. Approving :)

PR #521, opened 2022-11-29
Changes for mount manager tests

denisonbarbosa:
Refactoring the current mount policy manager tests to prepare for the addition of the system mount tests.


internal/policies/mount/mount_test.go
```diff
@@ -94,14 +94,14 @@ func TestApplyPolicy(t *testing.T) {
...
-		"error when user has invalid uid":                                            {userReturnedUID: "invalid", wantErr: true},
-		"error when user has invalid gid":                                            {userReturnedGID: "invalid", wantErr: true},
-		"error when userDir has invalid permissions":                                 {readOnlyUsersDir: true, wantErr: true},
-		"error when path already exists as a directory":                              {pathAlreadyExists: true, wantErr: true},
-		"error when cleanup with invalid user":                                       {entries: []string{"no entries"}, objectName: "dont exist", wantErr: true},
-		"error when cleanup with no entries and path already exists as a directory":  {entries: []string{"no entries"}, pathAlreadyExists: true, wantErr: true},
-		"error when cleanup with empty entry and path already exists as a directory": {entries: []string{"entry with no value"}, pathAlreadyExists: true, wantErr: true},
+		"user, error when user is not found":                                               {objectName: "dont exist", wantErr: true},
```
didrocks, line 97: I like running all error cases with `TestFoo/error`, I think (we did that for machine and users in other tests) that we can postfix rather with `(user)` or `(machine)` and so do that in the positive cases too.

didrocks (changes requested):
The rest looks good!

internal/policies/mount/mount_test.go
denisonbarbosa: What if we go for something like `error-user` or `error user`? This way, you can still run all the error cases with `Test/error` while still keeping them separated enough to run something like `Test/user` or `Test/error_user`

didrocks: If we want to go that way, we should use the same pattern for every tests. So I suggest that we comply with the other tests for now (knowing that we can then use `.*/error.*user` already).

And we can move all the code base to another pattern if felt necessary.

didrocks: yes, but the test name all contains user and machine, normally

internal/policies/mount/mount_test.go
```diff
@@ -67,25 +67,25 @@ func TestApplyPolicy(t *testing.T) {
...
 
 		// Badly formatted entries.
-		"successfully apply policy trimming whitespaces":           {entries: []string{"entry with spaces"}},
-		"successfully apply policy trimming sequential linebreaks": {entries: []string{"entry with multiple linebreaks"}},
-		"creates only users_user dir if the entry is empty":        {entries: []string{"entry with no value"}},
-		"creates only users dir if there are no entries":           {entries: []string{"no entries"}},
+		"user, successfully apply policy trimming whitespaces":           {entries: []string{"entry with spaces"}},
+		"user, successfully apply policy trimming sequential linebreaks": {entries: []string{"entry with multiple linebreaks"}},
+		"user, creates only users_user dir if the entry is empty":        {entries: []string{"entry with no value"}},
+		"user, creates only users dir if there are no entries":           {entries: []string{"no entries"}},
```
didrocks, line 83: I think some of them won't depend on the user, like no entries, badly formatted entries?

denisonbarbosa: Well, the policy is applied differently if we are doing it for users or for machine

didrocks: you’re correct, this is one of the parameter.

didrocks (approved):

PR #524, opened 2022-12-01
Updating yaml files with the new policy definition for system mounts

denisonbarbosa:


cmd/admxgen/defs/mount.yaml
```diff
@@ -17,3 +17,25 @@
   type: "mount"
   meta:
     strategy: "append"
+
+- key: "/system-mounts"
+  displayname: "System mounts"
+  explaintext: |
+    Define network shares that will be mounted for the system.
+    If more shares are defined higher in other GPO, the entries listed here will be appended to the list and duplicates will be removed.
```
didrocks, line 25: `higher in the GPO hierarchy`? (if we are not consistent with other descriptions, we should)

cmd/admxgen/defs/mount.yaml
```diff
@@ -17,3 +17,25 @@
...
+    
+    Values should be in the format: <protocol>://<hostname-or-ip>/shared-dir.
+    Examples:
+      nfs://example.com/nfs_share
+      smb://example.com/smb_share
+      ftp://example.com/ftp_share
```
didrocks, line 31: ftp can be user@password ? As there is no kerberos tickets for those mounts, I think we should for all protocols, provides the user / password basic authentication scheme in the example.

didrocks (changes requested):
Some comments suggestion

cmd/admxgen/defs/mount.yaml
denisonbarbosa: Well... This complicate things a lot. We don't account for this in the policy manager (nor anywhere else in the code, tbh). And, as those values will be stored in the system in a file (take the user mounts, for example), shouldn't we avoid having this kind of authentication for safety reasons?

didrocks: So, what’s the difference for system mount between:
`nfs://example.com/nfs_share` and `[anonymous]nfs://example.com/nfs_share`?

denisonbarbosa: The options that we define on the unit file, basically...

didrocks: Can you be a little bit more precise? (and how does it effect a mount where we don’t pass username/password?)

denisonbarbosa: Well, from what I've read, system mounts do not have a specific "anonymous" option for them. Which means I'll need to set some other options to specify that the user that is interacting with the mounted location shouldn't be checked and so on...

cmd/admxgen/defs/mount.yaml
```diff
@@ -17,3 +17,25 @@
...
+
+    Anonymous mounts should be prefixed by the [anonymous] tag:
+      [anonymous]<protocol>://<hostname-or-ip>/shared-dir
```
didrocks, line 34: Note: this needs to be updated with the removal of [anynoumous] tag and telling that only anynomous mount are supported for system mounts.

cmd/admxgen/defs/mount.yaml
```diff
@@ -26,3 +26,32 @@
...
+    Values should be in the format: 
+        <protocol>://<hostname-or-ip>/<shared-dir>
+    e.g.
+        nfs://example_nfs.com/nfs_shared_dir
+        cifs://example_cifs.com/cifs_shared_dir
+        fuse://ftp_share_server.com
+
+    This pattern must be followed, otherwise the policy will not be applied.
+    
+    By the default, the mounts will be done in anonymous mode. In case of authentication needed, a krb5 tag can be added to the value, e.g.
```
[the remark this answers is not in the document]
denisonbarbosa: Did you share the folder anonymously?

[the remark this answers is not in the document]
denisonbarbosa: IIRC NFS has a different approach than anonymous per se... It has something called "access based enumeration" which basically translates to: "everyone can mount this, but they will only see and do what they are allowed to"

cmd/admxgen/defs/mount.yaml
```diff
@@ -26,3 +26,32 @@
...
+  explaintext: |
+    Define network shares that will be mounted for the system.
+    If more shares are defined higher in the GPO hierarchy, the entries listed here will be appended to the list and duplicates will be removed.
+    
+    Values should be in the format: 
+        <protocol>://<hostname-or-ip>/<shared-dir>
+    e.g.
+        nfs://example_nfs.com/nfs_shared_dir
+        cifs://example_cifs.com/cifs_shared_dir
+        fuse://ftp_share_server.com
```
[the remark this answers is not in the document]
denisonbarbosa: We accept [both](https://github.com/ubuntu/adsys/blob/f6cf6e2d06391a646651201533b2dc5dd836ba84/internal/policies/mount/mount.go#L310) 😄.

However, as I've said in the commit rationale, we should list the correct input in the policy to avoid creating some "laziness"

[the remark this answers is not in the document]
denisonbarbosa: For the user is different... Gio supports the smb:// directly, so there's no need to do it.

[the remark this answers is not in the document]
denisonbarbosa: After what we talked in IIRC, about gio having to use smb:// instead of cifs:// and so on, do you think we can merge this?

GabrielNagy (approved):

Issue #523, opened 2022-12-01
Issues with FQDN Hostname in Environment with Subdomain


snussbaumermpreis:
We have a Active Directory Domain with the following Scheme: <name.xy.xy>
So if I join a Host and set its Hostname to FQDN (hostname.name.xy.xy) as required for other services, adsysctl update does not work, the folliwing errors come up:

Searching for account failed with: Failed to find account hostname.name.xy.xy
Searching for account failed with: Failed to find account hostname.name

So in the second attempt adsys seems to try to use just the Hostname but as the domain has an additional level it does not remove the complete domainpart. The fix should be rather simple -> cut the hostname at the first dot.

denisonbarbosa:
Hey @snussbaumermpreis, thanks for opening the issue. We weren't considering FQDN hostnames being set in /etc/hostname, so #526 should fix your problem! 
However, I would recommend you to take another look at your setup and check if you can configure the required fqdn, for the other services that you've mentioned, in /etc/hosts and keep the machine hostname single-labeled, as recommended by the [hostname man page](https://man7.org/linux/man-pages/man5/hostname.5.html) and the [debian documentation](https://www.debian.org/doc/manuals/debian-reference/ch03.en.html#_the_hostname).

PR #526, opened 2022-12-02
Improvements on the handling of hostnames to account for FQDN

denisonbarbosa:
Fixes a bug where FQDN hostnames were being treated as the account name to be queried for when searching for GPOs. Addresses issue #523.

DEENG-538

### DIFF
--- a/cmd/adsysd/client/policy.go
+++ b/cmd/adsysd/client/policy.go
@@ -344,10 +344,7 @@ func (a *App) update(isComputer, updateAll bool, target, krb5cc string) error {
 			return err
 		}
 		// for malconfigured machines where /proc/sys/kernel/hostname returns the fqdn and not only the machine name, strip it
-		if i := strings.Index(hostname, "."); i > 0 {
-			hostname = hostname[:i]
-		}
-		target = hostname
+		target, _, _ = strings.Cut(hostname, ".")
 	}
 
 	// Update for current user
--- a/internal/ad/ad.go
+++ b/internal/ad/ad.go
@@ -565,6 +565,7 @@ func (ad *AD) NormalizeTargetName(ctx context.Context, target string, objectClas
 	return target, nil
 }
 
+// Hostname returns the hostname of the current client.
 func (ad *AD) Hostname() string {
 	return ad.hostname
 }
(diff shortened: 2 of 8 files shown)


internal/ad/adsys-gpolist
```diff
@@ -234,6 +234,9 @@ def main():
     # Users don’t need @, as we already have the specific-domain ticket
     if args.objectclass == ObjectClass.user:
         accountname = accountname.split('@')[0]
+    else:
+        # Removes the domain part on FQDN (hostname.domain) hostnames from the account name
+        accountname = accountname.split('.')[0]
```
didrocks, line 239: Are you sure this is the right place for this fix? Shouldn’t it be at the same level when we normalize the user name in the Go code itself? I feel that we may store some cache with FQDN when we should not…

didrocks (changes requested):
I’m pretty sure we should do the test way higher in the stack, when we receive the arguments from the client, to normalize it everywhere and drop the FQDN from the start.

internal/ad/adsys-gpolist
denisonbarbosa: It could be (but then why are we only cutting off the domain here?). I basically tracked the error message that the user reported and, when I've seen the if with the split at @, I thought that this script could be run directly.

didrocks: > (but then why are we only cutting off the domain here?)
When we normalize the user, we normalize the user to `user@domain.com` and this is what is used to store all caches and so on (as an user can be in multiple domain). So, the is the normalized name.

This scripts though (through samba) as the domain as a parameter and won’t accept the user in this form, this is why we are stripping it.

The machine is only part of one domain, this is why the cached files on disk should not contain the domain.

> I thought that this script could be run directly.
No, it’s embeeded in Go code and executed only by it (it’s even not accessible for the user apart from explicitely requesting it for debugging purposes)

denisonbarbosa: I'm trying to set up a test for it, as it appears we don't have any

didrocks (approved):
That looks good! FYI, I have done a direct commit to the branch as you forgot about the description comment update during the live review (https://github.com/ubuntu/adsys/pull/526/commits/48466139730322f3dc4c4882769ab1d442b1c26c).

Otherwise, as said yesterday, this is the right way to fix it IMHO, approving, good work!

PR #529, opened 2022-12-05
Updating apply policy behavior

denisonbarbosa:
After some talk with @didrocks, we decided that the default mount behavior should be anonymous mounting with the krb5 authentication being transferred to a tag. The purpose of this change is to make the user mounts have the same behavior as the system mounts, which are anonymous by default.


didrocks (commented):
Only 28 files modified? I’m almost disappointed :p

Otherwise, the changes (albeit my comments below) are the one we describe and it makes a lot more sense in general (even after a night of sleep over it). Thanks for tackling this!

You mixed 2 things in a single commit: the returned type of parsing the entry is a different matter than this authentication reversal, but as the diff is small, I’ll let this pass on. It should have been at least 2 commits (and ideally 2 PRs).

Also, It would have been nice for a longer description in the commit message to explain why we do this reverse (having a single syntax for system and user mounts). Once you merge it, you can maybe ensure that the extended description is a little bit more verbose? Imagine we git log to get the history of the change, here, we would have some descriptive content.

Finally, we mentioned it yesterday but I think you miss it: you need to update the .yaml description (so that admx/adml are updated too). Can you do that before I give my approval on this PR?

I think your goal is then to cherry-pick the merged commits for the system mount? (as you mixed things that are in main and not in main, I fear that may be complicated unless your rebase on a non merged branch, which isn’t ideal). I thought we mention yesterday to put the part which is impacting main against it, so that it’s easier fror you to then iterate over the system mount branch, but tell me what  you need to do.

didrocks (approved):
Once https://github.com/ubuntu/adsys/pull/530 is merged to that one, feel free to merge it! Approved :)

PR #532, opened 2022-12-06
Changing default mount to anonymous instead of krb authenticated.

denisonbarbosa:
When dealing with the system mounts, we (me and @didrocks) realized that the default mount behavior is to mount anonymously.
In order to avoid having the user manager behaving differently, we decided to switch the default mount process (which relates to the entries without any tag e.g. smb://domain.com/shared-dir) to anonymous and have both managers behave in a similar way.


didrocks (approved):
LGTM! I think backporting this to main will be easier :)

PR #533, opened 2022-12-06
Change user mount manager behavior: now it stops and errors out when it's not able to set up the policy.

denisonbarbosa:
After some discussion, it was clear that some manager behavior would need to change. We needed to have a clearer definition of when a manager should fail and why.
Now, the mount manager will error out whenever it fails to set up the require files for the mounting process.


didrocks (approved):
Excellent! That will align everything neatly :) Approving.

PR #536, opened 2022-12-06
Reflecting default mount changes to binary

denisonbarbosa:


didrocks:
I’ll let you check the test failure, which seems real.

Let’s step up the review a little bit, here the commit message doesn’t really help "Reflecting the changes on the entry tags". What’s the changes? :p Do you mind rewriting it with an extended description (see my comment on the last merged PR which was already on this, this is why, as you are directly redoing the same thing, that I hold here :p)

didrocks (approved):
Thanks! Those are way better commit messages and tests now pass :)

Approving.

PR #541, opened 2022-12-07
Adding the adsys-user-mount.service file to systemd/user

denisonbarbosa:
DEENG-542

### DIFF
--- a/systemd/user/adsys-user-mounts.service
+++ b/systemd/user/adsys-user-mounts.service
@@ -1,0 +1,13 @@
+[Unit]
+Description=ADSys user mount handler
+After=network-online.service
+ConditionPathExists=/run/adsys/users/%U/mounts
+ConditionPathExists=/usr/libexec/adsys_mount
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStart=/usr/libexec/adsys_mount /run/adsys/users/%U/mounts
+
+[Install]
+WantedBy=default.target


systemd/user/adsys-user-mounts.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=ADSys user mount handler
+After=network-online.service
+ConditionPathExists=/run/adsys/users/%U/mounts
+ConditionPathExists=/usr/libexec/adsys_mount
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
```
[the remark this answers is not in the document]
denisonbarbosa: Tbh, I have no idea if that's precisely what we want. This would trigger the binary to run again after a policy refresh, which could lead to lots of "location is already mounted" warns so maybe it's better to lock it to a `one time per session` strategy?
I'm not sure about it, btw... This is just to trigger some thinking and discussion about it.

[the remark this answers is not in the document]
denisonbarbosa: Actually, this is already what happens btw... We don't control this service from the adsys daemon. This is something that will be enabled when the adsys package is installed and then it's up to systemd, which means it will only happen at log in.

systemd/user/adsys-user-mounts.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=ADSys user mount handler
+After=network-online.service
+ConditionPathExists=/run/adsys/users/%U/mounts
+ConditionPathExists=/usr/libexec/adsys_mount
```
didrocks, line 5: I don’t think the second ConditionPathExists really makes sense. We rarely have a service checking for the binaries it’s running if they exists or not.

If the service is installed, it’s expected for the binary to be present on disk, or it’s a packaging or system consistency issue.

Similarly, if a binary is using an asset (like an image), the fact that the image is present is a delivery and consistency issue, not a code one.

Making sense?

denisonbarbosa: It makes sense, indeed. But, even though the binary is supposed to be there, I felt like it would be good to have this condition explicitly defined in the unit for clarity. Wdyt?

didrocks: Actually, I would prefer not:
- as said, this is a system issue to not have the binary when you the service
- if we keep the ConditionPathExists, it means that the unit is skipped, not in failure, when it should rather fail :)

denisonbarbosa: Ack, I will fix it in #544.

PR #544, opened 2022-12-09
System mounts policy

denisonbarbosa:
Adds the System Mounts policy to ADSys.


cmd/admxgen/defs/mount.yaml
```diff
@@ -20,9 +20,39 @@
...
+        ftp://ftp_share_server.com
+
+    This pattern must be followed, otherwise the policy will not be applied.
+    
+    By default, the mounts will be done in anonymous mode. In case of authentication needed, a krb5 tag can be added to the value, e.g.
+        [krb5]<protocol>://<hostname-or-ip>/<shared-dir>
+
+    If the tag is added, the mount will require Kerberos authentication in order to occur.
+
+    The supported protocols / filsystems are the same as the ones supported by the mount command.
```
didrocks, line 51: You shold mention the systemd mount unit doc there.

denisonbarbosa: Well, the [systemd.mount]{https://www.freedesktop.org/software/systemd/man/systemd.mount.html} doc is kind of lackluster. It just tells you about the syntax of the mount section, basically. Information about the mount protocols and mount options is given on the [mount man page]{https://manpages.ubuntu.com/manpages/jammy/en/man8/mount.8.html} and the systemd.mount actually points you towards it, that's why I decided to reference directly the mount page, since the user does not need to worry about the unit syntax.

didrocks: ack then :)

didrocks:
ok, I think I’m done with different PRs around this one. See the details on each one. In addition to those:
- I think it’s worth exploring a way to not remove the cargo lock file (see the exchange on the packaging PR)
- as we discussed yesterday, some CI to update the X- vendored deps rust will update to fix the FTBS in kinetic+ and having it always up to date.
- and finally, update debian/copyright (yes, I know, this is exciting…)

didrocks (approved):
Once all the tests pass with the last commit, feel free to merge. Done! :)

PR #548, opened 2022-12-12
Get krb5cc using kinit for winbind backend

GabrielNagy:
Unlike sssd, winbind doesn't store the Kerberos cached credentials for the host anywhere, so we have to request this ourselves.

The simplest way to do this is to interact with the system keytab and obtain the ticket for the principal using the `kinit` command.

To be able to implement this in the correct location, the backend interface now also returns an error for the `HostKrb5CCNAME` function, even though not all backends might make use of it.

DEENG-489

### DIFF
--- a/internal/ad/ad.go
+++ b/internal/ad/ad.go
@@ -191,7 +191,7 @@ func (ad *AD) GetPolicies(ctx context.Context, objectName string, objectClass Ob
 		src := userKrb5CCName
 		// there is no env var for machine: get sss ccache
 		if objectClass == ComputerObject {
-			src, err = ad.configBackend.HostKrb5CCNAME()
+			src, err = ad.configBackend.HostKrb5CCName()
 			if err != nil {
 				return pols, err
 			}
--- a/internal/ad/backends/winbind/winbind.go
+++ b/internal/ad/backends/winbind/winbind.go
@@ -126,8 +126,8 @@ func (w Winbind) Domain() string {
 	return w.domain
 }
 
-// HostKrb5CCNAME returns the absolute path of the machine krb5 ticket.
-func (w Winbind) HostKrb5CCNAME() (string, error) {
+// HostKrb5CCName returns the absolute path of the machine krb5 ticket.
+func (w Winbind) HostKrb5CCName() (string, error) {
 	target := "/tmp/krb5cc_0"
 
 	// Uppercase domain and hostname
--- a/internal/ad/ad_test.go
+++ b/internal/ad/ad_test.go
@@ -459,7 +459,7 @@ func TestGetPolicies(t *testing.T) {
 			objectClass: ad.ComputerObject,
 			backend: mock.Backend{
 				Dom:                "gpoonly.com",
-				HostKrb5CCNAMEPath: "dont-exist",
+				HostKrb5CCNamePath: "dont-exist",
 				Online:             true,
 			},
 			gpoListArgs: []string{"gpoonly.com", hostname + ":standard"},
@@ -535,11 +535,11 @@ func TestGetPolicies(t *testing.T) {
 				tc.backend.ServURL = "ldap://myserver." + tc.backend.Dom
 			}
 			// we file in host_ccache to not have to reset it in every single test
-			if tc.backend.HostKrb5CCNAMEPath == "" {
-				tc.backend.HostKrb5CCNAMEPath = filepath.Join(t.TempDir(), "host_ccache")
+			if tc.backend.HostKrb5CCNamePath == "" {
+				tc.backend.HostKrb5CCNamePath = filepath.Join(t.TempDir(), "host_ccache")
 			}
-			if tc.backend.HostKrb5CCNAMEPath != "dont-exist" {
-				testutils.CreatePath(t, tc.backend.HostKrb5CCNAMEPath)
+			if tc.backend.HostKrb5CCNamePath != "dont-exist" {
+				testutils.CreatePath(t, tc.backend.HostKrb5CCNamePath)
 			}
 
 			var krb5CCName string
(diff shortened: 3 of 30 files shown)


internal/ad/backends/winbind/winbind.go
```diff
@@ -110,8 +134,21 @@ func (w Winbind) Domain() string {
...
+func (w Winbind) HostKrb5CCNAME() (string, error) {
+	target := "/tmp/krb5cc_0"
+
+	// Uppercase domain and hostname, and trim domain from hostname if needed
+	domain := strings.ToUpper(w.domain)
+	hostname := strings.TrimSuffix(strings.ToUpper(w.hostname), "."+domain)
+
+	principal := fmt.Sprintf("%s$@%s", hostname, domain)
+	cmdArgs := append(w.kinitCmd, "-k", principal, "-c", target)
+	if cmd, err := exec.Command(cmdArgs[0], cmdArgs[1:]...).CombinedOutput(); err != nil {
```
denisonbarbosa, line 146: I'm pretty sure the smbsafe mutex stuff, that I've mentioned to you about the cmd executions, applies here as well.

denisonbarbosa (commented):
Looks very good! I had some comments, but nothing too big.
Except the smbsafe part.That's quite important...

internal/ad/backends/winbind/winbind.go
```diff
@@ -110,8 +134,21 @@ func (w Winbind) Domain() string {
...
 
 // HostKrb5CCNAME returns the absolute path of the machine krb5 ticket.
-func (w Winbind) HostKrb5CCNAME() string {
-	return w.hostKrb5CCNAME
+func (w Winbind) HostKrb5CCNAME() (string, error) {
+	target := "/tmp/krb5cc_0"
+
+	// Uppercase domain and hostname, and trim domain from hostname if needed
+	domain := strings.ToUpper(w.domain)
+	hostname := strings.TrimSuffix(strings.ToUpper(w.hostname), "."+domain)
```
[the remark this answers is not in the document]
denisonbarbosa: You're still treating the FQDN here. Is there any occasion where this may happen? Since the winbind back end is being created by the ad (and in the ad we already handle this use case), I can't see a place where this can happen (outside of an enforced environment in the tests).

denisonbarbosa (approved):
Apart from the last doubts about the FQDN, everything is addressed. Feel free to resolve the last comment and merge this once you handled it (or not)

internal/ad/ad.go
```diff
@@ -186,12 +186,15 @@ func (ad *AD) GetPolicies(ctx context.Context, objectName string, objectClass Ob
...
 	}
-	// Create a ccache symlink on first fetch for futur calls (on refresh for instance)
+	// Create a ccache symlink on first fetch for future calls (on refresh for instance)
 	if userKrb5CCName != "" || objectClass == ComputerObject {
 		src := userKrb5CCName
 		// there is no env var for machine: get sss ccache
 		if objectClass == ComputerObject {
-			src = ad.configBackend.HostKrb5CCNAME()
+			src, err = ad.configBackend.HostKrb5CCName()
+			if err != nil {
```
didrocks, line 195: We are adding a new error condition here (and the mock can trigger it), but I don’t see any additional tests (which are using the mock IIRC) to trigger that code path?

internal/ad/backends/winbind/winbind.go
```diff
@@ -79,10 +82,26 @@ type Config struct {
 	ADDomain string `mapstructure:"ad_domain"` // bypass domain name detection and use this domain
 }
 
+// Option reprents an optional function to change the winbind backend.
```
didrocks, line 85: represents*

internal/ad/backends/winbind/winbind.go
```diff
@@ -109,9 +126,24 @@ func (w Winbind) Domain() string {
...
+// HostKrb5CCName returns the absolute path of the machine krb5 ticket.
+func (w Winbind) HostKrb5CCName() (string, error) {
+	target := "/tmp/krb5cc_0"
+
+	// Uppercase domain and hostname
+	domain := strings.ToUpper(w.domain)
+	hostname := strings.ToUpper(w.hostname)
+
+	principal := fmt.Sprintf("%s$@%s", hostname, domain)
+	cmdArgs := append(w.kinitCmd, "-k", principal, "-c", target)
```
didrocks, line 138: As we are using kinit, shouldn’t the adsys package depends on klibc-utils explicitly?

internal/ad/backends/winbind/winbind_test.go
```diff
@@ -63,19 +71,62 @@ func TestWinbind(t *testing.T) {
...
+			backend, err := winbind.New(context.Background(), config, hostname, winbind.WithKinitCmd(kinitCmd))
 			if tc.wantErr {
 				require.Error(t, err, "New should have errored out")
 				return
 			}
 
 			got := testutils.FormatBackendCalls(t, backend)
+
+			// Check kinit command
+			gotKinitCmd := []byte("not executed\n")
```
didrocks, line 89: We thus have "not executed" in the test golden file written in case of error requested in kinit. When reading it, I feel this to be a little bit weird… (also, it doesn’t differentiate not being executed and erroring out).

internal/ad/backends/winbind/winbind_test.go
```diff
@@ -32,8 +32,10 @@ func TestWinbind(t *testing.T) {
 		wbclientBehavior string
 		staticADDomain   string
 		staticADServer   string
+		hostname         string
```
didrocks, line 35: You don’t have a test case and don’t assign hostname to anything in any test (which is a valid test to add btw :)). So, they all end up with ubuntu.

PR #549, opened 2022-12-12
Rust packaging

schopin-pro:
It was both easier and hairier than I thought it would have been.

I wasn't able to actually build the package, as there are some test failures in the apparmor policy (doesn't seem related to the Rust code at all?), and when I use nocheck to get past those, there are some windows things that are missing at the install stage.

I believe I've taken it far enough for y'all to fix those?


denisonbarbosa (approved):
LGTM. Thanks a lot for the help with the packaging and the insights you provided!

debian/rules
```diff
@@ -24,17 +24,67 @@ export DH_GOLANG_INSTALL_ALL := 1
...
+	# need to update the Vendored-Sources field. The initial packaging work
+	# didn't include said field since the vendoring is done on the fly.
+
+override_dh_auto_configure:
+	dh_auto_configure
+	# Normally we keep the lockfile around when using vendored deps, except we
+	# had to mess with the checksums of system-deps and so cargo complains.
+	# That doesn't actually matter since dh-cargo configured cargo not to query
+	# the network for deps, instead relying on the vendored ones.
+	rm -f Cargo.lock
```
didrocks, line 68: I’m not a big fan of this. It means that running debuild in the source directory removes some files in the source package itself, which is packaging nono in general. Also, we can’t `dpkg-buildpackage -nc` to continue a failed build after some fixes we wanted too if we removed the vendor directory.

Of course `Cargo.lock` is less useful from a security team perspective as `XS-Vendored-Sources-Rust` is keeping the exact version of vendored dependency for us, but I think this issue may be considered.

Of course,

schopin-pro:
Hmm, fair point. I'm pretty much never using debuild directly, only
using sbuild, so I hadn't caught onto that part.

I guess we could update the lockfile during the vendoring step? I'm not
a huge fan of the approach as this means the lockfile in the sources
won't be the one in the VCS.

Something along those lines (untested):

```diff
override_dh_auto_clean:
 	dh_auto_clean
 	dh_auto_clean --buildsystem=cargo
 	# Create the vendor directory when building the source package
 	[ -d vendor/ ] || go mod vendor
 	# system-deps crate includes an empty .a file which is filtered out when
 	# building the source package and cargo gets confused when trying to verify
 	# integrity. So we just put in an empty checksum file instead.
 	if [ ! -d $(CARGO_VENDOR_DIR)/ ]; then \
 		$(CARGO) vendor $(CARGO_VENDOR_DIR); \
 		cp debian/cargo-checksum.json
$(CARGO_VENDOR_DIR)/system-deps/.cargo-checksum.json; \
+		$(CARGO) --config
source.crates-io.replace-with=\"vendored-sources\" --config
sources.vendored-sources.directory=\"$(CARGO_VENDOR_DIR)\" --offline
generate-lockfile; \
 		[ ! -e $(DH_CARGO_VENDORED_SOURCES) ] || $(DH_CARGO_VENDORED_SOURCES); \
 	fi
 	# That last line above is to warn whoever's doing the packaging that they
 	# need to update the Vendored-Sources field. The initial packaging work
 	# didn't include said field since the vendoring is done on the fly.
```

didrocks:
I’m unsure to undestand the base issue first. Do you mind expanding it to me?

From what I see in other languages and I think is the correct behaviour:
* upstream defines a .lock file, this is the components and respective versions that every developer working on a project agrees upon. This thus needs to be respected (it seems that your proposal changes this and so, vendored dependency versions are not necessarly the ones used in the lock file?). Similarly, by removing it in the configuration, we don’t have this double check?
* normally, packaging don’t touch upstream files after running, even if they are not ran out of source tree. At most, they add files (preferably under the debian/ directory‘ and upstream toolchain creates files/directories for the build itself, preferably, under a architecture-specific target.

Here, it seems that the issue is a mismatch with the checksum files generated by some Rust tools? Again, I’m unsure to understand exactly the `system-deps` thingy, do you mind giving some insights?

schopin-pro:
Your understanding of lockfiles is correct.

From what I can tell, system-deps is a helper to deal with system
libraries. According to `cargo tree -i system-deps`, it's
a reverse-dependency of gio-sys and glib-sys. My educated guess is that
it's used to conditionally compile some bindings against newer APIs
depending on the version of the system's glib.

This crate ships some .a files as part of its test suite. One might
argue whether or not a release tarball should include its testsuite, but
I tend to think it's fairly reasonable to do so.

So far, so good. However, `adsys` is a `3.0 (native)` package, which
tries to be clever and filters out lots of files at build time,
including the .a file, changing the checksums of the crate. The old
checksums are recorded at two places:

* in the .cargo-checksum.json
  file which is generated during the upload to crates.io and holds
  checksums for every file in the crate, as well as the checksum for the
  entire crate,
* in `adsys`'s Cargo.lock file, which for each crate records its global
  checksum.

To work around the first issue, I replace the .cargo-checksum.json file
with a bogus one (weirdly enough, the global checksum isn't checked by
`cargo`). The exact same thing is done by dh-cargo for individual
packages (all the librust-*-dev packages).

To remove the Cargo.lock issue, the easiest way is simply to remove it
and let `cargo` regenerate it against the vendored versions (or the
packaged ones in the case of dh-cargo and a standard Debian package). In
the vendored case the result will be pretty much the same file except
that the checksums will be different. If the package was a `3.0 (quilt)`
package the cleanest way would probably have been to use a patch to
change the checksums in the Cargo.lock file (if I ever manage to figure
out how it's computed...), but here...

Please note that while this specific issue can be traced to the
native package thing, you might want to remove files from vendored deps
anyway. For instance if your dependency tree ends up with the winapi
crates, there are massive binary blobs in there. Or there could be full
C library sources embedded, which would be a PITA from a d/copyright
standpoint.

didrocks:
Thanks for the detailed explanation. I understand the issue way better now. We’ll try to explore some more fine-graine cases looking at how Cargo works if we can only skip the checksum checks in some way or other potential workarounds.

PR #552, opened 2022-12-14
Wrap apparmor calls with smbsafe

### LINKED ISSUE #542: Calls to apparmor should be wrapped with smbsafe
We missed this part when implementing the apparmor manager.

From the `smbsafe` documentation:
> libsmbclient overrides sigchild without setting SA_ONSTACK
It means that any cmd.Wait() would segfault when ran concurrently with this.

GabrielNagy:
This is a requirement for executed commands because `libsmbclient` overrides sigchild without setting `SA_ONSTACK`, so any `cmd.Wait()` would segfault when ran concurrently with this.

Closes #542 / DEENG-541

### DIFF
--- a/internal/policies/apparmor/apparmor.go
+++ b/internal/policies/apparmor/apparmor.go
@@ -21,6 +21,7 @@ import (
 	log "github.com/ubuntu/adsys/internal/grpc/logstreamer"
 	"github.com/ubuntu/adsys/internal/i18n"
 	"github.com/ubuntu/adsys/internal/policies/entry"
+	"github.com/ubuntu/adsys/internal/smbsafe"
 	"golang.org/x/exp/slices"
 )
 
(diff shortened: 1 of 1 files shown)


internal/policies/apparmor/apparmor.go
```diff
@@ -229,6 +230,8 @@ func (m *Manager) applyMachinePolicy(ctx context.Context, e entry.Entry, apparmo
 		// #nosec G204 - We are in control of the arguments
 		cmd := exec.CommandContext(ctx, apparmorParserCmd[0], apparmorParserCmd[1:]...)
 		cmd.Dir = m.apparmorDir
+		smbsafe.WaitExec()
+		defer smbsafe.DoneExec()
```
didrocks, line 234: defer encompasses for now the `os.RemoveAll` wich may take some time. I think it’s better to release the lock as soon as possible.

Of course, think about doing it in the error case too :)

PR #553, opened 2022-12-14
Adding cifs-utils and curlftpfs as dependencies for the adsys package

denisonbarbosa:
I might be completely wrong, but I think that we only need to add some dependencies to the package and then the packaging for the system mounts is done.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -34,6 +34,8 @@ Depends: ${shlibs:Depends},
          sssd,
          sssd-dbus,
          apparmor,
+         cifs-utils,
+         curlftpfs,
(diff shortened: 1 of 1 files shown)


debian/control
```diff
@@ -34,6 +34,8 @@ Depends: ${shlibs:Depends},
          sssd,
          sssd-dbus,
          apparmor,
+         cifs-utils,
+         curlftpfs,
```
didrocks, line 38: This is in universe, so we either downgrade it to suggests + add some documentation in the system mount adsys doc page or file a MIR for it.

I think suggestion number 1.

PR #555, opened 2023-01-03
Adjusting package dependencies

denisonbarbosa:
Moving curlftpfs package from a dependency to a recommendation, as it's on universe rather than main.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -35,9 +35,9 @@ Depends: ${shlibs:Depends},
          sssd-dbus,
          apparmor,
          cifs-utils,
-         curlftpfs,
 Recommends: ${misc:Recommends},
             ubuntu-advantage-desktop-daemon,
+            curlftpfs,
 Description: ${source:Synopsis}
  ${source:Extended-Description}
 
--- a/debian/control
+++ b/debian/control
@@ -37,7 +37,8 @@ Depends: ${shlibs:Depends},
          cifs-utils,
 Recommends: ${misc:Recommends},
             ubuntu-advantage-desktop-daemon,
-            curlftpfs,
+Suggests:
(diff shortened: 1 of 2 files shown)


debian/control
```diff
@@ -35,9 +35,9 @@ Depends: ${shlibs:Depends},
          sssd-dbus,
          apparmor,
          cifs-utils,
-         curlftpfs,
 Recommends: ${misc:Recommends},
             ubuntu-advantage-desktop-daemon,
+            curlftpfs,
```
didrocks, line 40: This is a Recommends, not a Suggests. As we discussed duringour HO, Ubuntu installs recommends by default, but not Suggests.

doc/09.-Network-shares.md
```diff
@@ -12,6 +12,8 @@ The mount process for these mounts is triggered at the moment a client logs in.
 
 All protocols supported by the [mount command](https://manpages.ubuntu.com/manpages/jammy/en/man8/mount.8.html) should work out of the box. However, the only tested ones are `smb`, `ftp` and `nfs`.
 
+The backends for the protocols `smb` and `nfs` are automatically enabled when installing the adsys package. In order to enable the backend for `ftp` mounts, the user must install the recommended `curlftpfs` package.
```
didrocks, line 15: Maybe mentions that this is valid on Ubuntu? (in case some other distros carries adsys one day)

didrocks (changes requested):

debian/control
denisonbarbosa: Oh, I switched the definitions mentally... My bad, brain is still a bit slow after the holidays

didrocks (approved):
let’s goooooooooo

PR #556, opened 2023-01-03
Winbind backend updates

GabrielNagy:
More details in the individual commits.


didrocks (approved):
Perfect, this is addressing all comments, thanks for that! :)

PR #557, opened 2023-01-03
Tighten smbsafe wrapping in the apparmor manager

GabrielNagy:
Instead of deferring, call DoneExec immediately after executing the commands in case subsequent logic can occur in the function.


didrocks (approved):
I wasn’t pushing for the second change as we are exiting right away (apart from line wrapping) from the function, but this is safe too, let’s merge! Good work and sorry for not having spotted it before talking with Denis and pointing this was lacking!

PR #558, opened 2023-01-04
System mounts test improvements

denisonbarbosa:
Addressing comments made on #522 and a few other adjustments.


internal/policies/mount/mount_test.go
```diff
@@ -116,7 +121,10 @@ func TestApplyPolicy(t *testing.T) {
 		"system, successfully apply policy filtering out unsupported keys": {entries: []string{"entry with multiple values", "entry with one value"}, keys: []string{"unsupported", "system-mounts"}, isComputer: true},
 
 		// Special cases.
-		"system, successfully apply policy with kerberos tagged values": {entries: []string{"entry with kerberos auth tags"}, isComputer: true},
+		"system, successfully apply policy with kerberos tagged values":                         {entries: []string{"entry with kerberos auth tags"}, isComputer: true},
+		"system, successfully apply policy prioritizing the first value found, despite the tag": {entries: []string{"entry with same values tagged and untagged"}, isComputer: true},
+		"system, only emit a warning when starting new units and systemctl fails":               {isComputer: true, firstSystemCtlFailingArgs: []string{"start"}},
```
didrocks, line 126: TODO: add a new test for stop too, when disable --now is split (see https://github.com/ubuntu/adsys/pull/522#discussion_r1061546752).

didrocks (changes requested):

internal/policies/mount/mount_test.go
```diff
@@ -95,7 +97,9 @@ func TestApplyPolicy(t *testing.T) {
 		"user, successfully apply policy filtering out unsupported keys": {entries: []string{"entry with multiple values", "entry with one value"}, keys: []string{"unsupported", "user-mounts"}},
 
 		// Special cases.
-		"user, successfully apply policy with kerberos auth tags": {entries: []string{"entry with kerberos auth tags"}},
+		"user, successfully apply policy with kerberos auth tags":                             {entries: []string{"entry with kerberos auth tags"}},
+		"user, successfully apply policy prioritizing the first value found, despite the tag": {entries: []string{"entry with same values tagged and untagged"}},
```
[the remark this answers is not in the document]
denisonbarbosa: I'm not sure I understood what you meant.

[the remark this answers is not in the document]
didrocks: a test case with the same tag twice (duplicated) and not only anynomous

didrocks:
that + see the discussion on MM for https://github.com/ubuntu/adsys/pull/522#discussion_r1061544546

internal/policies/mount/mount.go
```diff
@@ -374,10 +374,15 @@ func parseEntryValues(ctx context.Context, e entry.Entry) (p []string, err error
...
+		if _, ok := seen[v]; ok {
+			log.Debugf(ctx, i18n.G("Value %q is duplicated."), v)
+			continue
+		}
+
 		// Compares "normal" and prefixed values the same way, since the unit name will be the same.
 		tmp := strings.TrimPrefix(v, krbTag)
 		if _, ok := seen[tmp]; ok {
-			log.Debugf(ctx, i18n.G("Value %q is duplicated. Tagged values are the same as untagged ones."), v)
+			log.Warningf(ctx, i18n.G("The location %q was already set up to be mounted. Authenticated and anonymous mounts are considered the same."), tmp)
```
didrocks, line 385: I would say "Authenticated and anonymous mounts are considered the same." -> "Only the first mount setting will be considered".

or something in more proper english which was say exactly which option (that is in %q, I agree) is considered.

internal/policies/mount/mount_test.go
```diff
@@ -115,15 +116,17 @@ func TestApplyPolicy(t *testing.T) {
...
+		"system, successfully apply policy for entry with repeated tagged values": {entries: []string{"entry with repeated tagged values"}, isComputer: true},
+		"system, successfully apply policy filtering out unsupported keys":        {entries: []string{"entry with multiple values", "entry with one value"}, keys: []string{"unsupported", "system-mounts"}, isComputer: true},
 
 		// Special cases.
 		"system, successfully apply policy with kerberos tagged values":                         {entries: []string{"entry with kerberos auth tags"}, isComputer: true},
 		"system, successfully apply policy prioritizing the first value found, despite the tag": {entries: []string{"entry with same values tagged and untagged"}, isComputer: true},
 		"system, only emit a warning when starting new units and systemctl fails":               {isComputer: true, firstSystemCtlFailingArgs: []string{"start"}},
+		"system, only emit a warning when disabling previous units and systemctl fails":         {isComputer: true, secondCall: []string{"entry with multiple values"}, secondSystemCtlFailingArgs: []string{"stop"}},
```
didrocks, line 129: I think the message is misleading, you want to tell that it’s stopping which fails

denisonbarbosa: oops

internal/policies/mount/mount.go
denisonbarbosa: I took 30 minutes to figure out a way to write this

didrocks: The issue here is the `tmp` might be the entry with krb5, and it was stripped before getting into seen, so telling %q was tmp is misleading…

didrocks: OR, you store `seen[tmp] = v`. that way, we have the initial value…

denisonbarbosa: That's why I'm trying to add this message that tagged and untagged are the same under our eyes... If I print the tagged value, it could create confusion just the same...

didrocks: Yes, but in the future, it might be other things, and "tag" and just an internal naming, no?

I would be really bare then, as we can get the initial value, and just telling "considering this value and ignoring that one", that way, it’s clear to the sysadmins as they put both values somewhere.

(but for that, we need seen to be a map[string][string]

didrocks (approved):
Once the tests pass +1 :) Thanks for fixing all those!

PR #561, opened 2023-01-04
Fix FTBFS on Launchpad

GabrielNagy:
More details in the individual commits. I still haven't been able to pinpoint the ppc64el `string.Contains` issue:

```
    log_test.go:84: 
        	Error Trace:	/<<PKGBUILDDIR>>/obj-powerpc64le-linux-gnu/src/github.com/ubuntu/adsys/internal/grpc/logstreamer/log_test.go:258
        	            				/<<PKGBUILDDIR>>/obj-powerpc64le-linux-gnu/src/github.com/ubuntu/adsys/internal/grpc/logstreamer/log_test.go:84
        	Error:      	"level=error msg=error log" does not contain "error log"
        	Test:       	TestAllLogLevels
        	Messages:   	Should contain substring
--- FAIL: TestAllLogLevels (0.00s)
```
I've tried byte comparison, duplicating the substring, running `strings.Contains` prior to the assert to no avail. Next step is to get access to an actual ppc64el VM for debugging.

### DIFF
--- a/internal/grpc/logstreamer/log_test.go
+++ b/internal/grpc/logstreamer/log_test.go
@@ -60,7 +60,10 @@ func TestAllLogLevels(t *testing.T) {
 	log.Info(stream.Context(), "info log")
 	log.Debug(stream.Context(), "debug log")
 	log.Warning(stream.Context(), "warning log")
-	log.Error(stream.Context(), "error log")
+
+	// Log an extra whitespace here because the substring assertion fails
+	// otherwise on the LP ppc64el builder
+	log.Error(stream.Context(), "error  log")
 
 	log.Infoln(stream.Context(), "infoln log")
 	log.Debugln(stream.Context(), "debugln log")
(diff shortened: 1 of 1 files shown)


internal/policies/apparmor/apparmor_test.go
```diff
@@ -42,32 +42,30 @@ func TestApplyPolicy(t *testing.T) {
 		removeUnusedAssetsError bool
 		apparmorParserError     string
 
+		integrationTest bool
```
didrocks, line 45: ouch, I missed that, I think we should really stay away of integration tests at this level and try to mock every commands we are using (contrary to real integration tests). Do you think we can mock the parser and control its behaviour as we do with other commands rather?

didrocks (changes requested):

internal/policies/apparmor/apparmor_test.go
didrocks: Hem, I don’t like we don’t understand that. Is it due to the parser talking with the kernel and this being denied? Do you mind having a look at this?

(Really, mixing integration tests here really strikes me as something we shouldn’t do for long term maintainance)

didrocks: phew! I really prefer that :)

didrocks (approved):
Wayyyyyyyyyyyyyyyyy better. Thanks for figuring those out! :)

PR #564, opened 2023-01-05
Updating debian/copyright to contain the new vendored sources

denisonbarbosa:


didrocks (approved):
I couldn’t check all files, but grabbed many and looked at them, this looks good! Thanks for doing this tedious work.

PR #567, opened 2023-01-06
Add winbind as an alternative dependency for adsys

GabrielNagy:
In case the user already has sssd or winbind installed, do not require them to install the other one as well. Prioritize sssd since that's the main backend we expect users to use.

DEENG-494


denisonbarbosa (approved):
Looks like a nice quality of life change. Well done!

Ps.: Maybe we should consider the use case of a user trying to switch the back end and it not being installed...

debian/control
```diff
@@ -32,8 +32,8 @@ Depends: ${shlibs:Depends},
          python3,
          python3-samba,
          samba-dsdb-modules,
-         sssd,
-         sssd-dbus,
+         sssd | winbind,
+         sssd-dbus | winbind,
          apparmor,
          cifs-utils,
          krb5-user,
```
didrocks, line 39: I think we don’t need krb5-user for sssd use case. Mind adding the alternative there too?

didrocks (changes requested):
One small dep to clarify that needs changing IMHO, otherwise, very good!

didrocks (approved):
Thanks a lot!

denisonbarbosa:
> > Looks like a nice quality of life change. Well done!
> > Ps.: Maybe we should consider the use case of a user trying to switch the back end and it not being installed...
> 
> Mmm if they consciously switch the backend they probably already understand that they need a different package. I don't think we can enforce this at the packaging level thinking

I wouldn't count on it... 😄 
But alas, I agree that there isn't much we can do about it.

PR #568, opened 2023-01-06
Fixing failing auto update Rust dependencies action

denisonbarbosa:
Action was failing with code 1, so I changed a few things in the bash code to reduce a few of the nested code that was there.

### DIFF
--- a/.github/workflows/auto-updates.yaml
+++ b/.github/workflows/auto-updates.yaml
@@ -131,8 +131,9 @@ jobs:
       - name: Update XS-Vendored-Sources-Rust
         run: |
           set -eu
-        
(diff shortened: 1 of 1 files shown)


.github/workflows/auto-updates.yaml
```diff
@@ -131,8 +131,10 @@ jobs:
       - name: Update XS-Vendored-Sources-Rust
         run: |
           set -eu
-        
-          OUTPUT=$((CARGO_VENDOR_DIR=vendor_rust/ /usr/share/cargo/bin/dh-cargo-vendored-sources 2>&1 || true) | grep XS-Vendored-Sources-Rust:)
+
+          export CARGO_VENDOR_DIR=vendor_rust/ 
+
```
didrocks, line 136: nitpick: I don’t think we need an empty line there, as it only applied to the command below, so let’s group them. After this, you can merge :)

didrocks (approved):

PR #570, opened 2023-01-09
Add integration tests for the winbind backend

GabrielNagy:
Refactor integration tests to support selecting different backends and add some basic tests for winbind.

We use the same libwbclient mocking logic as the winbind package tests.

DEENG-492

### DIFF
--- a/cmd/adsysd/integration_tests/adsys_test.go
+++ b/cmd/adsysd/integration_tests/adsys_test.go
@@ -79,7 +79,7 @@ func TestMain(m *testing.M) {
 func TestStartAndStopDaemon(t *testing.T) {
 	systemAnswer(t, "polkit_yes")
 
-	conf := createConf(t, "", "sssd")
+	conf := createConf(t)
 	quit := runDaemon(t, conf)
 	quit()
 }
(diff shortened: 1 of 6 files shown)


denisonbarbosa (approved):
Everything looks great to me. Well done!

cmd/adsysd/integration_tests/adsys_test.go
```diff
@@ -79,7 +79,7 @@ func TestMain(m *testing.M) {
 func TestStartAndStopDaemon(t *testing.T) {
 	systemAnswer(t, "polkit_yes")
 
-	conf := createConf(t, "")
+	conf := createConf(t, "", "sssd")
```
didrocks, line 82: This is set on every tests, and the number of induced changes triggers me that it’s maybe a little bit too much.

I think this is a good use case for optional function parameter? (Maybe doing the same for adsysDir too while we are at it). (Default being sssd ofc)

cmd/adsysd/integration_tests/adsys_test.go
```diff
@@ -228,7 +228,7 @@ policykit_dir: %s/polkit-1
 apparmor_dir: %s/apparmor.d/adsys
 apparmorfs_dir: %s/apparmorfs
 systemunit_dir: %s/systemd/system
-`, dir, dir, dir, dir, dir, dir, dir, dir, dir, dir)), 0600)
+`, dir, dir, dir, backend, dir, dir, dir, dir, dir, dir, dir)), 0600)
```
didrocks, line 231: dir dir dir dir dir oh backend, dir dir dir dir :p

cmd/adsysd/integration_tests/adsysctl_policy_test.go
```diff
@@ -904,7 +953,10 @@ func TestPolicyUpdate(t *testing.T) {
 					require.NoError(t, err, "Setup: could not create machine sss cache")
 				}
 				if krb5.src != "" {
-					krb5.src = filepath.Join(krb5currentDir, krb5.src)
+					// Don't modify absolute paths
+					if !strings.HasPrefix(krb5.src, "/") {
```
didrocks, line 957: You can use filepath.IsAbs() and then, don’t need the comment :p

cmd/adsysd/integration_tests/adsysctl_policy_test.go
```diff
@@ -1083,6 +1135,8 @@ func setupSubprocessForTest(t *testing.T, currentUser string, otherUsers ...stri
 	err := exec.Command("pkg-config", "--exists", "nss_wrapper").Run()
 	require.NoError(t, err, "libnss-wrapper is not installed on disk, either skip integration tests or install it")
 
+	mockWinbindLibPath := testutils.BuildWinbindMock(t, "../../../internal/ad/backends/winbind")
```
didrocks, line 1138: integration tests have a global variable (OMG!) named rootProjectDir :)

didrocks (changes requested):
Some small minor enhancements that can be added. Nice that things start to click like moving the Build utility to reuse in both case :) Well done!
Sad that we can’t really test kinit though. Did you try to build the package and autopkgtests? Some runs as root and it may impact if kinit is run on mock content?

didrocks (approved):
The changes LGTM then! It’s way more readable now with the conf using optionl types.

Thanks for checking on the autopkgtests and package build.

PR #571, opened 2023-01-10
Workaround Cargo.lock removal

denisonbarbosa:
Instead of completely removing the lockfile, we only move it instead. This still tricks Cargo and makes it build using the checksums of our local vendored dependencies. After the build process (for both the binary and the tests), we restore the lockfile to the upstream version.


denisonbarbosa:
Can you guys try to build it locally? The Rust part is working on my machine, but the Go tests freeze for some reason.

didrocks (approved):
One nitpick otherwise the rest looks good!

I’ll let Gabriel do the package build test with a failure in the middle of the """transaction"""

PR #572, opened 2023-01-11
Fix object name in mount tests

GabrielNagy:
This was always called with "ubuntu" - causing some tests to fail on autopkgtests, specifically the user not found/invalid user tests, as autopkgtests run under the "ubuntu" user.


denisonbarbosa (approved):
Oh, I didn't see that at all. Is this change enough to avoid breaking the autopkg tests? If you tested it and it fixes the problem, that's fine by me.
Nice job spotting this. Well done!

didrocks (approved):
obvious once you found it, and good! +1 :)

PR #580, opened 2023-01-13
Rust suggestions

denisonbarbosa:
Some fixes and improvements suggested in #444.

Thanks for the tips @liushuyu, they were all accepted and well needed.

### DIFF
--- a/internal/policies/mount/adsys_mount/Cargo.toml
+++ b/internal/policies/mount/adsys_mount/Cargo.toml
@@ -12,10 +12,10 @@ path = "src/lib.rs"
 
 [dependencies]
 anyhow = "^1"
-gio = "0.16"
-glib = "0.16"
-log = "0.4"
+gio = "=0.16.7"
+glib = "=0.16.7"
+log = "=0.4.17"
 serde = { version = "^1", features = ["derive"] }
 serde_json = "^1"
 thiserror = "^1"
-time = { version = "0.3", features = ["formatting", "macros"] }
+time = { version = "=0.3.17", features = ["formatting", "macros"] }
--- a/internal/policies/mount/adsys_mount/Cargo.toml
+++ b/internal/policies/mount/adsys_mount/Cargo.toml
@@ -11,11 +11,11 @@ name = "lib"
 path = "src/lib.rs"
 
 [dependencies]
-anyhow = "1.0.65"
-chrono = "0.4.22"
-gio = "0.16.2"
-glib = "0.16.2"
-log = "0.4.17"
-serde = { version = "1.0.147", features = ["derive"] }
-serde_json = "1.0.87"
-thiserror = "1.0.37"
+anyhow = "^1"
+gio = "0.16"
+glib = "0.16"
+log = "0.4"
+serde = { version = "^1", features = ["derive"] }
+serde_json = "^1"
+thiserror = "^1"
+time = { version = "0.3", features = ["formatting", "macros"] }
--- a/internal/policies/mount/adsys_mount/src/logger.rs
+++ b/internal/policies/mount/adsys_mount/src/logger.rs
@@ -1,4 +1,4 @@
-use chrono::Utc;
+use time::{self, macros::format_description, OffsetDateTime};
 
 use log::{Level, Log, Metadata, Record};
 /// Logger used by adsys_mount to provide context about the mount process.
@@ -15,12 +15,14 @@ impl Log for Logger {
             return;
         }
 
-        eprintln!(
-            "{} - {}: {}",
-            Utc::now().format("%d/%m/%Y %H:%M:%S"),
-            record.level(),
-            record.args()
-        );
+        let fmt_time = match OffsetDateTime::now_utc().format(format_description!(
+            "[day]/[month]/[year] [hour]:[minute]:[second]"
+        )) {
+            Ok(t) => t,
+            Err(_) => String::from("00/00/0000 00:00:00"),
+        };
+
+        eprintln!("{} - {}: {}", fmt_time, record.level(), record.args());
     }
 
     fn flush(&self) {}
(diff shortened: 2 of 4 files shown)


internal/policies/mount/adsys_mount/Cargo.toml
```diff
@@ -11,11 +11,11 @@ name = "lib"
 path = "src/lib.rs"
 
 [dependencies]
-anyhow = "1.0.65"
-chrono = "0.4.22"
```
didrocks, line 15: For minor unstable versions, I would keep the exact version.

didrocks (commented):
One small suggestion maybe on unstable version which can break for any minor point releases.

internal/policies/mount/adsys_mount/Cargo.toml
denisonbarbosa: You mean as in `gio = "0.16.x` rather than `gio = "0.16"`?
I have no idea how cargo decides this, to be honest. If it only fetches the versions which have docs, I think this should be safe.

internal/policies/mount/adsys_mount/src/logger.rs
```diff
@@ -15,12 +15,14 @@ impl Log for Logger {
...
-            record.args()
-        );
+        let fmt_time = match OffsetDateTime::now_utc().format(format_description!(
+            "[day]/[month]/[year] [hour]:[minute]:[second]"
+        )) {
+            Ok(t) => t,
+            Err(_) => String::from("00/00/0000 00:00:00"),
+        };
+
+        eprintln!("{} - {}: {}", fmt_time, record.level(), record.args());
```
liushuyu, line 25: For this pattern, I recommend using a short-hand form:

```suggestion
        eprintln!(
            "{} - {}: {}",
            OffsetDateTime::now_utc()
                .format(format_description!(
                    "[day]/[month]/[year] [hour]:[minute]:[second]"
                ))
                .unwrap_or_else(|_| "00/00/0000 00:00:00".to_string()),
            record.level(),
            record.args()
        );
```

`unwrap_or_else` is lazily evaluated (meaning it will only run the closure code when the condition is `Err`), so don't worry about whether it will cause performance issues.

internal/policies/mount/adsys_mount/Cargo.toml
liushuyu: I think this boils down to the upstream commitment to semantic versioning.

I think specifying `0.3` for the `time` crate is fine, and its upstream recommends doing that: https://time-rs.github.io/book/index.html.

As for `glib` and `gio` stuff, their documentation also seems to recommend specifying the minor version and omitting the patch version: https://gtk-rs.org/gtk-rs-core/stable/latest/docs/gio/.

denisonbarbosa: In this case, I would also rather have the version pinned. We can be a little more safer this way and keep some versioning consistency for the crates.

internal/policies/mount/adsys_mount/src/logger.rs
denisonbarbosa: Tbh, I think the shortened form is just less readable without achieving much else. In this cases, I'd rather go for readability instead of micro optimization.

liushuyu (approved):

internal/policies/mount/adsys_mount/src/logger.rs
schopin-pro: Readability is in the eye of the beholder.

Using raw `match` on Result or Option for such transformations is not really idiomatic, most Rustaceans would use and expect `unwrap_or` or `unwrap_or_else` here. My personal take is that if I `see unwrap_or*` I know that the `Ok` path is the expected, usual one, which my brain doesn't immediately see with a raw `match`. So, with that in mind, the first draft would give you this:

```rust

        let fmt_time = OffsetDateTime::now_utc()
            .format(format_description!(
                "[day]/[month]/[year] [hour]:[minute]:[second]"
            ))
            .unwrap_or(|_| "00/00/0000 00:00:00".into());

        eprintln!("{} - {}: {}", fmt_time, record.level(), record.args());
```

However, this code will give you the following `clippy` warning (done with the `rust-clippy` archive package):

```
warning: use of `unwrap_or` followed by a function call
 --> src/main.rs:9:10
  |
9 |         .unwrap_or("00/00/0000 00:00:00".into());
  |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: try this: `unwrap_or_else(|_| "00/00/0000 00:00:00".into())`
  |
  = note: `#[warn(clippy::or_fun_call)]` on by default
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#or_fun_call
```

I'm actually surprised that `cargo clippy --fix` didn't fix this automatically! Here's the fixed version, gone through `cargo fmt`.

```rust
        let fmt_time = OffsetDateTime::now_utc()
            .format(format_description!(
                "[day]/[month]/[year] [hour]:[minute]:[second]"
            ))
            .unwrap_or_else(|_| "00/00/0000 00:00:00".into());

        eprintln!("{} - {}: {}", fmt_time, record.level(), record.args());
```

So, what you see as a micro-optimization is just the natural consequence of the idiomatic use of Result methods + avoidance of a well-known anti-pattern.

Now, if you really want to go into the weeds of micro-optimization, you could even avoid the allocation in the fallback case (**please don't use this** ;-) ):

```rust
    let fmt_time = OffsetDateTime::now_utc()
        .format(format_description!(
            "[day]/[month]/[year] [hour]:[minute]:[second]"
        ));
    // temp values don't outlive the statement they're used from, hence needing to assign the initial Result<String, _> to a named variable.
    // That's OK though, name shadowing is embraced within Rust (maybe this is one of the reasons for it?)
    let fmt_time = fmt_time
        .as_ref()
        .map(|s| s.as_str())
        .unwrap_or("00/00/0000 00:00:00");
```

Again: please don't use that last one.

PR #583, opened 2023-01-16
Clean-up the packaging scripts in `user-mount-handler`

liushuyu:
This pull request utilizes a behavior in the Debian Cargo wrapper to workaround several issues where the Rust crate can not be compiled correctly due to non-overridable options in the said wrapper.

Also, use the real checksum in the facade manifest to avoid the need for updating `Cargo.lock` file.

### DIFF
--- a/debian/cargo-checksum.json
+++ b/debian/cargo-checksum.json
@@ -1,1 +1,1 @@
-{"package":"Could not get crate checksum","files":{}}
+{"package":"2955b1fe31e1fa2fbd1976b71cc69a606d7d4da16f6de3333d0c92d51419aeff","files":{}}
--- a/.github/workflows/auto-updates.yaml
+++ b/.github/workflows/auto-updates.yaml
@@ -162,3 +162,66 @@ jobs:
         # TODO: Change this when merging it into main
         run: |
           git push origin auto-update-rust-vendored-sources:user-mount-handler
+
+  update-fake-checksum-file:
+    name: 'Update the fake checksum file at d/cargo-checksum.json'
+    needs: update-readme-clid-ref
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          # TODO: Change this when merging it into main
+          ref: user-mount-handler
+      - name: 'Install GitHub script dependencies'
+        run: npm install toml
+      - name: 'Update the checksum file'
+        uses: actions/github-script@v5
+        id: update-checksum
+        with:
+          script: |
+            const toml = require("toml");
+            const fs = require("node:fs");
+            const checksum_path = "./debian/cargo-checksum.json";
+            /** @typedef LockEntry @prop {string} name @prop {string} checksum @prop {string} version */
+            /** Read and parse the lock file
+            * @param {string} path
+            * @returns {any} */
+            function readLock(path) {
+              const lock = fs.readFileSync(path, { encoding: "utf-8" });
+              return toml.parse(lock);
+            }
+
+            const lockfile = readLock("./Cargo.lock");
+            /** @type {LockEntry?} */ const sd_block = lockfile.package.find(
+              (/** @type {LockEntry} */ x) => x.name === "system-deps"
+            );
+            if (!sd_block) throw new Error("system-deps not found in the lock file");
+            const old_cksum_file = require(checksum_path);
+            if (old_cksum_file.package === sd_block.checksum) {
+                console.info("Checksum is up-to-date. No update needed.");
+                return false;
+            }
+            const new_cksum_file =
+              JSON.stringify({
+                package: sd_block.checksum,
+                files: {},
+              }) + "\n";
+            fs.writeFileSync(checksum_path, new_cksum_file);
+            console.info("Checksum file updated.");
+            return true;
+      - name: Create Pull Request
+        if: ${{ steps.update-checksum.outputs.result == 'true' }}
+        # V5 Beta needed because of https://github.com/peter-evans/create-pull-request/issues/1170
+        uses: peter-evans/create-pull-request@v5-beta
+        with:
+          commit-message: Auto update Rust vendored sources checksums
+          title: Auto update Rust vendored sources checksums
+          labels: control, automated pr
+          body: "[Auto-generated pull request](https://github.com/ubuntu/adsys/actions/workflows/auto-updates.yaml) by GitHub Action"
+          branch: auto-update-rust-checksum
+          token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Push branch
+        if: ${{ steps.update-checksum.outputs.result == 'true' }}
+        # TODO: Change this when merging it into main
+        run: |
+          git push origin auto-update-rust-checksum:user-mount-handler
--- a/.gitignore
+++ b/.gitignore
@@ -38,3 +38,8 @@ debian/adsys-windows
 /adsysd
 /adsysctl
 /adsys.yaml
+
+# GitHub CI temporary files
+node_modules
+package-lock.json
+package.json
--- a/debian/rules
+++ b/debian/rules
@@ -50,6 +50,9 @@ override_dh_auto_clean:
 	# system-deps crate includes an empty .a file which is filtered out when
 	# building the source package and cargo gets confused when trying to verify
 	# integrity. So we just put in an empty checksum file instead.
+	# The checksum in the file comes from the Cargo.lock file under the `system-deps`
+	# dependency segment. This fake checksum file needs to be updated after each
+	# adsys update.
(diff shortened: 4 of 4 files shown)


debian/cargo-checksum.json
```diff
@@ -1 +1 @@
-{"package":"Could not get crate checksum","files":{}}
+{"package":"2955b1fe31e1fa2fbd1976b71cc69a606d7d4da16f6de3333d0c92d51419aeff","files":{}}
```
schopin-pro, line 1: How do you compute the checksum? That's something I was unable to find!

liushuyu: > How do you compute the checksum? That's something I was unable to find!

That's easy. You don't actually need to compute anything:

https://github.com/ubuntu/adsys/blob/dc51c88f909be948ce30379c9ed282850d1201d7/Cargo.lock#L373-L377

schopin-pro: You might want to document this, then, as it's a good way to have a hard-to-understand FTBFS in the future. Maybe a comment in d/rules ?

liushuyu: > You might want to document this, then, as it's a good way to have a hard-to-understand FTBFS in the future. Maybe a comment in d/rules ?

Okay, I will add a comment in `d/rules` to explain this.

denisonbarbosa: Even though I like seeing the rules file clean again, the scalability of this suggestion worries me. Right now, this is a problem related only to the system-deps dependency, but this can change. I'm definitely not an expert, but I'm assuming this checksum could change if the version changes and then we would need to update this yet again.
Yes, it's ugly. But tricking cargo into thinking there is no lockfile at build time still seems to be the better solution to allow build and still scale.

liushuyu: Done. Comment addressed: added new comments to explain some of the more obscure workarounds.

liushuyu: > Even though I like seeing the rules file clean again, the scalability of this suggestion worries me. Right now, this is a problem related only to the system-deps dependency, but this can change. I'm definitely not an expert, but I'm assuming this checksum could change if the version changes and then we would need to update this yet again.
> Yes, it's ugly. But tricking cargo into thinking there is no lockfile at build time still seems to be the better solution to allow build and still scale.

We can see how this goes in the future: if this becomes unmanageable, we can restore the original behaviour.

After going through all the dependencies of this package, I think the only issue will probably be `gio` and `glib` (because `system-deps` is a transitive dependency of them), and they don't add dependencies that often. I believe this is going to be fine for quite a while.

liushuyu:
> Do you think this is something we can do there too?

I think this is a yes!


> and then extracting the checksum to refresh `debian/rules`

I believe this is possible.

I can definitely look into this, do you want me to roll this CI change into this pull request or I should open a new one?

didrocks:
> I can definitely look into this, do you want me to roll this CI change into this pull request or I should open a new one?

Oh excellent! I think this PR is a great opportunity to fix it for good :) So if you can add it here. My only request is to append this as a job in `[auto-updates.yaml](https://github.com/ubuntu/adsys/blob/main/.github/workflows/auto-updates.yaml)` to ensure all our "udpates" automation are in a single workflow (also, for obvious concurrency reasons when committing+pushing the branch, we don’t want the jobs to run in parallel)

liushuyu:
> Oh excellent! I think this PR is a great opportunity to fix it for good :) So if you can add it here. My only request is to append this as a job in [auto-updates.yaml](https://github.com/ubuntu/adsys/blob/main/.github/workflows/auto-updates.yaml) to ensure all our "udpates" automation are in a single workflow (also, for obvious concurrency reasons when committing+pushing the branch, we don’t want the jobs to run in parallel)

Done. A new CI job has been added. I can't test this job in my repository since it requires the same branch layout as this repository.

debian/cargo-checksum.json
denisonbarbosa: Yes, I think that with the action this should be fine, specially since I don't think there will be more Rust on adsys anytime soon and until there we can probably hope that cargo provides us with a bit more overridable options for this stuff.
Thanks for taking some time with this!

denisonbarbosa (approved):
This looks very good to me! Thanks for taking some time on this, it's definitely more readable and less "workaroundy" than it was before, I appreciate it!

didrocks:
I am still a little bit puzzled on the reason we need that file (as stated in the Debian Rust Policy, this seems to be only for library crates: https://wiki.debian.org/Teams/RustPackaging/Policy), and why only for system-deps crate), but I think we can live with it for now :)

Thanks for the action, it seems fine with me and should work as expected :))

However, I did try to build a source package and upload it to my ppa to test a package build and some of your changes 
 FTBFS:error: failed to open: /usr/.crates.toml. Looks like some paths are not fully set? Do you mind having a look?

I have just gbp buildpackage -S and dput the .changes file to mine.

# Here we utilise a logical flaw in the cargo wrapper to our advantage:
# when specifying DEB_CARGO_CRATE_IN_REGISTRY=1, the wrapper will
# avoid adding the `--path` option, so that we can specify it ourselves
DEB_HOST_GNU_TYPE=x86_64-linux-gnu \
DEB_HOST_RUST_TYPE=x86_64-unknown-linux-gnu \
CARGO_HOME=/<<PKGBUILDDIR>>/debian/cargo_home \
DEB_CARGO_CRATE=adsys_mount_0.1.0 \
DEB_CARGO_CRATE_IN_REGISTRY=1 \
        /usr/share/cargo/bin/cargo install --path internal/policies/mount/adsys_mount
debian cargo wrapper: options, profiles, parallel: ['noautodbgsym', 'parallel=4'] ['noudeb'] ['-j4']
debian cargo wrapper: rust_type, gnu_type: x86_64-unknown-linux-gnu, x86_64-linux-gnu
debian cargo wrapper: installing into destdir '' prefix '/usr'
debian cargo wrapper: running subprocess (['env', 'RUST_BACKTRACE=1', 'CARGO_TARGET_DIR=/<<PKGBUILDDIR>>/target', '/usr/bin/cargo', '-Zavoid-dev-deps', 'install', '--verbose', '--verbose', '-j4', '--target', 'x86_64-unknown-linux-gnu', '--path', 'internal/policies/mount/adsys_mount', 'adsys_mount', '--vers', '0.1.0', '--root', '/usr'],) {'check': True}
error: failed to open: /usr/.crates.toml

Caused by:
  Permission denied (os error 13)
Traceback (most recent call last):
  File "/usr/share/cargo/bin/cargo", line 241, in <module>
    sys.exit(main(*sys.argv[1:]))
             ^^^^^^^^^^^^^^^^^^^
  File "/usr/share/cargo/bin/cargo", line 231, in main
    return install(os.getenv("DESTDIR", ""),
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/share/cargo/bin/cargo", line 134, in install
    logrun(["env", "RUST_BACKTRACE=1",
  File "/usr/share/cargo/bin/cargo", line 76, in logrun
    return subprocess.run(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/subprocess.py", line 571, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['env', 'RUST_BACKTRACE=1', 'CARGO_TARGET_DIR=/<<PKGBUILDDIR>>/target', '/usr/bin/cargo', '-Zavoid-dev-deps', 'install', '--verbose', '--verbose', '-j4', '--target', 'x86_64-unknown-linux-gnu', '--path', 'internal/policies/mount/adsys_mount', 'adsys_mount', '--vers', '0.1.0', '--root', '/usr']' returned non-zero exit status 101.
make[1]: *** [debian/rules:95: override_dh_auto_install] Error 1
make[1]: Leaving directory '/<<PKGBUILDDIR>>'
make: *** [debian/rules:43: binary] Error 2
dpkg-buildpackage: error: fakeroot debian/rules binary subprocess returned exit status 2


Full loogs on https://launchpadlibrarian.net/646840241/buildlog_ubuntu-lunar-amd64.adsys_0.9.2.test1_BUILDING.txt.gz

liushuyu:
> Full logs on https://launchpadlibrarian.net/646840241/buildlog_ubuntu-lunar-amd64.adsys_0.9.2.test1_BUILDING.txt.gz

I think I made a mistake when committing the changes, and I have now corrected that. Can you try again?

didrocks (approved):
Yeah! Everything builds now and pass. Awesome work! Thanks a lot for your contribution which will make updating those files and keeping upstream versions as the deciding factors without compromising the source tree.

Approving and merging

PR #585, opened 2023-01-20
Add proxy manager that sets environment proxy configuration

GabrielNagy:
Implement an API that will be used by a future proxy policy manager and a CLI program to set the proxy configuration on the system.

Currently it only supports system-wide proxy configuration via an environment file under `/etc/environment.d`, but in the future it will also configure APT and GSettings proxy settings.

DEENG-553


internal/proxymanager/proxymanager.go
```diff
@@ -0,0 +1,117 @@
...
+
+	return &Manager{
+		proxies:               proxies,
+		aptConfigPath:         args.aptConfigPath,
+		environmentConfigPath: args.environmentConfigPath,
+	}, nil
+}
+
+// Apply applies the proxy configuration to the system.
+func (m *Manager) Apply(ctx context.Context) error {
```
denisonbarbosa, line 61: I think this should be named `ApplyPolicy` to be in line with our other managers.

internal/proxymanager/config.go
```diff
@@ -0,0 +1,19 @@
+package proxymanager
+
+// Config is the exported proxy configuration type for users of the proxymanager API.
+type Config struct {
+	HTTPProxy  string
+	HTTPSProxy string
+	FTPProxy   string
+	SocksProxy string
+	NoProxy    string
```
denisonbarbosa, line 9: I got slapped by the `clippy` tool in Rust when I set up a Struct like this as well. It told me something like `if you are (post | pre) fixing every field in here with the same word, there's probably a better way to write it`.
Even though this is not Rust, I quite liked the suggestion and maybe we can figure something out here?

internal/proxymanager/proxymanager.go
```diff
@@ -0,0 +1,117 @@
...
+	if exists, prevContent, err := prevConfIfExists(m.environmentConfigPath); exists && prevContent == content {
+		log.Debugf(ctx, fmt.Sprintf("Environment proxy configuration at %q is already up to date", m.environmentConfigPath))
+		return nil
+	} else if err != nil {
+		return err
+	}
+
+	// Check if the parent directory exists - attempt to create the structure if not
+	environmentConfigDir := filepath.Dir(m.environmentConfigPath)
+	if _, err := os.Stat(filepath.Dir(m.environmentConfigPath)); errors.Is(err, os.ErrNotExist) {
```
denisonbarbosa, line 89: You already declared the `environmentConfigDir` var above with the `filepath.Dir` returned value, there's no need to call it again inside `os.Stat`

internal/proxymanager/proxymanager_test.go
```diff
@@ -0,0 +1,160 @@
...
+	for name, tc := range tests {
+		tc := tc
+		name := name
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+
+			_, err := proxymanager.New(context.Background(), tc.config)
+			if tc.wantErr {
+				// We don't return here as we want to check that the apparmor
+				// dir is in the expected state even in error cases
```
denisonbarbosa, line 46: Some leftover from a ctrl+c?

internal/proxymanager/proxymanager_test.go
```diff
@@ -0,0 +1,160 @@
...
+				require.Error(t, err, "New should have failed but didn't")
+				return
+			}
+			require.NoError(t, err, "New failed but shouldn't have")
+		})
+	}
+}
+
+func TestApply(t *testing.T) {
```
denisonbarbosa, line 55: Test name will change as well if you agree with the function naming comment.

internal/proxymanager/proxymanager_test.go
```diff
@@ -0,0 +1,160 @@
...
+			NoProxy: "localhost,127.0.0.1",
+		}},
+		"Password is escaped":                          {config: proxymanager.Config{HTTPProxy: "http://username:p@$$:w0rd@example.com:8080"}},
+		"Escaped password is not escaped again":        {config: proxymanager.Config{HTTPProxy: "http://username:p%40$$%3Aw0rd@example.com:8080"}},
+		"Domain username is escaped":                   {config: proxymanager.Config{HTTPProxy: `http://EXAMPLE\bobsmith:p@$$:w0rd@example.com:8080`}},
+		"Domain username without password is escaped":  {config: proxymanager.Config{HTTPProxy: `http://EXAMPLE\bobsmith@example.com:8080`}},
+		"Escaped domain username is not escaped again": {config: proxymanager.Config{HTTPProxy: `http://EXAMPLE%5Cbobsmith@example.com:8080`}},
+
+		// Error cases
+		"Environment config file is read-only": {readOnlyFiles: []string{envConfigPath}, prevContents: map[string]string{envConfigPath: "something"}, wantErr: true},
```
denisonbarbosa, line 100: Don't forget that our error cases are supposed to be prefixed with `Error when ...`, `Error on ...`, etc.

internal/proxymanager/proxysetting.go
```diff
@@ -0,0 +1,159 @@
...
+
+func escapeURLUsername(uri string) string {
+	// Attempt to unescape the string first, discarding any error
+	// At best, this prevents us from escaping the URL multiple times
+	// At worst, the URL is not affected (we will treat % signs as part of the
+	// credentials and escape them later)
+	uri, _ = url.PathUnescape(uri)
+
+	// Regexp to check if the URI contains credentials
+	r := regexp.MustCompile(`^\w+://(?:(?P<credentials>.*:?.*)@)[a-zA-Z0-9.-]+(:[0-9]+)?/?$`)
```
denisonbarbosa, line 146: I don't know if using something that could panic is a good approach here, maybe using `regexp.Compile` and treating the error is better.

denisonbarbosa (changes requested):
Looks very nice, despite a few comments. Well done!

PR #590, opened 2023-01-24
Fix for auto update Rust actions

denisonbarbosa:
Fixing Rust auto update actions


didrocks (approved):
Thanks for fixing the CI!

PR #610, opened 2023-02-24
Fix erroneous non alternative dep on krb5-user

didrocks:
The commit introducing it does not mention krb5-user, it was probably an incorrect rebase on main.


jibel (approved):

didrocks:
yes, it’s blocked in proposed due to that and some autopkgtests failing, we will discuss it during our meeting :)

PR #614, opened 2023-02-27
Replacing os.WriteFile after init in the config tests

denisonbarbosa:
Viper relies on fsnotify to trigger a config update. However, the fsnotify docs state that the Write event is also triggered by truncating the file. This means that the update callback was being called too soon and we were unmarshaling the file contents before they were properly written.
To avoid this problem, the os.WriteFile call had to be replaced and now we manually open the file, write to it and then close it to properly flush the changes and only then have the callback triggered.

DEENG-607

### DIFF
--- a/internal/config/config_test.go
+++ b/internal/config/config_test.go
@@ -312,8 +312,16 @@ func TestInit(t *testing.T) {
 				callbackPhase = 2
 				phaseMu.Unlock()
 
-				err = os.WriteFile(filepath.Join(configDir, prefix+".yaml"), []byte(tc.changeConfigWith), 0600)
-				require.NoError(t, err, "Setup: failed to write initial config file")
+				// fsnotify docs states that truncating a file will also trigger a Write event, which means that the callback can be triggered
+				// before the file contents are properly written. In order to avoid that, we must manually open the file and write the new content
+				// to it to avoid truncating the file and triggering the event too early.
+				configFile, err := os.OpenFile(filepath.Join(configDir, prefix+".yaml"), os.O_RDWR|os.O_CREATE, os.ModePerm)
+				require.NoError(t, err, "Setup: failed to open config file")
+				defer configFile.Close()
+				_, err = configFile.WriteString(tc.changeConfigWith)
+				require.NoError(t, err, "Setup: failed to write new content to config file")
+				configFile.Close()
+
 				// Let’s force a sync to make sure the file is written on disk
 				syscall.Sync()
 				select {
--- a/internal/config/config_test.go
+++ b/internal/config/config_test.go
@@ -96,6 +96,8 @@ func TestInit(t *testing.T) {
 		notInConfigDir     bool
 		changeConfigWith   string
 
+		directlyChangeConfigFile bool
+
 		subcommand             bool
 		configFlagOnSubcommand bool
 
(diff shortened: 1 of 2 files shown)


CarlosNihelton (approved):
Straightforward fix. LGTM 👍🏽

internal/config/config_test.go
```diff
@@ -312,8 +312,15 @@ func TestInit(t *testing.T) {
...
 				phaseMu.Unlock()
 
-				err = os.WriteFile(filepath.Join(configDir, prefix+".yaml"), []byte(tc.changeConfigWith), 0600)
-				require.NoError(t, err, "Setup: failed to write initial config file")
+				// fsnotify docs states that truncating a file will also trigger a Write event, which means that the callback can be triggered
+				// before the file contents are properly written. In order to avoid that, we must manually open the file and write the new content
+				// to it to avoid truncating the file and triggered the event too early.
+				configFile, err := os.OpenFile(filepath.Join(configDir, prefix+".yaml"), os.O_RDWR|os.O_CREATE, os.ModePerm)
+				require.NoError(t, err, "Setup: failed to open config file")
+				_, err = configFile.WriteString(tc.changeConfigWith)
```
didrocks, line 320: If write faiols here, the file is not closed, You still need to defer the Close after the first require.NoError in addition to the manual close.

didrocks (changes requested):
As we discussed yesterday, nice work and hunting!
One change requested to ensure the file is always closed.

However, we also mentionned yesterday to open a file in viper for listening to the rename event if this is missed and referenced it here. I think this should be done before merging.

denisonbarbosa:
> However, we also mentionned yesterday to open a file in viper for listening to the rename event if this is missed and referenced it here. I think this should be done before merging.

@didrocks This is a bit tricky... The `fsnotify.Rename` event is actually two events at once:
```Go
//   fsnotify.Rename    A path was renamed. A rename is always sent with the
//                      old path as Event.Name, and a Create event will be
//                      sent with the new name. Renames are only sent for
//                      paths that are currently watched; e.g. moving an
//                      unmonitored file into a monitored directory will
//                      show up as just a Create. Similarly, renaming a file
//                      to outside a monitored directory will show up as
//                      only a Rename.
```
Technically, viper is watching for the rename event indirectly (since the watch is done in the directory and renaming a file in the directory will count as a `fsnotify.Create` event). So, the problem with the .new strategy is actually in our code (on line 82 of config.go):
```Go
if e.Op != fsnotify.Write {
    return
}
```
We are only considering `Write` events, which means that the `Create` event is discarded and our tests fail (when trying to rename the file).

didrocks:
Ah, you mentionned yesterday it was viper not listening to Rename, so maybe we can fix our code if we are the one gating on it and go with the .New() strategy?

denisonbarbosa:
> Ah, you mentionned yesterday it was viper not listening to Rename, so maybe we can fix our code if we are the one gating on it and go with the + ".new" strategy?

Sure, we can. However, shouldn't we keep the "changing the same file" behavior since it's the one that's more likely to happen in a real use case?

didrocks:
> However, shouldn't we keep the "changing the same file" behavior since it's the one that's more likely to happen in a real use case?

Good point, you’re right. Actually, hem, 2 tests? :)

internal/config/config_test.go
```diff
@@ -149,8 +151,12 @@ func TestInit(t *testing.T) {
 		},
 
 		// Configuration changes support
-		"Configuration changed": {
-			configFileContent: "value: filecontentvalue", changeConfigWith: "value: filecontentvaluerefreshed",
+		"Configuration changed directly": {
+			configFileContent: "value: filecontentvalue", changeConfigWith: "value: filecontentvaluerefreshed", directlyChangeConfigFile: true,
+			want: "filecontentvaluerefreshed", wantCallbackCalled: 2,
+		},
+		"Configuration changed indirectly": {
```
didrocks, line 158: Configuration file overwritten?

didrocks (approved):
Excellent! Just rename the new test case itself to be more descriptive and then, we are good to go! :)

PR #631, opened 2023-03-14
Adding guideline docs to policy managers

### LINKED ISSUE #528: Write documentation on managers.go on error handling in manager
General idea is:
- things that impacts generation of files on disk / policy application / configuration -> error on first err returned
- things that make some execution failing out of our control (unit starting, scripts failing) -> warnings.

We apply a configuration policy, not enforce execution.

Write this doc and guidelines in managers.go.

denisonbarbosa:
When creating the mount policy manager, we realized that we needed a defined way of handling errors in the managers. Which errors should prevent authentication? Which errors should only be warned to the user?

After some discussion, we defined an expected general behavior that all managers should have and this must documented on the general manager, defined in manager.go

Closes #528
DEENG-539

### DIFF
--- a/internal/policies/apparmor/apparmor.go
+++ b/internal/policies/apparmor/apparmor.go
@@ -1,4 +1,21 @@
 // Package apparmor provides a manager to apply apparmor policies.
+//
+// The policy manager first checks if apparmor_parser is available
+// (file exists and is executable) and proceeds differently if these
+// requirements are not met, depending on whether there are configured entries in
+// the GPO:
+// - no entries: a warning is logged and the manager returns without error
+// - entries: the manager returns an error if apparmor_parser is not found
+//
+// Next, if we found no entries to apply (either to them not existing or being
+// disabled), we attempt to unload all rules managed by ADSys.
+//
+// If there are entries to apply, based on the object type (machine or user), we
+// attempt to apply them. This process is more clearly outlined in the
+// ApplyPolicy function documentation.
+//
+// If any errors occur during the policy apply process, the manager will attempt
+// to restore the initial state of the system before returning an error.
 package apparmor
 
 import (
--- a/internal/policies/dconf/dconf.go
+++ b/internal/policies/dconf/dconf.go
@@ -1,4 +1,28 @@
 // Package dconf is the policy manager for dconf entry types.
+//
+// The manager will create additional dconf databases: system-db:<username> and system-db:machine.
+// Values specified in the GPO will be appended to those databases, according to their configuration.
+// Dconf applies the values from bottom to top and stops checking values for a certain key at the
+// moment it finds a lock.
+// Default values specified by the policy will be added to the profile database, along with locks to
+// their correspondent keys, in order to enforce the requested values.
+//
+// The manager will parse the values and try to fix some formatting problems, but if something goes
+// wrong when applying the profile or updating dconf, an error is returned.
+// However, ADSys will not check for the correctness of the values being assigned and it's up to the
+// admin to ensure that the requested value is assignable to the key it is being assigned to.
+//
+// Notes or common keys between user and machine:
+//
+// 1. Machine is not configured (no value, no lock) -> upper layers will be taken into account, which can be the user
+// one or user default value.
+//
+// 2. Machine is configured (value, and lock) -> the lock will "stick" our desired value and enforce it.
+//
+// 3. Machine configuration is set to deleted, conveying "I want the default value from the system" (no value, and lock)
+// -> the lock will "stick" the desired value to the layer of current value of Machine. As machine doesn’t have any
+// value and is the lowest in the stack (the first one to be processed), this will thus enforce the default system
+// configuration for that setting.
 package dconf
 
 import (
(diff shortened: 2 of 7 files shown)


internal/policies/manager.go
```diff
@@ -23,6 +23,32 @@ import (
 	"golang.org/x/sync/errgroup"
 )
 
+/*
+
+ADSys is expected to apply configuration policies, not enforce them. We are responsible for
```
didrocks, line 28: I would put that on top, so that they show up as package documentation (https://go.dev/doc/comment).

didrocks (changes requested):
Looking good! Nothing to add on the wording.

I think we should just move them as package documentation so that godoc can show them.
On this card, I was additionall mapping this for every kind of manager that we describe that can error out in ApplyPolicy() or not. Do you mind creating a separate card for that to ensure every manager documents them?

internal/policies/apparmor/apparmor.go
```diff
@@ -1,4 +1,20 @@
 // Package apparmor provides a manager to apply apparmor policies.
+//
+// The policy manager first checks for the existence of the apparmor_parser
+// executable, and proceeds differently if the executable is not found,
+// depending on whether there are confingured entries in the GPO:
+// - no entries: a warning is logged and the manager returns without error
+// - entries: the manager returns an error if apparmor_parser is not found
```
didrocks, line 7: I’m unsure we need to precise that, because you say that "If any errors occur during the policy apply process, the manager will attempt…" so if apparmor_parser is not found, this falls under this case. I would just remove this part.

internal/policies/dconf/dconf.go
```diff
@@ -1,4 +1,28 @@
...
+//
+// The manager will create additional dconf databases: system-db:<username> and system-db:machine.
+// Values specified in the GPO will be appended to those databases, according to their configuration.
+// Dconf applies the values from bottom to top and stops checking values for a certain key at the
+// moment it finds a lock.
+// Default values specified by the policy will be added to the profile database, along with locks to
+// their correspondent keys, in order to enforce the requested values.
+//
+// The manager will parse the values and try to fix some formatting problems, but if something goes
+// wrong when applying the profile or updating dconf, authentication is prevented.
```
didrocks, line 11: I would say, instead of "authentication is prevented", because we know this is the general behavior of a policy manager when we return an error (and it’s handled above in the stack), that we return an error (the consequence is implicit).

internal/policies/manager.go
```diff
@@ -1,3 +1,25 @@
...
+// properly configuring those policies and set up the machine to apply them, but we can not ensure that
+// they will be executed as this relies on lots of variables and system functionalities.
+// As such, policy managers are expected to only prevent authentication when something goes wrong
+// during the configuration of the policies, whilst ensuring that the user will be warned should a
+// configuration fail to execute.
+//
+// We should prevent authentication on errors such as:
+//   - failed to copy the policy requested assets to the machine;
+//   - failed to parse policy configuration values;
+//   - failed to write the requested files on disk;
```
didrocks, line 13: We can add a 4th case: missing binary or binary execution to enforce a policy not present (dconf update, apparmor_parser not present…)

internal/policies/mount/mount.go
```diff
@@ -1,7 +1,16 @@
...
-// computer mounts will be handled directly by systemd, via mount units.
+// Package mount provides the policy manager to handle file sharing policies.
+//
+// The manager behavior differs depending on the object type.
+// For system mounts, it will parse the requested values and create systemd mount units to mount the
+// requested shared locations.
+// For user mounts, it will parse the requested values, store them in a separate file which will be
+// handled by a helper binary that will mount the requested shared locations using gio.
+//
+// Should the manager fail to setup the policy with the requested entries and its values,
```
didrocks, line 9: Do you mind expanding how this setup can fail, it’s only writing files on disk?

internal/policies/mount/mount.go
```diff
@@ -1,7 +1,16 @@
...
+// an error will be returned and the login is prevented.
+// However, if the manager creates the files needed and setup all the required steps, it's up to
+// the correctness of the specified entries values and gvfs to mount the requested shared drives.
+// Should an error occur during this step, ADSys will log it without preventing the authentication.
```
didrocks, line 13: Well, it’s rather through systemd (mount units for system or our excellent Rust binary for users) logs that our admin will access it? (so that she/he doesn’t look at the adsys logs and don’t see it).

internal/policies/scripts/scripts.go
```diff
@@ -1,6 +1,16 @@
...
 //
-// It handles also the script assets themselves.
+// This manager configures the scripts that will be executed in the following steps:
+//   - machine: startup and shutdown;
+//   - user: login and logout;
+//
+// The manager will download the requested assets and set up the scripts to be executed during the
+// mentioned steps. The machine scripts will be executed by ADSys itself and the user ones will be
+// handled by systemd through user units.
+// If the manager fails to configure the required assets, the applying process will fail and
```
didrocks, line 10: s/configure/download and find/

didrocks (changes requested):
Excellent (and very needed) work! I have some suggestions of things to add to improve the content of it, but it will be really helpful in the long term IMHO. :)

internal/policies/apparmor/apparmor.go
didrocks: yes, but what if the binary is there, but not executable, should we specify it to explicitly or that goes rather to the "if apply fails" dogma? :)

denisonbarbosa: We check it's existence with `exec.LookPath()` which explicitly mentions that it looks for an executable file, so even if the binary is there but isn't executable we exit without doing anything.
I agree that we should add this to the doc as well, I'll get on with it.

didrocks (approved):

PR #632, opened 2023-03-14
Replace testutils.Setenv

### LINKED ISSUE #608: Remove testutils.Setenv
Go has a t.SetEnv() from 1.17, remove our testutils function.

denisonbarbosa:
Since Go 1.17, the type `testing.T` has a method [t.Setenv](https://pkg.go.dev/testing#T.Setenv), which does precisely what we used to do in `testutils.Setenv`, so we should refactor the code to use the offered tools.

Closes #608 
DEENG-604


didrocks (approved):
Always good to see less code with passing tests :)

PR #634, opened 2023-03-14
Use SetArgs with variadic of strings instead of changeArgs

### LINKED ISSUE #595: Use SetArgs in tests
In cmd/adsysd (and probably adsysctl), remove changeArgs and use SetArgs, taking a variadic instead.

denisonbarbosa:
Closes #595 
DEENG-573


didrocks (commented):
The goal of that change was to run more tests in parallel. Do you mind trying to check which ones we can and annotate them?

(all the ones which don’t touch stderr/stdout/stdin are good candidates now)

PR #636, opened 2023-03-14
Fixing crates checksum issue

denisonbarbosa:
Our previous strategy of copying a fake checksum file to the problematic crate was unsustainable, since dependencies can change (either directly or indirectly) when updating our code (or our dependencies version) and new crates with .a files could emerge.
To avoid such problems in the future, the idea is to simplify the checksums when building the source package. This way, we can still use the general crate checksum while also being able to build the package without the .a files.


didrocks (approved):
Ok, porting that from aad-auth to adsys looks good! :)

didrocks (approved):

PR #637, opened 2023-03-15
Add system proxy policy manager

GabrielNagy:
This PR contains ADMX/ADML keys, the system proxy manager implementation and packaging updates. More details are in the commits themselves.

### DIFF
--- a/policies/Ubuntu/all/Ubuntu.adml
+++ b/policies/Ubuntu/all/Ubuntu.adml
@@ -26,6 +26,7 @@
       <string id="UbuntuDisplaySystemWideApplicationConfinement">System-wide application confinement</string>
       <string id="UbuntuDisplayPowerManagement">Power Management</string>
       <string id="UbuntuDisplaySystemDriveMapping">System Drive Mapping</string>
+      <string id="UbuntuDisplaySystemProxyConfiguration">System proxy configuration</string>
       <string id="UbuntuDisplaySessionManagement">Session management</string>
       <string id="UbuntuDisplayUserScripts">User Scripts</string>
       <string id="UbuntuDisplayUserApplicationConfinement">User application confinement</string>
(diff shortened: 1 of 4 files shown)


denisonbarbosa (commented):
A few comments, but it's looking really nice. Well done!

internal/policies/proxy/proxy.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	dpkgStatusCmd := append(m.dpkgCmd, "ubuntu-proxy-manager")
+
+	smbsafe.WaitExec()
+	// #nosec G204 - We are in control of the arguments
+	out, err := exec.CommandContext(ctx, dpkgStatusCmd[0], dpkgStatusCmd[1:]...).CombinedOutput()
```
didrocks, line 109: This is nice, but I wouldn’t add really dpkg knowledge here to the code base. I would just execute the dbus call (as if it was any other binaries) and returns an errors.

Or, if you want to be nicer as it’s a suggest, you can try to see if the hardcoded path is there, but really… unsure. I think the documentation can cover it.

It’s a little bit similar than systemd (even if I doubt we will run on Devuan anytime soon), or without dconf, for servers…

didrocks (changes requested):
I have one modification request, otherwise, excellent work! :)

You are going to deal with integration tests with a separate Jira card/PR, right? (which, contrary to this local mock, will need then to create files on disk to assert on them)

didrocks:
> I was going to ask about that. Initially I wanted to handle them as part of this PR but then I realized that the functional part of this manager lives in ubuntu-proxy-manager. How do you suggest we handle this coupling? Should we install the package as part of the integration tests? If we do we probably want to make some stuff like the "root" configuration [here](https://github.com/ubuntu/ubuntu-proxy-manager/blob/d245cfeab273c06339f3fbe972d7ad634ecf1850/internal/proxy/proxy.go#L55) (and maybe the timeout) externally configurable so we can override them via CLI.

This is really as you wish, you can either create a mock in the docker containers exposing the dbus API and acting on it, or you can expose the real thing there, and then, try to trigger some main use case (installed/not installed). I wouldn’t go as far as triggering the timeout, this is "dbus returned an error to us" from a black box perspective and we should just rely on this error.

didrocks (commented):
Excellent, thanks for the integration tests too! Ack on the approach.

I think you only updated C4F393CA-AD9A-4595-AEBC-3FA6EE484285 in up to date state because the GPO is corrupted on purpose in the old state? However, you have not update the one in `testdata/AD`, do you mind checking if this is relevant?

Finally, feel free to update the docker image on the repo and rerun the tests so that we can have the integration tests. You should have the correct permissions for it.

didrocks (approved):
So much green :)

internal/policies/apparmor/apparmor.go
```diff
@@ -1,4 +1,20 @@
...
+//
+// Next, if we found no entries to apply (either to them not existing or being
+// disabled), we attempt to unload all rules managed by ADSys.
+//
+// If there are entries to apply, based on the object type (machine or user), we
+// attempt to apply them. This process is more clearly outlined in the
+// ApplyPolicy function documentation.
+//
+// If any errors occur during the policy apply process, the manager will attempt
+// to restore the initial state of the system before returning an error.
```
denisonbarbosa, line 17: This shouldn't be here... When I've said that you could push to the branch of the already opened PR, I was referring to #631 (My bad, should've been more clear).

denisonbarbosa (commented):
Looking great! There's a little mistake in the diff, but feel free to merge after that is fixed.

internal/policies/proxy/proxy_test.go
```diff
@@ -0,0 +1,136 @@
...
+				"socks://example.com:8080",
+				"localhost,127.0.0.1",
+				"http://example.com:8080/proxy.pac",
+			},
+		},
+
+		// User cases
+		"User, no entries":        {isUser: true},
+		"User, non-empty entries": {isUser: true, entries: []entry.Entry{{Key: "not-applied", Value: "not-applied"}}},
+
```
didrocks, line 64: I don’t see a test case with no error (we won’t test the warning), when getting errDBusServiceUnknownName. Is that line covered? (I only see the one in the integration tests)

didrocks (changes requested):
I see the fun you had with Errors.As :) I think there is one package test case missing apart if I missed anytyhing.

internal/policies/proxy/proxy_test.go
didrocks: you’re right, force pushing makes reviewing the diff of diff harder :) I was expecting it to be in that table testing. Fine by me then, let’s approve. Thanks!

didrocks (approved):

PR #642, opened 2023-03-17
Add proxy manager documentation

GabrielNagy:
Fixes DEENG-558

### DIFF
--- a/doc/06.-Privileges-management.md
+++ b/doc/06.-Privileges-management.md
@@ -8,7 +8,7 @@ All those settings are globally enforced on the machine and are available at `Co
 
 ## Feature availability
 
-This feature is available only for the subscribers of the **Ubuntu Advantage** offer.
+This feature is available only for subscribers of **Ubuntu Pro**.
 
 ## Rules precedence
 
--- a/doc/07.-Scripts-execution.md
+++ b/doc/07.-Scripts-execution.md
@@ -13,7 +13,7 @@ Scripts can be shell scripts, or any binary that can be executed on Linux.
 
 ## Feature availability
 
-This feature is available only for the subscribers of the **Ubuntu Advantage** offer.
+This feature is available only for subscribers of **Ubuntu Pro**.
 
 ## Rules precedence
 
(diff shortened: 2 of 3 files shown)


doc/06.-Privileges-management.md
```diff
@@ -8,7 +8,7 @@ All those settings are globally enforced on the machine and are available at `Co
 
 ## Feature availability
 
-This feature is available only for the subscribers of the **Ubuntu Advantage** offer.
+This feature is available only for subscribers of **Ubuntu Pro**.
```
denisonbarbosa, line 11: Thanks for fixing those!

doc/10.-Proxy-settings.md
```diff
@@ -0,0 +1,46 @@
+# Proxy settings
+
+The proxy manager allows AD administrators to enforce proxy settings on the clients. Currently, only system-wide proxy settings are supported.
```
denisonbarbosa, line 3: Maybe we should use `apply | configure` rather than `enforce`? We don't do any checks after applying the proxy, IIRC, so technically we don't really enforce it.

doc/10.-Proxy-settings.md
```diff
@@ -0,0 +1,46 @@
...
+
+Proxy settings are configurable under the following GPO path:
+
+* System-wide level, located in `Computer Configuration > Policies > Administrative Templates > Ubuntu > Client management > System proxy configuration`
```
denisonbarbosa, line 7: An image here showing the navigation bar on the GPO configuration window would be nice.

Maybe we can use the current image for it and then add another one with the actual configuration window opened and some example content?

denisonbarbosa (commented):
Looks great, well done! Some suggestions, but nothing major.

cmd/generate_completion_documentation.go
```diff
@@ -65,7 +65,7 @@ func main() {
 		if generators.InstallOnlyMode() {
 			return
 		}
-		updateFromCmd(commands, filepath.Join("doc", "14.-Command-line-reference.md"))
+		updateFromCmd(commands, filepath.Join("doc", "15.-Command-line-reference.md"))
```
didrocks, line 68: I think a nice gardening activity will be to use globbing to match it :) but not related to that one)

didrocks (approved):
I have nothing to add compared to what Denison told

PR #647, opened 2023-03-23
Use systemd via D-Bus instead of shell commands

### LINKED ISSUE #425: Replace systemctl direct call in scripts with dbus call
This allows a more integrated experience by not having to rely on external command.

This will need:
- to be mocked in the package tests
- to be mocked in the integration tests
- remove the whole command override

GabrielNagy:
Leverage the go-systemd API via D-Bus instead of shelling out to systemctl.

Closes #425 / DEENG-483


denisonbarbosa (commented):
I liked it a lot. Very nice job!

internal/systemd/systemd.go
```diff
@@ -0,0 +1,99 @@
...
+// DefaultCaller is the default implementation of Caller.
+type DefaultCaller struct {
+	conn *systemdDbus.Conn
+}
+
+// jobDone is the string returned by systemd when a job completed successfully.
+const jobDone = "done"
+
+// NewWithConn returns a new systemdCaller using the given dbus connection.
+func NewWithConn(bus *dbus.Conn) (*DefaultCaller, error) {
```
didrocks, line 35: This is the only constructor of this systemd object, correct, maybe simply name it `New` then?

internal/systemd/systemd.go
```diff
@@ -0,0 +1,99 @@
...
+	conn, err := systemdDbus.NewConnection(func() (*dbus.Conn, error) { return bus, nil })
+	if err != nil {
+		return nil, err
+	}
+
+	return &DefaultCaller{conn: conn}, nil
+}
+
+// StartUnit starts the given unit.
+func (s *DefaultCaller) StartUnit(ctx context.Context, unit string) (err error) {
```
didrocks, line 45: for this one and all the others: there is no change on s itself, right? should that be a copy then and not a pointer?

internal/systemd/mock_test.go
```diff
@@ -0,0 +1,203 @@
...
+	jobPath := dbus.ObjectPath(fmt.Sprintf("%s/Job/%d", consts.SystemdDbusObjectPath, s.nextJobID))
+
+	// Emit JobNew signal
+	err := s.conn.Emit(
+		dbus.ObjectPath(consts.SystemdDbusObjectPath),
+		fmt.Sprintf("%s.JobNew", consts.SystemdDbusManagerInterface),
+		s.nextJobID, jobPath, name,
+	)
+	if err != nil {
+		log.Fatalf("failed to emit JobNew signal: %v", err)
```
didrocks, line 112: `log.Fatalf` will exit(1) right away, which will then not teardown properly the `go test` framework. You should rather `panic()` which is recovered in the `go test`.

internal/systemd/mock_test.go
```diff
@@ -0,0 +1,203 @@
...
+		s.nextJobID, jobPath, name, jobStatus,
+	)
+	if err != nil {
+		log.Fatalf("failed to emit JobRemoved signal: %v", err)
+	}
+
+	return jobPath
+}
+
+func TestMain(m *testing.M) {
```
didrocks, line 133: I would put this `TestMain` in `systemd_{est`, as this is where we have the tests.

internal/systemd/mock_test.go
```diff
@@ -0,0 +1,203 @@
...
+	// export systemd structure
+	defer testutils.StartLocalSystemBus()()
+
+	debug := flag.Bool("verbose", false, "Print debug log level information within the test")
+	flag.Parse()
+
+	var connOpts []dbus.ConnOption
+	if *debug {
```
didrocks, line 141: We have a card for this (we should handle debug logs consistently), but let’s get that in.

internal/systemd/systemd.go
```diff
@@ -0,0 +1,99 @@
...
+	defer decorate.OnError(&err, i18n.G("failed to start unit %s"), unit)
+
+	reschan := make(chan string)
```
didrocks, line 48: you never close the channel. Same on others.

internal/systemd/systemd_test.go
```diff
@@ -0,0 +1,95 @@
...
+	systemdCaller, err := systemd.NewWithConn(bus)
+	require.NoError(t, err, "Setup: failed to create systemd caller")
+
+	err = systemdCaller.StartUnit(ctx, "service-1.service")
+	require.NoError(t, err, "StartUnit shouldn't have failed but it did")
+
+	err = systemdCaller.StartUnit(ctx, "not-a-service.service")
+	require.Error(t, err, "StartUnit should have failed for nonexistent service")
+
+	err = systemdCaller.StartUnit(ctx, "fail-to-start-stop.service")
```
didrocks, line 28: small like a good candidate for table testing to test all of those use cases separately :) (same on other tests)

didrocks (changes requested):
Very nice work! I guess you are getting very familiar with dbus and systemd too :)

I have some suggestions to enhance the code a little bit further, but I’m very pleased with it!

internal/systemd/systemd.go
```diff
@@ -0,0 +1,99 @@
...
+	"errors"
+
+	systemdDbus "github.com/coreos/go-systemd/v22/dbus"
+	"github.com/godbus/dbus/v5"
+	"github.com/ubuntu/adsys/internal/i18n"
+	"github.com/ubuntu/decorate"
+)
+
+// Caller is the interface to interact with systemd.
+type Caller interface {
```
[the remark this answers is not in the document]
didrocks: exactly!

didrocks (changes requested):
I think there are some leftovers to do. Please push it in followup commits so that it’s easier to track the changes :)

internal/systemd/systemd_test.go
```diff
@@ -0,0 +1,170 @@
...
+
+	bus := testutils.NewDbusConn(t)
+
+	systemdCaller, err := systemd.New(bus)
+	require.NoError(t, err, "Setup: failed to create systemd caller")
+
+	err = systemdCaller.DaemonReload(ctx)
+	require.NoError(t, err, "DaemonReload shouldn't have failed but it did")
+
+	s.setErrorOnReload()
```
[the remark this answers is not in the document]
didrocks: That’s true, BUT, if here, you are mixing the 2 tests in a potential single error (in addition to make it less extensible). Do you think this will be too much boilerplate to do like the other one?

internal/policies/mount/export_test.go
```diff
@@ -11,9 +13,9 @@ func WithUserLookup(f func(string) (*user.User, error)) Option {
...
 
-// SetSystemCtlCmd allows to override the systemCtlCmd of the Manager for the tests.
-// This is used instead of a option function because we need to control the systemctl command
-// in multiple occasions during tests.
-func (m *Manager) SetSystemCtlCmd(args []string) {
-	m.systemCtlCmd = append(args, "systemctl")
+// SetSystemdCaller allows to override the systemdCaller of the Manager for the tests.
+// This is used instead of a option function because we need to control the
+// behavior of the mock in multiple occasions during tests.
+func (m *Manager) SetSystemdCaller(systemdCaller systemd.Caller) {
```
didrocks, line 19: It should use the mount interface systemdCaller

internal/systemd/systemd_test.go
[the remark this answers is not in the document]
didrocks: I have a strong feeling, but I can’t really argue enough with the assert (apart that you need to look at the test if the first fail alone or not :p), so let’s go with this (even if meh :p).

didrocks (approved):
That’s a huge enhancement overall, thanks for tackling this!

Issue #509, opened 2022-11-21
Investigate changes in coverage when no related files are modified


GabrielNagy:
From time to time we get reports of coverage changing on PRs when there are no changes in the affected files (e.g. dependency bumps, documentation updates). Both Codecov (which we use) and Coveralls don't seem to have a way to inspect this coverage diff online, so our best bet would be to download the raw coverage output from Codecov and compare it with local test runs to see the difference.

Example PRs:
- https://github.com/ubuntu/adsys/pull/506
- https://github.com/ubuntu/adsys/pull/503

If the changes prove to be too complicated we should open separate issues for them once we figure out the problems.

PR #651, opened 2023-03-24
Changing covermode to set in qa.yaml

denisonbarbosa:
We were using `covermode=count` in `adsys`, which isn't what we really want. This PR changes the `covermode` to `set`, which basically translates to `true` or `false` if the line is covered or not. Could help with #509.

DEENG-639


didrocks (approved):
Let’s see how it goes.

PR #658, opened 2023-03-29
UDENG-133 Fixing some i18n problems

denisonbarbosa:
We had some issues related to the i18n package. This PR aims to solve them.

### DIFF
--- a/cmd/adsysd/main.go
+++ b/cmd/adsysd/main.go
@@ -22,6 +22,7 @@ import (
 
 func main() {
 	var a app
+	i18n.InitI18nDomain(consts.TEXTDOMAIN)
 	switch filepath.Base(os.Args[0]) {
 	case daemon.CmdName:
 		a = daemon.New()
@@ -39,7 +40,6 @@ type app interface {
 }
 
 func run(a app) int {
-	i18n.InitI18nDomain(consts.TEXTDOMAIN)
 	defer installSignalHandler(a)()
 
 	log.SetFormatter(&log.TextFormatter{
--- a/cmd/adwatchd/commands/run.go
+++ b/cmd/adwatchd/commands/run.go
@@ -37,7 +37,7 @@ If a GPT.ini file does not exist for a directory, a warning will be issued and t
 					if !a.config.Force {
 						return fmt.Errorf(i18n.G("%s, use --force to override"), msg)
 					}
-					log.Warningf(context.Background(), i18n.G(msg))
+					log.Warningf(context.Background(), msg)
 				}
 			}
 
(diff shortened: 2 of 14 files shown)


cmd/adwatchd/commands/run.go
```diff
@@ -18,12 +18,8 @@ func (a *App) installRun() {
 	cmd := &cobra.Command{
 		Use:   "run",
 		Short: i18n.G("Starts the directory watch loop"),
-		Long: i18n.G(`Can run as a service through the service manager or interactively as a standalone application.
-
-The program will monitor the configured directories for changes and bump the appropriate GPT.ini versions anytime a change is detected.
-If a GPT.ini file does not exist for a directory, a warning will be issued and the file will be created. If the GPT.ini file is incompatible or malformed, the program will report an error.
-`),
-		Args: cobra.NoArgs,
+		Long:  i18n.G("Can run as a service through the service manager or interactively as a standalone application.\n\nThe program will monitor the configured directories for changes and bump the appropriate GPT.ini versions anytime a change is detected.\nIf a GPT.ini file does not exist for a directory, a warning will be issued and the file will be created. If the GPT.ini file is incompatible or malformed, the program will report an error.\n"),
```
CarlosNihelton, line 21:
```suggestion
Long:  i18n.G("Can run as a service through the service manager or interactively as a standalone application.\n\nThe program will monitor the configured directories for changes and bump the appropriate GPT.ini versions anytime a change is detected.\nIf a GPT.ini file does not exist for a directory, a warning will be issued and the file will be created.\nIf the GPT.ini file is incompatible or malformed, the program will report an error.\n"),
```

The `If the GPT.init...` sentence lives in its own line.

CarlosNihelton: Viewing the diff in the "files" tab made me think that there is:
image

cmd/adsysd/main.go
```diff
@@ -22,6 +22,7 @@ import (
 
 func main() {
 	var a app
+	i18n.InitI18nDomain(consts.TEXTDOMAIN)
```
didrocks, line 25: I think the a declaration is for the daemon below. So I would just move that one up. It’s not C89 :)

didrocks (changes requested):
One requested change, and I think we should, before merging this, consider if we can’t search for an alternative on string extractions.

EduardGomezEscandell:
> can we consider/research any other approach to extract the sentences in po files and to be compatible with bacquote?

Regarding extracting the strings: It is easy, I did it as a side thing here https://github.com/EduardGomezEscandell/go-gettext-adapter. I used this to generate PR https://github.com/ubuntu/adsys/pull/659.

The issue is maintaining this code (or a cleaner, better, not-done-in-one-afternoon version of it) and wether it is worth the hassle.

The linked repo already scans and finds the strings, it'd be a question of putting them in our potfile instead of sanitizing them.

didrocks:
And with the new i18n, I think we can drop this one!

PR #672, opened 2023-04-05
Print logs when downloadables are up to date

### LINKED ISSUE #454: Print logs when GPO is already up to date
We are lacking logs for telling when we don’t update a GPO or assets.

GabrielNagy:
We print logs when downloading GPOs/assets but not when they are already up to date. Handle this, and also add a debug print that lists the local and remote versions of the GPT.INI files.

Fixes #454 / DEENG-504


didrocks (approved):
Those small things making the admin life easier, thanks also for showing up the diff between local/remote version :)

PR #684, opened 2023-04-24
Go user mount handler

denisonbarbosa:
Using Rust for the user mount handler was affecting the maintainability of the project, since we had to support another language ecosystem for dependencies, packaging, etc.

Since the reason beyond the usage of Rust was to test its behavior and integration with Go to see if we could use it in `aad-auth` (and this objective is already fulfilled), we decided to rewrite the user mount handler in Go and turn it into a hidden command of `adsysd` rather than a separate binary to be executed.

UDENG-212


internal/policies/mount/handler.go
```diff
@@ -0,0 +1,192 @@
...
+//nolint:revive // We can't use _ for multiple unused parameters in exported functions.
+func askPassword(op *C.GMountOperation, unused1, unused2, unused3 *C.char, flags C.GAskPasswordFlags) {
+	rCode := C.GMountOperationResult(C.G_MOUNT_OPERATION_ABORTED)
+
+	// Checks if the mount operation is anonymous and if the anonymous flag is supported.
+	if C.g_mount_operation_get_anonymous(op) == C.TRUE {
+		if flags&C.G_ASK_PASSWORD_ANONYMOUS_SUPPORTED == C.G_ASK_PASSWORD_ANONYMOUS_SUPPORTED {
+			rCode = C.GMountOperationResult(C.G_MOUNT_OPERATION_HANDLED)
+		}
+	} else {
```
didrocks, line 168: let’s turns that as if looked like a switch with `else if os.Getenv…` ?

internal/policies/mount/handler.go
```diff
@@ -0,0 +1,192 @@
...
+// mountDone is the callback function that will be called when the mount operation is done.
+//
+//export mountDone
+func mountDone(sourceObject *C.GObject, res *C.GAsyncResult, _ C.gpointer) {
+	f := C.to_g_file(sourceObject)
+
+	var err *C.GError
+	C.g_file_mount_enclosing_volume_finish(f, res, &err)
+
+	doneMsg := msg{path: C.GoString(C.g_file_get_uri(f))}
```
didrocks, line 186: reading https://docs.gtk.org/gio/method.File.get_uri.html: you need to free the returned value from `g_file_get_uri` or this will be a memory leak.

didrocks (changes requested):
Nearly there! Very few comments, but those should be quite easy to fix :)

internal/policies/mount/handler.go
```diff
@@ -0,0 +1,192 @@
...
+	mainLoop := C.g_main_loop_new(C.g_main_context_default(), C.FALSE)
+
+	doneMain := make(chan struct{})
+	go func() {
+		C.g_main_loop_run(mainLoop)
+		close(doneMain)
+	}()
+
+	// watches the mountsChan channel for the results of the mount operations.
+	for i := 0; i < len(entries); {
```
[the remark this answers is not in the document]
denisonbarbosa: Sure. I thought this way would be clearer that we don't care for the slice content, only for the size, but syntactically it's just worse to read, as you pointed out.

internal/policies/mount/handler.go
denisonbarbosa: Damn... Tried so hard to avoid those and still have some. I'll fix it.

didrocks (approved):
Excellent, let’s merge!

PR #691, opened 2023-05-08
Handle case mismatches in GPT.INI file name

### LINKED ISSUE #667: Cannot determine local downloadable version for some GPT.INI files
## Description

When adsys checks whether the local/remote versions of downloadables, it always uses the uppercase form of `GPT.INI`. In most cases this file is automatically generated by Windows using the uppercase file name, however for `assets` directories, users can create the file by themselves. If they opt for a different spelling like `GPT.ini` or `gpt.ini`, subsequent downloads will not be able to determine the cached version and default to 0, causing the assets to always be downloaded.

https://github.com/ubuntu/adsys/blob/bebbb4a183d634699c83fb75ba8dccfa936c1dc3/internal/ad/download.go#L169-L170

## Reproduction

Create an assets directory under sysvol, containing a file named `GPT.ini` (lowercase extension) with the following contents:

```ini
[General]
Version=1
displayName=Ubuntu Assets Directory
```

Updating policies should then always redownload the asset files. This can be confirmed by adding/removing files from the assets directory - without incrementing the GPT.ini version - and executing `adsysctl update` on the client.

GabrielNagy:
Windows filesystems are case-insensitive, but things can get hairy when we parse the assets in Linux land, which for the most part is case-sensitive.

GPT files are always uppercase (both filename and extension) when created by Windows, but we cannot fully rely on this as we instruct users to create their own GPT.INI files for the assets directories. As such, upper/lowercase combinations might crop up, such as `GPT.ini`, `gpt.ini`, etc.

Up until now we've always considered `GPT.INI` to be the "canonical" path, but we can be a bit more defensive by walking the local directory and doing a case-insensitive comparison on each entry, stopping when we find a matching GPT.INI file.

This is not such a visible bug because our default behavior is to redownload the assets if we encounter any issues when parsing the local GPT.INI file, so we have the side effect of always redownloading assets even if there are no changes.

Fixes #667 / UDENG-227

### DIFF
--- a/internal/ad/download.go
+++ b/internal/ad/download.go
@@ -40,6 +40,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"strings"
 	"sync"
 
 	"github.com/mvo5/libsmbclient-go"
@@ -172,12 +173,13 @@ func needsDownload(ctx context.Context, client *libsmbclient.Client, g *download
 	defer g.mu.RUnlock()
 
 	var localVersion, remoteVersion int
-	gptIniPath := filepath.Join(localPath, "GPT.INI")
-	if f, err := os.Open(filepath.Clean(gptIniPath)); err == nil {
-		defer decorate.LogFuncOnErrorContext(ctx, f.Close)
+	if gptIniPath, err := findLocalGPTIni(localPath); err == nil {
+		if f, err := os.Open(filepath.Clean(gptIniPath)); err == nil {
+			defer decorate.LogFuncOnErrorContext(ctx, f.Close)
 
-		if localVersion, err = getGPOVersion(ctx, f, g.name); err != nil {
-			log.Warningf(ctx, "Invalid local GPT.INI for %s: %v\nDownloading it again…", g.name, err)
+			if localVersion, err = getGPOVersion(ctx, f, g.name); err != nil {
+				log.Warningf(ctx, "Invalid local GPT.INI for %s: %v\nDownloading it again…", g.name, err)
+			}
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


internal/ad/download.go
```diff
@@ -330,3 +332,21 @@ func downloadRecursive(ctx context.Context, client *libsmbclient.Client, url, de
...
+// To account for case differences in the filename/extension, we have to walk
+// the directory and check each entry.
+func findLocalGPTIni(path string) (string, error) {
+	entries, err := os.ReadDir(path)
+	if err != nil {
+		return "", fmt.Errorf(i18n.G("could not read directory %q: %w"), path, err)
+	}
+
+	for _, entry := range entries {
+		if strings.EqualFold(entry.Name(), "GPT.INI") {
```
didrocks, line 346: Let’s imagine that you are using a samba machine to distribute the GPO (so Linux), you have `GPT.INI` and `gpt.ini`, which one is going to win then? Shouldn’t we first always try the "canonical" one before fallbacking to case insensitive research?
Does the samba shared mount even supports this?

internal/ad/testdata/AD/SYSVOL/assetsonly.com/Distrolowercasegpt/GPT.ini
```diff
@@ -0,0 +1,3 @@
+[General]
```
didrocks, line 1: nickpitt on the folder name, but it’s actually the extension which is lowercase, not the whole name.

didrocks (commented):
One feedback requested to know how we try to privilege the "should work" path and a nitpick on the naming.

Let’s resolve those 2 before going on.

internal/ad/testdata/AD/SYSVOL/assetsonly.com/Distrolowercasegpt/GPT.ini
didrocks: :p

didrocks (approved):
Perfect!

PR #696, opened 2023-05-10
Ensure empty state for dconf policy manager

### LINKED ISSUE #678: Ensure the dconf empty state leaves no leftover files
When applying dconf policies (specifically user policies) that are empty, we still get leftover files under `profile/<username>`. We should clean those up as well.

Additionally, confirm that all managers have tests for the empty state.

GabrielNagy:
To aid in the usage of the purge command, unify the empty state of the dconf manager similar to our other managers.

For user policies, we can safely remove the DB/profile if we have no entries to apply. For machine policies we cannot do this, as we do not know if there are other user policies that might depend on it. Instead, we handle purging the machine DBs/profiles in our prerm purge script.

This required a few changes in tests that expected profile data to still exist when applying a GPO with 0 entries.

Fixes #678 / UDENG-221


internal/policies/dconf/dconf.go
```diff
@@ -95,12 +95,27 @@ func (m *Manager) ApplyPolicy(ctx context.Context, objectName string, isComputer
 	dbsPath := filepath.Join(dconfDir, "db")
 	dbPath := filepath.Join(dbsPath, objectName+".d")
 
-	if !isComputer {
+	if !isComputer && len(entries) > 0 {
```
didrocks, line 98: Should we have a test with user len(entries) == 0 and no computer data to ensure we don’t fail?

Or is it `"User empty state": {entries: []entry.Entry{}, existingDconfDir: "-"},` (which could be renamed "User empty state without machine policy")?

didrocks (commented):
Excellent work! Just one pending question before finale approval :)

didrocks (approved):
Thanks for the additional test case! Looks good to me :)

PR #699, opened 2023-05-16
Rework krb5cc handling logic

GabrielNagy:
Following the Heimdal issues surfaced after the samba 4.16.0 release we have to rethink the krb5cc handling logic with the following in mind:
- we can no longer use symlinks as Kerberos tickets
- the ticket has to be owned by the running user (in our case root)

This is not related to a change in samba itself, but to the Kerberos implementation vendored in samba (Heimdal Kerberos). Between samba 4.15.0 and 4.16.0 there has been a massive update in the vendored code, pulling in around [10 years worth of changes](https://github.com/samba-team/samba/commit/40b65c840e03bd5eb7f3b02fe80144650c63c005). Among these changes we were able to find information confirming the 2 issues above, and we concluded they are not considered to be bugs in the [upstream implementation](https://github.com/heimdal/heimdal/commit/2a565482f491621fe8a906c990aa890be153a906) (and related [issue](https://github.com/heimdal/heimdal/issues/306)).

Another important thing to note is that the issues above do not manifest if samba is compiled and linked against the MIT Kerberos library. However, MIT Kerberos appears to still be considered experimental (as of today), and neither Ubuntu or other distros are compiling samba with MIT Kerberos.

As such, we have to work around these "issues" within adsys. Instead of tracking the ticket's lifetime using a symlink, we rely on the following approach:

* still track original ticket with a symlink
* copy the ticket as a root-owned ticket
* track refresh with timestamps on the symlink itself.

Symlinks will remain stored under KRB5CACHEDIR, whereas the actual copies will reside under KRB5CACHEDIR/tickets.

Closes UDENG-289

### DIFF
--- a/internal/ad/ad.go
+++ b/internal/ad/ad.go
@@ -424,7 +424,7 @@ func (ad *AD) ensureKrb5CCCopy(krb5CCSymlink, krb5CCCopyName string) error {
 		return fmt.Errorf(i18n.G("failed to read krb5cc symlink: %w"), err)
 	}
 
-	if copyStat, err := os.Stat(krb5CCCopyName); err == nil {
+	if copyStat, err := os.Lstat(krb5CCCopyName); err == nil && copyStat.Mode()&os.ModeSymlink == 0 {
 		// We already have a copy of the ticket, let's check if we need to update it
 		srcStat, err := os.Stat(krb5CCSrc)
 		if err != nil {
--- a/internal/ad/ad_test.go
+++ b/internal/ad/ad_test.go
@@ -746,8 +746,9 @@ func TestGetPoliciesWorkflows(t *testing.T) {
 		userKrb5CCBaseName1 string
 		userKrb5CCBaseName2 string
 
-		restart      bool
-		modifyKrb5CC bool
+		restart       bool
+		modifyKrb5CC  bool
+		symlinkKrb5CC bool
 
 		wantErr bool
 	}{
@@ -782,6 +783,13 @@ func TestGetPoliciesWorkflows(t *testing.T) {
 			userKrb5CCBaseName1: "bob",
 			userKrb5CCBaseName2: "EMPTY",
 			modifyKrb5CC:        true,
+		},
+		"Second call without Krb5CCName replaces previous symlinks in the original path": {
+			objectName1:         "bob@ASSETSANDGPO.COM",
+			objectName2:         "bob@ASSETSANDGPO.COM",
+			userKrb5CCBaseName1: "bob",
+			userKrb5CCBaseName2: "EMPTY",
+			symlinkKrb5CC:       true,
 		},
 
 		// Machine for assets cases
@@ -850,6 +858,13 @@ func TestGetPoliciesWorkflows(t *testing.T) {
 				require.NoError(t, err, "Setup: cannot modify krb5cc file")
 			}
 
+			if tc.symlinkKrb5CC {
+				err = os.Remove(filepath.Join(rundir, "krb5cc", tc.objectName1))
+				require.NoError(t, err, "Setup: cannot remove krb5cc copy")
+				err = os.Symlink(krb5CCName, filepath.Join(rundir, "krb5cc", tc.objectName1))
+				require.NoError(t, err, "Setup: cannot create symlink")
+			}
+
 			// Recreate the ticket if needed or reset it to empty for refresh
 			if tc.userKrb5CCBaseName2 != "" {
 				if tc.userKrb5CCBaseName2 == "EMPTY" {
@@ -876,6 +891,10 @@ func TestGetPoliciesWorkflows(t *testing.T) {
 				require.NoError(t, err, "Setup: can't read krb5cc file contents")
 				assert.Equal(t, "KRB5 Ticket modified content", string(krb5ccContents), "changes to krb5cc file should be reflected in the copy")
 			}
+
+			stat, err := os.Lstat(filepath.Join(rundir, "krb5cc", tc.objectName2))
+			require.NoError(t, err, "Setup: can't stat krb5cc file")
+			require.True(t, stat.Mode().IsRegular(), "krb5cc file should be a regular file")
 
 			wantPolicyDir = filepath.Join("testdata", "sysvolcache", strings.ToLower(tc.objectName2))
 			if tc.objectName2 == hostname {


internal/ad/ad.go
```diff
@@ -357,15 +360,32 @@ func (ad *AD) ListUsers(ctx context.Context, active bool) (users []string, err e
...
-
+// ensureKrb5CCName manages user ccname ticket lifetimes.
+// It handles concurrent calls, and works by first creating a symlink to the
+// actual ticket for tracking purposes.
+// It then makes a copy of the ticket, changing ownership to the calling user
+// (root) in order to satisfy the Heimdal implementation of Kerberos.
+// To track modifications to the krb5cc file we look at the mtime of the
+// original ticket and keep it up to date with the copy.
+func (ad *AD) ensureKrb5CCName(srcKrb5CCName, krb5CCCopyName, objectName string) error {
 	ad.Lock()
```
denisonbarbosa, line 371: Maybe some more error decorations here would help as well

internal/ad/ad.go
```diff
@@ -389,6 +409,57 @@ func (ad *AD) ensureKrb5CCName(srcKrb5CCName, dstKrb5CCName string) (err error)
 	return nil
 }
 
+// ensureKrb5CCCopy checks if the cached krb5cc ticket is up to date and updates
+// it if not.
+func ensureKrb5CCCopy(srcKrb5CCName, krb5CCCopyName string) (err error) {
```
denisonbarbosa, line 414: Same thing on the error decoration here...

Maybe it's better to stop decorating on the "helper" functions and decorate only the error on `ensureKrb5CCName` to avoid too much decorate chains?

internal/ad/ad.go
```diff
@@ -389,6 +409,57 @@ func (ad *AD) ensureKrb5CCName(srcKrb5CCName, dstKrb5CCName string) (err error)
...
+		return fmt.Errorf(i18n.G("failed to open destination file: %w"), err)
+	}
+	defer dstFile.Close()
+
+	if _, err := io.Copy(dstFile, srcFile); err != nil {
+		return fmt.Errorf(i18n.G("failed to copy file: %w"), err)
+	}
+
+	return nil
+}
```
denisonbarbosa, line 461: Instead of having two file descriptors opened and potentially opening ourselves for tragedy, can't we read from `src` and then safely write (i.e. writing as .new and then renaming it) to `dest`?

denisonbarbosa (changes requested):
Looks very nice, well done! I have some suggestions that I feel like should be attended, but otherwise: amazing work!

internal/ad/ad.go
denisonbarbosa: I was thinking more in the way of reading the file first with `os.ReadFile` and then writing it and renaming with `os.WriteFile` and `os.Rename`. This way we can avoid the `io` package alltogether.

denisonbarbosa (approved):
Looks great! Well done for spotting and fixing it!

cmd/adsysd/integration_tests/adsysctl_policy_test.go
```diff
@@ -451,7 +451,7 @@ func TestPolicyUpdate(t *testing.T) {
 			initState: "old-data",
 			krb5ccNamesState: []krb5ccNamesWithState{
 				{
-					src:          "ccache_EXAMPLE.COM",
+					src:          "ccache_OFFLINE",
```
didrocks, line 454: waow, nicely spotted! (or thanks for the test to uncover a future bug :p). I was puzzled for a short while, until I opened the corresponding sssd.conf and confirm with your commit comments :)

didrocks (changes requested):
Nice work! It’s always almost a deception when you see how much the investigation took compared to the "fix" (or rather workaround here). Well done!

I have one small request to make the whole logic more understandable in the future (I think, as an admin, I open the cache and see directly what is in use).

However, I wonder about the coverage. Codecov is not really helpful as it doesn’t show up the file diff change. Do you think that some tests in ad_test.go could be added to trigger some uncovered tests and not only relying on the integration tests to cover it?

(I’m thinking in particular about a test with ticket, ad, get it, then updating the ticket, ad away, -> copy was refreshed)

didrocks:
> @didrocks sorry for the aggresive force-pushing blush the only changes are in the first and last commit (which is newly added).

I was on the starting block to complain, but now that you apologized… :p

cmd/adsysd/integration_tests/adsysctl_service_test.go
```diff
@@ -333,7 +333,7 @@ func TestServiceStatus(t *testing.T) {
 
 			// Create users krb5cc and GPO caches
 			if !tc.noCacheUsersMachine {
-				krb5UserDir := filepath.Join(adsysDir, "run", "krb5cc")
+				krb5UserDir := filepath.Join(adsysDir, "run", "krb5cc", "tracking")
```
didrocks, line 336: should we only create that one and not the ticket copy too to simulate a real run state?

internal/ad/ad_test.go
```diff
@@ -1098,7 +1109,7 @@ func TestListUsers(t *testing.T) {
 			cachedir, rundir := t.TempDir(), t.TempDir()
 
 			// populate rundir with users and machines
-			krb5CacheDir := filepath.Join(rundir, "krb5cc")
+			krb5CacheDir := filepath.Join(rundir, "krb5cc", "tracking")
```
didrocks, line 1112: Ditto here, shouldn’t we create a real system state with the copy too?

internal/ad/ad.go
```diff
@@ -340,7 +341,7 @@ func (ad *AD) ListUsers(ctx context.Context, active bool) (users []string, err e
 
 	cacheDir := ad.policiesCacheDir
 	if active {
-		cacheDir = ad.krb5CacheDir
+		cacheDir = filepath.Join(ad.krb5CacheDir, "tracking")
```
didrocks, line 344: So, here, we are not checking the symlinks are still valid, there is thus no advantaged compared to only readDir the cacheDir (and ignoring subdirectories).

So, I suggest we either:
* rely on the symlink, but in that case, it would be good to see if the destination still exists (and ignoring those that don't, with a test for it)
* keep the current behaviour, but then basing on `krb5CacheDir`

didrocks (changes requested):
I’m not sure I’m fully following the logic here (in addition to the comment I mentioned).

I don’t see the symlink to be reused anywhere after being created. The idea is that any "all policy updates" (so, without explicit env variable to get the ticket), should get the real ticket location from the symlink and assess if a copy refresh is in order? I may be missing it completely, but I even reopened the file to see if there is any hidden logic I’m missing.

If I am correct, it means that we are missing a test to setup something for a given user, then, modifying ticket and ask for a refresh without pointing to that user in particular, and check that the copy has been refreshed.

didrocks:
> That's a very good observation which I totally missed, thanks for that. I've now updated the implementation to attempt the copy even if an explicit krb5ccname is not passed and validated the behavior with a package test.

No worry, I think that shows we should have an automated test for this behavior in the `ad` package, wdyt?

cmd/adsysd/integration_tests/adsysctl_policy_test.go
```diff
@@ -344,6 +344,26 @@ func TestPolicyUpdate(t *testing.T) {
 					machine:      true,
 				},
 			}},
+		"Refresh with one dangling symlink does not fail": {args: []string{"--all"},
```
didrocks, line 347: `s/does not fail/ignores the respective user/` ?

didrocks (approved):
Excellent, one last comment on the test name, then feel free to merge :)

Woow, good to see this merged, especially with, in the end, some improvements with the symlink checked :) Well done!

PR #703, opened 2023-05-18
Updating pam module to properly set DCONF_PROFILE

### LINKED ISSUE #686: dconf profile
When using "default_domain_suffix" in sssd.conf the DCONF_PROFILE environment variable gets set to USERNAME instead of USERNAME@DOMAIN When the created dconf profiles and db are USERNAME@DOMAIN . Making it not possible for domain users to use applications like gedit. I tested using USERNAME@DOMAIN to login and it set the correct DCONF_PROFILE and was able to use gedit.

denisonbarbosa:
Our pam module wasn't parsing sssd.conf to check for the default_domain_suffix configuration value when setting the DCONF_PROFILE variable. This means that, for cases where the user authenticates without an explicit domain, the DCONF_PROFILE would be set to the username only, instead of user@domain.

By parsing the config value, now we are able to set the environment properly regardless of the username configuration.

Closes #686 / UDENG-331

### DIFF
--- a/pam/pam_adsys.c
+++ b/pam/pam_adsys.c
@@ -45,6 +45,7 @@
 #include <security/pam_modutil.h>
 
 #define ADSYS_POLICIES_DIR "/var/cache/adsys/policies/%s"
+#define SSSD_CONF_PATH "/etc/sssd/sssd.conf"
 
 /*
  * Refresh the group policies of current user
@@ -198,19 +199,96 @@ static int update_machine_policy(pam_handle_t *pamh, int debug) {
 }
 
 /*
+ * Get default domain suffix from SSSD_CONF_PATH
+ */
+static char *get_default_sss_domain(pam_handle_t *pamh) {
+    FILE *f = fopen(SSSD_CONF_PATH, "r");
+    if (f == NULL) {
+        pam_syslog(pamh, LOG_ERR, "Failed to open sssd.conf");
+        return NULL;
+    }
+
+    size_t buffsize = 256;
+    char *buf = malloc(sizeof(char) * buffsize);
+    char *domain = NULL;
+    while (getline(&buf, &buffsize, f) != -1) {
+        char *line = buf;
+        // ignores whitespaces listed before the config key
+        while (strlen(line) > 0 && (*line == ' ' || *line == '\t')) {
+            line++;
+        }
+        if (strncmp(line, "default_domain_suffix", 21) == 0) {
+            domain = strchr(line, '=');
+            if (domain == NULL) {
+                pam_syslog(pamh, LOG_ERR, "Could not find value for key 'default_domain_suffix' in sssd.conf");
+                break;
+            }
+            while (*++domain == ' ') continue;
+            char *newline = strchr(domain, '\n');
+            if (newline != NULL) {
+                *newline = '\0';
+            }
+            break;
+        }
+    }
+    fclose(f);
+
+    char *ret = strdup(domain);
+    free(buf);
+    return ret;
+}
+
+/*
+ * Converts domain\user to user@domain format
+ */
+static char *slash_to_at_username(const char *username) {
+    char *backslash = strchr(username, '\\');
+    if (backslash != NULL) {
+        char *ret = malloc((strlen(username) + 1) * sizeof(char));
+        strcpy(ret, backslash + 1);
+        strcat(ret, "@");
+        strncpy(ret + strlen(ret), username, backslash - username);
+        return ret;
+    }
+    return strdup(username);
+}
+
+/*
  * Set DCONF_PROFILE for current user
  */
 static int set_dconf_profile(pam_handle_t *pamh, const char *username, int debug) {
     int retval = PAM_SUCCESS;
 
+    char *profile_name = slash_to_at_username(username);
+
+    // We need to check if the profile name does not already contain the domain.
+    if (strchr(profile_name, '@') == NULL) {
+        char *domain = get_default_sss_domain(pamh);
+        if (domain != NULL) {
+            free(profile_name);
+            profile_name = (char *)malloc((strlen(username) + strlen(domain) + 2) * sizeof(char));
+            strcpy(profile_name, username);
+            strcat(profile_name, "@");
+            strcat(profile_name, domain);
+            free(domain);
+        }
+    }
+    // We need to lowercase the profile_name, as it can have uppercased letters and we
+    // always normalize it in adsys.
+    for (char *s = profile_name; *s; s++) {
+        *s = tolower(*s);
+    }
+
     char *envvar;
-    if (asprintf(&envvar, "DCONF_PROFILE=%s", username) < 0) {
+    if (asprintf(&envvar, "DCONF_PROFILE=%s", profile_name) < 0) {
         pam_syslog(pamh, LOG_CRIT, "out of memory");
+        free(profile_name);
         return PAM_BUF_ERR;
     }
 
     retval = pam_putenv(pamh, envvar);
     _pam_drop(envvar);
+    free(profile_name);
     return retval;
 }
 


pam/pam_adsys.c
```diff
@@ -223,14 +262,31 @@ static int set_dconf_profile(pam_handle_t * pamh, const char *username,
 {
 	int retval = PAM_SUCCESS;
 
+	char* profile_name = (char*)malloc((strlen(username) + 1) * sizeof(char));
+	strcpy(profile_name, username);
+
+	// We need to check if the profile name does not already contain the domain or is in the \\format
+	if (strchr(profile_name, '@') == NULL && strchr(profile_name, '\\') == NULL) {
+		char* domain = get_default_domain(pamh);
```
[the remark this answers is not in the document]
denisonbarbosa: Well, we come in after this goes through `gdm` and `pam_sss`, so I think we don't need to lowercase the name or consider it coming without TLD as its value comes from a configuration file and if it's unset properly, it shouldn't work and we shouldn't try to make it work (I guess).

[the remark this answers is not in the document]
denisonbarbosa: Oh, I see it now. Yeah, it makes sense... I'll update it to lowercase the domain that we get from `sssd.conf`. The `$USER` is a nice idea, but I'm not sure if it's already set when `pam_aad` is called and what are the full implications on using it.

pam/pam_adsys.c
```diff
@@ -215,6 +216,44 @@ static int update_machine_policy(pam_handle_t * pamh, int debug)
 	return PAM_SYSTEM_ERR;	/* will never be reached. */
 }
 
+/*
+ * Get default domain suffix from SSSD_CONF_PATH
+ */
+static char* get_default_domain(pam_handle_t* pamh) {
```
didrocks, line 222: `get_default_sss_domain()`? (remember that we might use winbind now!)

pam/pam_adsys.c
```diff
@@ -215,6 +216,44 @@ static int update_machine_policy(pam_handle_t * pamh, int debug)
...
+	if (f == NULL) {
+		pam_syslog(pamh, LOG_DEBUG, "Failed to open sssd.conf");
+		return "\0";
+	}
+
+	size_t buffsize = 256;
+	char* line = malloc(sizeof(char) * buffsize);
+	char* domain = "\0";
+	while (getline(&line, &buffsize, f) != -1) {
+		if (strncmp(line, "default_domain_suffix", 21) == 0) {
```
didrocks, line 233: Are you sure we can’t have a valid:

`default_domain_suffix = `

or even:

` default_domain_suffix=`

?

pam/pam_adsys.c
```diff
@@ -223,14 +262,36 @@ static int set_dconf_profile(pam_handle_t * pamh, const char *username,
 {
 	int retval = PAM_SUCCESS;
 
+	char* profile_name = (char*)malloc((strlen(username) + 1) * sizeof(char));
+	strcpy(profile_name, username);
+
+	// We need to check if the profile name does not already contain the domain or is in the \\format
+	if (strchr(profile_name, '@') == NULL && strchr(profile_name, '\\') == NULL) {
+		char* domain = get_default_domain(pamh);
```
didrocks, line 270: Also, I don’t see `domain` being freed?

pam/pam_adsys.c
```diff
@@ -223,14 +262,36 @@ static int set_dconf_profile(pam_handle_t * pamh, const char *username,
 {
 	int retval = PAM_SUCCESS;
 
+	char* profile_name = (char*)malloc((strlen(username) + 1) * sizeof(char));
+	strcpy(profile_name, username);
+
+	// We need to check if the profile name does not already contain the domain or is in the \\format
+	if (strchr(profile_name, '@') == NULL && strchr(profile_name, '\\') == NULL) {
```
didrocks, line 269: So if someone is using the syntax `domain\\user`, what is then the DCONF_PROFILE set to?

didrocks (changes requested):
Nice C journey :)
The style is correct and I didn’t spot anything obvious.

There are I guess some missing free (TBC), and some opened questions :)

pam/pam_adsys.c
denisonbarbosa: Once I `strcat` the `domain` into `profile_name`, when I free `profile_name` doesn't it free the `domain` as well?

pam/pam_adsys.c
denisonbarbosa: I don't know how pam handles this kind of username, but if it doesn't change it then it would be set to `\\domain\user`, so I guess the parsing should be a bit more robust

didrocks: I think you really should give it a try and see how it goes. If it doesn’t replace with @, then, we will have again the issue with DCONF_PROFILE for this use case.

pam/pam_adsys.c
didrocks:
```
       strcat()
              This  function catenates the string pointed to by src, after the
              string pointed to  by  dst  (overwriting  its  terminating  null
              byte).   The programmer is responsible for allocating a destina‐
              tion buffer large enough, that is, strlen(dst) +  strlen(src)  +
              1.
```.

So, this means that the new allocated has nothing to do to src and dst, and so, if we allocated any of them, we need to free them.

File: pam/pam_adsys.c (hunk 3 above)
Comment (didrocks) [reply pam/pam_adsys.c:233]: I’m unsure the first one will work, as the character index will be 22 and not 21.

Do you mind trying both in your setup?

Comment (denisonbarbosa) [reply pam/pam_adsys.c:233]: I was always using it as `default_domain_suffix = domain`

File: pam/pam_adsys.c (hunk 1 above)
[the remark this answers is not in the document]
Comment (denisonbarbosa) [reply pam/pam_adsys.c:270]: I'll change it to always lowercase the final string.

File: pam/pam_adsys.c [hunk 6]
```diff
@@ -219,27 +219,30 @@ static int update_machine_policy(pam_handle_t * pamh, int debug)
...
-				pam_syslog(pamh, LOG_DEBUG, "Could not find default_domain_suffix in sssd.conf");
+                pam_syslog(pamh, LOG_ERR,
+                    "Could not find value for key 'default_domain_suffix' in sssd.conf");
 				break;
 			}
-			domain++;
-			while (*domain == ' ') {
-				domain++;
-			}
+			while (*++domain == ' ') continue;
```
denisonbarbosa, line 245: This is the apex of C programming

didrocks (approved):
Excellent, thanks for the changes! I trust your local tests as unfortunately, this can only be done in the e2e tests.

pam/pam_adsys.c
```diff
@@ -215,6 +216,62 @@ static int update_machine_policy(pam_handle_t * pamh, int debug)
...
+	    char* line = buf;
+	    // ignores whitespaces listed before the config key
+	    while (strlen(line) > 0 && (*line == ' ' || *line == '\t')) {
+	        line++;
+        }
+		if (strncmp(line, "default_domain_suffix", 21) == 0) {
+			domain = strchr(line, '=');
+			if (domain == NULL) {
+                pam_syslog(pamh, LOG_ERR,
+                    "Could not find value for key 'default_domain_suffix' in sssd.conf");
```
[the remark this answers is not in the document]
denisonbarbosa: (same thing about the other comment)

pam/pam_adsys.c
```diff
@@ -215,6 +216,62 @@ static int update_machine_policy(pam_handle_t * pamh, int debug)
...
 
+/*
+ * Get default domain suffix from SSSD_CONF_PATH
+ */
+static char* get_default_sss_domain(pam_handle_t* pamh) {
+	FILE* f = fopen(SSSD_CONF_PATH, "r");
+	if (f == NULL) {
+        pam_syslog(pamh, LOG_ERR, "Failed to open sssd.conf");
+        return NULL;
+	}
```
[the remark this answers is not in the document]
denisonbarbosa: Oh, you're right. Thanks for spotting it!

PR #705, opened 2023-05-24
Replacing tab indentation in `pam_adsys.c`

denisonbarbosa:
In the `pam_adsys.c` file, we were using tabs for indentation. This replaces them with spaces to avoid future problems.


didrocks (approved):
And I was pretty sure my editor was configured to replace tabs with spaces… Probably because it’s opened in a Go workspace and it get confused. We should have a linter for this…

Thanks!

PR #707, opened 2023-05-24
Adding clang-format configuration and applying the linter to C source files

denisonbarbosa:
As discussed, we decided to add a linter to C source files in our projects.
This PR adds a .clang-format file to the root directory and applies the linter to our C files.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,7 @@
+# Configures clang-format for the project.
+# Options can be consulted on https://clang.llvm.org/docs/ClangFormatStyleOptions.html
+Language: Cpp # Clang format will only be applied to C/C++ files.
+BasedOnStyle: Google
+IndentWidth: 4
+ColumnLimit: 120
+UseTab: Never
--- a/.github/workflows/qa.yaml
+++ b/.github/workflows/qa.yaml
@@ -17,12 +17,11 @@ jobs:
   quality:
     name: Code quality and mod files up to date
     runs-on: ubuntu-latest
-    container: ubuntu:latest
     steps:
       - name: Install dependencies
         run: |
-          DEBIAN_FRONTEND=noninteractive apt update
-          DEBIAN_FRONTEND=noninteractive apt install -y ${{ env.apt_dependencies }}
+          sudo DEBIAN_FRONTEND=noninteractive apt update
+          sudo DEBIAN_FRONTEND=noninteractive apt install -y ${{ env.apt_dependencies }}
       - name: work around permission issue with git vulnerability (we are local here). TO REMOVE
         run: git config --global --add safe.directory /__w/adsys/adsys
       - uses: actions/checkout@v3
@@ -52,6 +51,12 @@ jobs:
         if: ${{ always() }}
       - name: Building
         run: go build ./...
+        if: ${{ always() }}
+      - name: C code formatting
+        uses: jidicula/clang-format-action@v4.11.0
+        with:
+          include-regex: '^.*\.(c|h)$' # Makes sure to run only on C (source and header) files
+          exclude-regex: 'vendor' # Excludes the vendor directory
         if: ${{ always() }}
 
   tests:
--- a/pam/pam_adsys.c
+++ b/pam/pam_adsys.c
@@ -31,26 +31,25 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
-#include <syslog.h>
 #include <sys/types.h>
 #include <sys/wait.h>
+#include <syslog.h>
 #include <unistd.h>
 
 #define PAM_SM_AUTH
 #define PAM_SM_SESSION
 
+#include <security/_pam_macros.h>
+#include <security/pam_ext.h>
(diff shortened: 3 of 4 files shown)


.clang-format
```diff
@@ -0,0 +1,7 @@
+# Configures clang-format for the project.
+# Options can be consulted on https://clang.llvm.org/docs/ClangFormatStyleOptions.html
+BasedOnStyle: Google
+IndentWidth: 4
+ColumnLimit: 120
+TabWidth: 4
+UseTab: Never
```
denisonbarbosa, line 7: These config values are up for debate. Let me know what you guys think about it.

.clang-format
```diff
@@ -0,0 +1,7 @@
+# Configures clang-format for the project.
+# Options can be consulted on https://clang.llvm.org/docs/ClangFormatStyleOptions.html
+BasedOnStyle: Google
+IndentWidth: 4
+ColumnLimit: 120
+TabWidth: 4
```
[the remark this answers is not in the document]
denisonbarbosa: I realized it as soon as I pushed the branch hahaha

denisonbarbosa:
> We should have a CI job checking this otherwise it would be easy to forget about clang-format. Not a high priority since we're not messing with C files a lot.

I agree. I was already looking into some existing `github-actions` that run `clang-format` on the repository

CarlosNihelton:
If you don't get satified with any existing action or feels they are overkill, you may just `apt install clang-format-16` and run it with `clang-format-16 -i` and then use `peter-evans/create-pull-request` action to push the diff.

denisonbarbosa:
Ok, so I think that running the job in a container is messing up the clang-format action. I'll create a PR to remove the Rust related stuff from adsys and then rebase this one into it.

didrocks (changes requested):
Github hickup: commented posted twice. Edited.

pam/pam_adsys.c
```diff
@@ -31,295 +31,261 @@
...
+        pam_syslog(pamh, LOG_ERR, "Failed to fork process");
+        return PAM_SYSTEM_ERR;
+    }
+
+    if (pid > 0) { /* parent */
+        pid_t retval;
+        int status = 0;
+
+        while ((retval = waitpid(pid, &status, 0)) == -1 && errno == EINTR)
+            ;
```
didrocks, line 90: That was my second point, I’m still worried about this one :)

denisonbarbosa: I'm not sure... This kind of formatting looks kind of familiar for these types of blocks without inside code. We can add a `continue` there to make sure the formatter behaves, wdyt? Might even make the code clearer

didrocks: as you wish, look for what is the easiest. It could just be buming to 160 characters :)

.github/workflows/qa.yaml
```diff
@@ -53,6 +52,11 @@ jobs:
       - name: Building
         run: go build ./...
         if: ${{ always() }}
+      - name: C code formatting
+        uses: jidicula/clang-format-action@v4.11.0
+        with:
+          exclude-regex: '(vendor|*.\.proto)'
```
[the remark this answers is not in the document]
didrocks: maybe ignore then (ino|pde|proto|cu) which is the default include regexp they are doing? But I would rather be explicit including than excluding, personnaly.

pam/pam_adsys.c
denisonbarbosa: Oh, no. The limit was fine. I think it uses the newline for the `;` just to make it clear that it is a `do nothing` line

.github/workflows/qa.yaml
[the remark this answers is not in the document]
denisonbarbosa: Yeah, not excluding it does not work. So we can just include .c files.

pam/pam_adsys.c
didrocks: soooooooooo ugly, can we try `while ((retval = waitpid(pid, &status, 0)) == -1 && errno == EINTR) {};` ?

didrocks: yes, because when I see "continue" I think of getting into the loop in a busy one, while actually, the while is blocking and never gets into `{}`

denisonbarbosa: This is what clang-format has to say about it:
```C
while ((retval = waitpid(pid, &status, 0)) == -1 && errno == EINTR) {
};
```

didrocks: hum, not that bad, wdyt?

denisonbarbosa: It still hurts, but it's definitely better than a random indented `;` in the middle of the code

didrocks: agreed

didrocks (approved):
Excellent! Once CI pass, feel free to merge it :)

pam/pam_adsys.c
denisonbarbosa: Honestly, I'm just glad I get to not look at
```C
int function()
{
}
```
anymore

PR #709, opened 2023-05-25
Removing Rust sources

denisonbarbosa:
As of #684, the user mounts are now handled by a hidden command in the daemon, written in Go. So we don't need Rust anymore. 
Goodbye cargo, you will be remembered...


didrocks (approved):
For once, Red is good :)

PR #711, opened 2023-05-26
Prepare changelog for 0.12.0 release

denisonbarbosa:
Opening the PR so that we can iterate over it and see what else needs to go in and go out.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,43 @@
+adsys (0.12.0) UNRELEASED; urgency=medium
+  [ Denison Barbosa ]
+  [ Didier Roche ]
+  [ Gabriel Nagy ]
+  [ Jean-Baptiste Lallement ]
+  * Release 0.12.0 (LP: #2020682)
+    - Fix DCONF_PROFILE not considering default_domain_suffix on sssd.conf
+    - Go implementation for the user mount handler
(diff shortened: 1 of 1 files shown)


debian/changelog
```diff
@@ -1,3 +1,38 @@
+adsys (0.12.0) mantic; urgency=medium
+  [ Denison Barbosa ]
+  [ Didier Roche ]
+  [ Gabriel Nagy ]
+  [ Jean-Baptiste Lallement ]
+  * Fix DCONF_PROFILE not considering default_domain_suffix on sssd.conf
+  * Go implementation for the user mount handler
+  * Remove Rust source code from adsys
+  * Rework Kerberos ticket handling logic
```
denisonbarbosa, line 9: @GabrielNagy, I feel like this isn't enough to say everything about what you did for the samba stuff, so what else should I add?

debian/changelog
```diff
@@ -1,3 +1,42 @@
+adsys (0.12.0) mantic; urgency=medium
```
didrocks, line 1: If you want to merge it before we really release it (as there is still one of two fixes in the pipe), keep the release `UNRELEASED`. Then, you can merge it anytime (squash it to a single commit probably).

didrocks (changes requested):

PR #713, opened 2023-05-29
Fix a couple of C issues that weren't spotted

denisonbarbosa:
Didn't realize that my editor had intellisense disabled (which is a bad idea for writing C code without being completely fluent), so this PR fixes some things that we missed on the reviews.

### DIFF
--- a/pam/pam_adsys.c
+++ b/pam/pam_adsys.c
@@ -25,6 +25,7 @@
 
 #define _GNU_SOURCE
 
+#include <ctype.h>
 #include <errno.h>
 #include <limits.h>
 #include <pwd.h>
@@ -223,7 +224,18 @@ static char *get_default_sss_domain(pam_handle_t *pamh) {
                 pam_syslog(pamh, LOG_ERR, "Could not find value for key 'default_domain_suffix' in sssd.conf");
                 break;
             }
-            while (*++domain == ' ') continue;
+            // Ignores whitespaces and tabs right after the '='
+            do {
+                domain++;
+            } while (strlen(domain) > 0 && (*domain == ' ' || *domain == '\t'));
+
+            // For cases where sssd.conf has something like "default_domain_suffix =       \n"
+            if (strlen(domain) <= 1) {
+                pam_syslog(pamh, LOG_ERR, "Could not find valid value for 'default_domain_suffix' in sssd.conf");
+                domain = NULL;
+                break;
+            }
+
             char *newline = strchr(domain, '\n');
             if (newline != NULL) {
                 *newline = '\0';
@@ -232,6 +244,11 @@ static char *get_default_sss_domain(pam_handle_t *pamh) {
         }
     }
     fclose(f);
+
+    if (domain == NULL) {
+        free(buf);
+        return NULL;
+    }
 
     char *ret = strdup(domain);
     free(buf);


pam/pam_adsys.c
```diff
@@ -25,6 +25,7 @@
 
 #define _GNU_SOURCE
 
+#include <ctype.h>
```
denisonbarbosa, line 28: We use `toLower()` to normalize the username, but we weren't including the `ctype.h` header, which ended up in a implict declaration (oops)

pam/pam_adsys.c
```diff
@@ -233,6 +234,11 @@ static char *get_default_sss_domain(pam_handle_t *pamh) {
     }
     fclose(f);
 
+    if (domain == NULL) {
+        free(buf);
+        return NULL;
+    }
+
```
denisonbarbosa, line 241: If `domain` is NULL in the end we need to return earlier, otherwise it can cause problems in `strdup()`

CarlosNihelton: Unrelated to your PR, but could `*++domain` evaluate to `\t`, for instance?

denisonbarbosa: It could, yeah. Why?

CarlosNihelton: This part of the code

```c
while (*++domain == ' ') continue;
            char *newline = strchr(domain, '\n');
            if (newline != NULL) {
                *newline = '\0';
            }
            break;
```

allows for `domain`  to contain `\t` and other space-like characters. Imagine a line in sssd.conf like:

```
default_domain_suffix\t=\tmydomain.com
```

It seems to allow for the above. Could that be a problem?

denisonbarbosa: Indeed, it seems like an issue. We'll have to test and see how sssd proceeds in these kinds of situation, but it they account for it, we must do the same as well.
Thanks for spotting it!

pam/pam_adsys.c
```diff
@@ -223,7 +224,16 @@ static char *get_default_sss_domain(pam_handle_t *pamh) {
...
+            // Ignores whitespaces and tabs right after the '='
+            domain++;
+            while (strlen(domain) > 0 && (*domain == ' ' || *domain == '\t')) domain++;
+
+            if (strlen(domain) == 0) {
+                pam_syslog(pamh, LOG_ERR, "Could not find value for key 'default_domain_suffix' in sssd.conf");
+                domain = NULL;
+                break;
+            }
+
```
denisonbarbosa, line 236: As @CarlosNihelton pointed out, we weren't considering cases when sssd.conf had something like `default_domain_suffix = \t domain.com`, so it could create issues when parsing it. I also added a check for cases where the config has something like `default_domain_suffix =         \n`.

pam/pam_adsys.c
```diff
@@ -223,7 +224,17 @@ static char *get_default_sss_domain(pam_handle_t *pamh) {
                 pam_syslog(pamh, LOG_ERR, "Could not find value for key 'default_domain_suffix' in sssd.conf");
                 break;
             }
-            while (*++domain == ' ') continue;
+            // Ignores whitespaces and tabs right after the '='
+            domain++;
+            while (strlen(domain) > 0 && (*domain == ' ' || *domain == '\t')) domain++;
```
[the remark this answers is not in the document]
denisonbarbosa: Damn, I completely forgot that this loop structure exists :joy:

pam/pam_adsys.c
```diff
@@ -223,7 +224,18 @@ static char *get_default_sss_domain(pam_handle_t *pamh) {
                 pam_syslog(pamh, LOG_ERR, "Could not find value for key 'default_domain_suffix' in sssd.conf");
                 break;
             }
-            while (*++domain == ' ') continue;
+            // Ignores whitespaces and tabs right after the '='
+            do {
+                domain++;
+            } while (strlen(domain) > 0 && (*domain == ' ' || *domain == '\t'));
```
denisonbarbosa, line 230: Is it better, though?

denisonbarbosa: There isn't a one liner, I think. The shortest way would be:
```C
do domain++;
while (strlen(domain) > 0 && (*domain == ' ' || *domain == '\t'));
```
But it looks so wrong

didrocks (approved):
Looks good with the latest fixes (even with the do while… hem :p)

PR #716, opened 2023-05-31
Fix [Install] target of adsys-user-mounts.service

denisonbarbosa:
Having `WantedBy=default.target` for the user mounts service was causing it to be installed under `graphical.target`, which is too early for `gvfs-daemon.service` to have the correct environment.

Switching to `WantedBy=graphical-session.target` delays the unit starting process a little bit and ensures we'll have the Kerberos ticket available at `KRB5CCNAME` necessary to handle authenticated mounts.

UDENG-518


didrocks (approved):
As discussed together… +1

PR #718, opened 2023-06-01
Address a few issues in smbsafe_test.go

denisonbarbosa:
- Increase sleep time to be a little more than the time we expect the routine to wait to prevent some time mismatches.
- Change type conversion to uint64 rather than int64 to avoid overflows.

### DIFF
--- a/internal/smbsafe/smbsafe_test.go
+++ b/internal/smbsafe/smbsafe_test.go
@@ -126,11 +126,11 @@ func TestMultipleSmbLocksOnlyReleaseOnLast(t *testing.T) {
 func shouldHaveWaited(t *testing.T, startpoint time.Time) {
 	t.Helper()
 
-	require.Less(t, int64(waitTime), int64(time.Since(startpoint)), "Should have waited")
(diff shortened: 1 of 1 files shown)


internal/smbsafe/smbsafe_test.go
```diff
@@ -54,7 +54,7 @@ func TestExclusiveLockSmb(t *testing.T) {
 	}()
 
 	// Wait more than minimum duration time to release the lock
-	time.Sleep(waitTime)
+	time.Sleep(waitTime + time.Millisecond)
```
didrocks, line 57: Instead of increasing the `time.Sleep` here, I would rather modify shouldHaveWaited and shouldNotHaveWaited to remove this millisecond. That way, all tests depending on it are impacted

didrocks (changes requested):

internal/smbsafe/smbsafe_test.go
denisonbarbosa: I agree. It's better to change directly in the helpers. It's updated now! Increased the thresholds by `+/- time.Millisecond` (`+` for comparisons where it should be higher and `-` for the ones where it should be lower)

didrocks (approved):
exactly the way I forsaw it!

PR #722, opened 2023-06-06
Fix typo on build command for the admxgen package

denisonbarbosa:
There was a typo in the build command for the admx package, which was resulting in the package not building with CGo disabled and thus causing problems when generating adm* files for different distros.


didrocks (approved):
easy fix :)

PR #724, opened 2023-06-06
Update policy definitions to include dconf key for dark mode background

### LINKED ISSUE #701: Update dconf policy manager with new key for dark mode background
Dconf now have a different key for the background of the dark mode. Right now, if the user background is set through the dconf policy (i.e. `Background > Picture URI`), this will only change the light mode background. We need to add the option to configure the dark mode background (or we could apply the same URI for both keys).

denisonbarbosa:
Adds policy to also configure the background picture for the dark mode, which uses a different key: picture-uri-dark.

Closes #701 / UDENG-340

### DIFF
--- a/cmd/admxgen/defs/categories.yaml
+++ b/cmd/admxgen/defs/categories.yaml
@@ -21,6 +21,7 @@ categories:
         defaultpolicyclass: "User"
         policies:
           - "/org/gnome/desktop/background/picture-uri"
+          - "/org/gnome/desktop/background/picture-uri-dark"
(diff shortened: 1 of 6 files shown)


denisonbarbosa:
ADMX generation for Focal failed because #722 isn't merged yet.

didrocks:
can you rebase to ensure the admx is all good then?

didrocks (approved):
let’s get all dark, the list of supported distro sounds correct to me :)

PR #762, opened 2023-08-08
Cleanly exit if cepces or getcert are not present

GabrielNagy:
In case `certmonger` or the `cepces` library are not installed we do not want to fail hard. Check for the existence of the binaries instead of debs to be more future-proof given that cepces is not yet packaged in Ubuntu.

Fixes UDENG-1156

--------------------------------------

The PR contains some logging improvements as well, and a small optimization to avoid creating the needed directories if the required binaries are not present.


didrocks (approved):
No remark to add. I think we should think when writing the doc to mention how to install them, but otherwise, it looks really great!

PR #765, opened 2023-08-10
Update packaging to support certificates policy

GabrielNagy:
See each individual commit for details.

Fixes UDENG-1157

Passing PPA build: https://launchpad.net/~gabuscus/+archive/ubuntu/ppa/+sourcepub/15065373/+listing-archive-extra
Local autopkgtests passing as well for amd64


denisonbarbosa (approved):
Looks good to me! I just wonder if the changelog update shouldn't be on a PR of its own, so it's easier for us to keep track of the releases (thoughts?).

didrocks:
Looks good to me too. I agree with @denisonbarbosa though: can we get the changelog in a separate PR?

PR #768, opened 2023-08-16
Workflow to auto-patch vendored Samba code

GabrielNagy:
Since our vendored Samba code has a few patches, a good way to ensure we keep it in sync with upstream is to apply an auto-patch workflow where we take the original files and apply a series of patches to get the
files to the final state. This way we ensure that we don't miss any of the changes that happen upstream.

The implementation is self-explanatory for the most part, taking inspiration from our other automated PR workflows. Auto-merging is disabled to give maintainers the opportunity to review (and test) the changes before merging anything in.

This means, in addition to vendoring the "final" versions in `./internal/policies/certificate`, we also need to vendor the upstream versions, and the series of patches to apply.

As there's no reliable way to trigger this workflow only on upstream code changes (i.e. webhooks), the next best thing is to have the workflow run on schedule. We don't expect changes to the vendored part of our codebase given that it's been around 5-6 months since the last commits, so running it on a weekly cadence should suffice. I'm also leaving the `workflow_dispatch` trigger on in case we want to run it on demand.

If the patching fails, the PR body will contain the hunks that failed to apply. In this case, the developer is expected to manually perform the actions of the workflow, updating the patches so they are applicable to the new Samba version.

Fixes UDENG-1113

### DIFF
--- a/.github/workflows/patch-vendored-samba.yaml
+++ b/.github/workflows/patch-vendored-samba.yaml
@@ -1,0 +1,91 @@
+name: Patch vendored Samba code
+
+on:
+  schedule:
+    - cron: '0 9 * * 1' # run on a weekly cadence
+  workflow_dispatch:
+
+env:
+  checkout_files: |
+    python/samba/gp/gp_cert_auto_enroll_ext.py
+    python/samba/gp/gpclass.py
+    python/samba/gp/util/logging.py
+
+jobs:
+  check-for-changes:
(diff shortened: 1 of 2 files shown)


denisonbarbosa (commented):
LGTM! Did you test it in a separate repo? I'd like to see how the PR would look like when there's changes to be made.

denisonbarbosa (approved):
Great! Thanks for linking the example PR!

PR #789, opened 2023-09-25
build: Fix pam_adsys build

GabrielNagy:
It looks like multiple `-tags` statements override each other so we must specify them in a comma separated way. This fixes the issue where `pam_adsys` is not built and shipped anymore.

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -4,8 +4,9 @@
 export GOCACHE=$(shell mktemp -d /tmp/gocache-XXXX)
 
 WIN_ARCHS := amd64 arm64
+# Tags should always come last as we append to them in other parts of the build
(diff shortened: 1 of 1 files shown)


debian/rules
```diff
@@ -5,7 +5,7 @@ export GOCACHE=$(shell mktemp -d /tmp/gocache-XXXX)
 
 WIN_ARCHS := amd64 arm64
 COMMON_GOFLAGS := -ldflags=-X=github.com/ubuntu/adsys/internal/consts.Version=$(shell dpkg-parsechangelog -S Version) -mod=vendor -tags=withmo
```
didrocks, line 7: Do you mind writing a comment here telling that tags should be last (as we will append it to other tags)?

didrocks (approved):
Thanks for working and fixing this promptly!

I just have a small suggestion on commenting, then, feel free to merge :)

EDIT: if you didn’t notice, the vulnerability is due to us needing rebuilding with Go 1.21.1. Let me quickly do a PR with this.

PR #790, opened 2023-09-25
build: Use go 1.21.1 to avoid security vulnerability

didrocks:
Ensure we are using the safe Go version as we are using parsing (even if we only do this in documentation generation).
Update packaging too.


didrocks:
Actuall, it’s markdown itself which is vulnerable, closing.

didrocks:
Probably because it’s an unstable fix version? Unsure.

PR #818, opened 2023-10-23
tests(e2e): Add barebones E2E test workflow

GabrielNagy:
Add a deprovisioning script and a GitHub Actions workflow that exercises the `provision_resources` suite which builds the ADSys DEB for all codenames then provisions a VM on Azure, joins it to the domain, installs ADSys, ultimately deprovisioning it.

This provides the main building block on top of which we will add the step(s) to run the actual tests.

For now the workflow will only be triggered by hand until we decide in what cases it's best for it to run automatically (being mindful of test duration and Azure costs).

A passing run can be seen on my fork here: https://github.com/GabrielNagy/adsys/actions/runs/6616242848

Fixes UDENG-1359

### DIFF
--- a/e2e/scripts/patches/jammy.patch
+++ b/e2e/scripts/patches/jammy.patch
@@ -1,5 +1,5 @@
 diff --git a/debian/control b/debian/control
-index d6fb19f2..85eab447 100644
+index d6fb19f2..93764223 100644
 --- a/debian/control
 +++ b/debian/control
 @@ -5,7 +5,7 @@ Maintainer: Ubuntu Developers <ubuntu-devel-discuss@lists.ubuntu.com>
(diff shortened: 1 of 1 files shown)


denisonbarbosa (approved):
LGTM. Good job and thanks for linking an example run as well!

.github/workflows/e2e-tests.yaml
```diff
@@ -0,0 +1,93 @@
+name: E2E - Run tests
+
+on:
+  workflow_dispatch:
+  # push:
```
didrocks, line 5: do you want to only run on `workflow_dispatch` for now until having all the ests ready?

.github/workflows/e2e-tests.yaml
```diff
@@ -0,0 +1,93 @@
...
+jobs:
+  supported-releases:
+    name: Build matrix for supported ADSys and Ubuntu releases
+    runs-on: ubuntu-latest
+    outputs:
+      matrix: ${{ steps.set-supported-releases.outputs.matrix }}
+    steps:
+      - name: Install needed binaries
+        run: |
+          sudo DEBIAN_FRONTEND=noninteractive apt update
```
didrocks, line 21: no need for this env variable on apt update (yeah, maybe it’s on another workflow I wrote, but shhhhh)

didrocks (dismissed):
That one was probably way easier to review than to write :) I just have an opened comment + a small fix suggestion! (you don’t use slog, but IIRC, that’s the case for other package in the same directory and we told that we would do it in one shot, no?)

Pre-approving based on this. Well done!

didrocks (approved):
Thanks for the changes!

denisonbarbosa (approved):

PR #825, opened 2023-10-27
docs: Removed empty tutorial section

jibel:
There is no tutorial at the moment so just drop it. We will readd it later once there are some.


didrocks (approved):
Looks good! I change your PR title to just "docs"

PR #826, opened 2023-10-27
fix(docs): Retries link checker on failure

jibel:
Some Ubuntu websites are unstable so retry up to 3 times.


didrocks (approved):
Let’s roll!

jibel:
Thanks for the review!

PR #828, opened 2023-10-27
tests(e2e): Configure AD before starting test suite

GabrielNagy:
The workflow is described in the Go/PowerShell scripts for the most part, but at a high level the new `02_provision_ad` step achieves the following:

- converts XML GPOs to POL
- uploads GPO structure to the DC under a client-specific path
- uploads and runs the PowerShell script responsible for creating client-specific OUs, GPOs and users

The PR contains more changes best reviewed commit by commit.

Fixes UDENG-1535

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,6 +40,7 @@ debian/adsys-windows
 
 # E2E testing
 inventory.yaml
+e2e/assets/gpo/**/*.pol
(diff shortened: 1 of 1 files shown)


denisonbarbosa (approved):
Amazing work!

PR #833, opened 2023-10-31
tests(e2e): Base VM image generation improvements

GabrielNagy:
Details are in the commit messages, basically this tightens up the base VM provisioning part and adds some updates that will make our life easier when running tests.

I've built a noble template locally and with these changes and confirmed they work properly.


denisonbarbosa (approved):
LGTM!

Issue #834, opened 2023-11-01
docs recommend wrong directory ProgramData for deployment of Active Directory Watch Daemon (adwatchd.exe)


Buster-de:
### Is there an existing issue for this?

- [X] I have searched the existing issues and found none that matched mine

### Describe the issue

https://github.com/ubuntu/adsys contains:

> After a successful installation, the `adwatchd.exe` executable will be available in the `/usr/share/adsys/windows` directory. We suggest you deploy this executable to a persistent directory of your choosing on the AD Domain Controller, such as `C:\ProgramData\Ubuntu\adsys\`.

ProgramData should only contain data, not executables.

"ProgramData" should be "Program Files", see https://learn.microsoft.com/en-us/windows-hardware/customize/desktop/unattend/microsoft-windows-shell-setup-folderlocations-programdata

Configuration file can be located in ProgramData, if adwatchd.exe does not only support reading current directory.

### Steps to reproduce it

read https://canonical-adsys.readthedocs-hosted.com/en/latest/how-to/set-up-adwatchd/ and compare with Microsoft's recommendation

### Double check your logs

- [X] I have redacted any sensitive information from the logs

jibel:
Thanks for reporting this issue. The doc will be updated accordingly.

PR #847, opened 2023-11-13
Fix installation path of adwatchd

jibel:
Replaced "ProgramData" with "Program Files" to follow MS documentation.

Closes: #834 

UDENG-1699


didrocks (approved):
Indeed, this is a better place for this executable!

PR #852, opened 2023-11-16
docs(completion): skip generated reference policies for completion

didrocks:
This documentation will just pollute the completion with many items and is less worthwhile on the command line.

-----

UDENG-1039


jibel (approved):
👍

PR #853, opened 2023-11-16
docs: sphinx reference doc index generation

didrocks:
Run a pre-hook to create index file in policy reference.
To be able to list the whole policy reference definition, we needs index file for each category.
Create them on the fly, listing the reference policies generated in the reference destination directory, before generating the finale documentation.

-----

UDENG-1450


jibel (approved):
👍

PR #855, opened 2023-11-16
ci(docs): reduce the amount of workflow to be run

didrocks:
* reduce the amount of documentation check in parallel
* remove now undeeded github wiki page autoupdate
* only trigger doc workflow on docs/ changes


jibel (approved):
Sounds good!

PR #857, opened 2023-11-20
docs: added tutorial section

jibel:
Added the section as part of the main index.
There is an initial tutorial which is actually a demo of certificates auto-enrollment.

UDENG-1787


didrocks (approved):
I don’t have an immediate answer on the failing test. I created a card (unassigned for now) in the current sprint as this should block our release too: https://warthogs.atlassian.net/browse/UDENG-1788

PR #886, opened 2024-01-02
deps(i18n): Move to external go-i18n and gotext

didrocks:
We did the work to avoid our internal/ package last cycle. Move now to
it.
We still need to validate each lines that we log or not.

This removes the need for snapcore go-gettext.

UDENG-1990


denisonbarbosa (approved):
LGTM!

PR #891, opened 2024-01-15
tests: add e2e tests for all policy managers

GabrielNagy:
I recommend reviewing this one commit by commit since it includes lots of minor fixes and improvements to support the e2e tests implementation. There's more context in the individual commit messages too.

I've ran the suite on focal, jammy and mantic and got passing runs, however I'll tinker with them some more once everything is set up on GitHub Actions (as part of a follow-up card).

Because this PR includes changes to the VM template provisioning script, we will have to refresh the templates once this is merged.

Fixes UDENG-1360

### DIFF
--- a/e2e/cmd/run_tests/04_test_pro_managers/main.go
+++ b/e2e/cmd/run_tests/04_test_pro_managers/main.go
@@ -110,8 +110,8 @@ func action(ctx context.Context, cmd *command.Command) error {
 		return fmt.Errorf("failed to confirm client is attached to Ubuntu Pro: %w", err)
 	}
 
-	// Reboot machine to re-apply machine policies
-	if err := client.Reboot(); err != nil {
+	// Start timer-triggered service to update policies
+	if _, err := client.Run(ctx, "systemctl restart adsys-gpo-refresh"); err != nil {
 		return err
 	}
 
@@ -147,7 +147,7 @@ func action(ctx context.Context, cmd *command.Command) error {
 	}
 
 	/// Scripts
-	if err := client.RequireFileExists(ctx, "/etc/created-by-adsys-machine-startup-script"); err != nil {
+	if err := client.RequireNoFileExists(ctx, "/etc/created-by-adsys-machine-startup-script"); err != nil {
 		return err
 	}
 	if err := client.RequireNoFileExists(ctx, "/etc/created-by-adsys-machine-shutdown-script"); err != nil {
@@ -182,11 +182,14 @@ ftp_proxy="http://127.0.0.1:8080"`); err != nil {
 		}
 	}
 
-	// Reboot and check shutdown script
+	// Reboot and check machine scripts
 	if err := client.Reboot(); err != nil {
 		return err
 	}
 	if err := client.RequireFileExists(ctx, "/etc/created-by-adsys-machine-shutdown-script"); err != nil {
+		return err
+	}
+	if err := client.RequireFileExists(ctx, "/etc/created-by-adsys-machine-startup-script"); err != nil {
 		return err
 	}
 


e2e/internal/remote/remote.go
```diff
@@ -197,3 +208,48 @@ func (c Client) Upload(localPath string, remotePath string) error {
...
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
+	defer cancel()
+
+	// Wait for the host to come back online
+	for {
+		select {
+		case <-ctx.Done():
+			return fmt.Errorf("host %q did not come back online after reboot", c.client.RemoteAddr().String())
+		default:
```
denisonbarbosa, line 246: You only use the context for timing out the loop here. Can't we just do ```<-time.After(time.Minute)``` instead?

e2e/cmd/run_tests/03_test_non_pro_managers/main.go
```diff
@@ -0,0 +1,116 @@
...
+The run is considered successful if the script exits with a zero exit code.
+
+The runner must be connected to the ADSys E2E tests VPN.`, filepath.Base(os.Args[0]))
+
+	return cmd.Execute(context.Background())
+}
+
+func validate(_ context.Context, cmd *command.Command) error {
+	var err error
+	sshKey, err = command.ValidateAndExpandPath(cmd.Inventory.SSHKeyPath, command.DefaultSSHKeyPath)
```
denisonbarbosa, line 48: We could use named return instead and avoid having the `var err error` line, but I'm not 100% sure what's the best approach here.

denisonbarbosa (commented):
I like it a lot, nicely done! Just one comment, but no big deal

e2e/internal/remote/remote.go
```diff
@@ -21,41 +21,57 @@ import (
...
-	return Client{client}, nil
+	return Client{
+		client: client,
+		config: config,
+		host:   host,
+	}, nil
+}
+
+// Close closes the SSH connection.
+func (c Client) Close() error {
```
didrocks, line 73: I would put a pointer on Client here, just to signal we change something on the underlying object (even if it’s a copy of an object handling a pointer has no special effect, but at least, semantically, we signal it correct)

e2e/scripts/patches/focal.patch
```diff
@@ -10,8 +10,8 @@ index d6fb19f2..f9f9fe0c 100644
 +Build-Depends: debhelper-compat (= 12),
                 dh-apport,
                 dh-golang,
--               golang-go (>= 2:1.20~),
-+               golang-1.20-go,
+-               golang-go (>= 2:1.21~),
++               golang-1.21-go,
```
didrocks, line 14: I always hate to read diffs of diffs :p

didrocks (changes requested):
Excellent work! I don’t have that may things to suggest, only minor things I spot.

I think this is really a very good way for us to ensure that most of our policies and base commands works.

Just curious, do you consider that the non pro and then pro attach + reboot is what is covering the "reupdate policies?"

I think there is also reupdate policies while the machine is booted, is there any plans for this? (I think we discussed that one the spec, but can’t remember). But already, just the fact that we got to that e2e level of tests, I’m thrilled! :)

e2e/cmd/run_tests/04_test_pro_managers/main.go
```diff
@@ -0,0 +1,248 @@
...
+	}
+
+	// Assert admin GPO policies were applied
+	client, err = remote.NewClient(cmd.Inventory.IP, fmt.Sprintf("%s-adm@warthogs.biz", cmd.Inventory.Hostname), remote.DomainUserPassword)
+	if err != nil {
+		return fmt.Errorf("failed to connect to VM: %w", err)
+	}
+
+	/// User mounts
+	// Workaround user mounts only being applied on graphical sessions
```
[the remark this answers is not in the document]
didrocks: retraced, Denis corrected me :) 

I think maybe a comment with the new client will be helpful!

didrocks (approved):
All good from my POV! Excellent work :)

denisonbarbosa (approved):
Looks good to me. Nicely done!

PR #898, opened 2024-01-30
ci: use commit prefix for Samba auto-update workflow

GabrielNagy:
Only set it in the PR title which will be used for the merge commit message.


didrocks (approved):
indeed!

PR #903, opened 2024-02-06
feat: get ticket path from krb5 when updating current user policies

GabrielNagy:
This leverages the libkrb5 API via CGo to determine the default path of the current user's cached ticket. According to the MIT Kerberos [documentation](https://web.mit.edu/kerberos/www/krb5-latest/doc/appdev/refs/api/krb5_cc_default_name.html) the order is:
- a prior call to `krb5_cc_set_default_name()`
- the `KRB5CCNAME` environment variable
- the `default_ccache_name` profile variable
- the operating system or build-time default value

The C function doesn't guarantee that the ticket is present on disk (or that the ticket is of `FILE` type for that matter), so we do the additional checks ourselves and return an error as needed.

Testing is accomplished via a C mock with a mention that integration tests will look a bit different in the sense that we will not override `krb5_init_context` for them, since that is a critical path required by samba to work. This is not a huge loss as we do more granular tests at the package level.

Similar to the `wbclient` mock, hoist the compilation logic to `testutils` in order to make it accessible from the integration tests.

-----

For now this logic is only used when updating policies for the current user and `KRB5CCNAME` is not set. The functionality is gated behind a new setting, `detect_cached_ticket`, which we default to false to maintain backwards compatibility. The same implementation will be used for the PAM module (where we will have a CLI command that runs `setuid` with the target user before calling `ad.TicketPath`), but I opened this one first for an easier review.

(Partly) fixes UDENG-2191

### DIFF
--- a/internal/ad/krb5_test.go
+++ b/internal/ad/krb5_test.go
@@ -3,7 +3,6 @@ package ad_test
 import (
 	"fmt"
 	"os"
-	"os/exec"
 	"path/filepath"
 	"strings"
 	"testing"
@@ -21,7 +20,7 @@ func TestTicketPath(t *testing.T) {
 	}
 
 	// We setup and rerun in a subprocess because we need to preload the mock libkrb5
-	if setupSubprocessForTest(t, mockLibPath) {
+	if testutils.PreloadLibInSubprocess(t, mockLibPath) {
 		return
 	}
 
@@ -71,62 +70,4 @@ func TestTicketPath(t *testing.T) {
 			require.Equal(t, wantOut, ticketPath, "Returned ticket path is not the expected one")
 		})
 	}
-}
-
-// setupSubprocessForTest prepares a subprocess preloading a shared library for running the tests.
-// Returns false if we are already in the subprocess and should continue.
-// Returns true if we prepare the subprocess and reexec ourself.
-func setupSubprocessForTest(t *testing.T, mockLibPath string) bool {
-	t.Helper()
-
-	if os.Getenv("GO_WANT_HELPER_PROCESS") == "1" {
-		return false
-	}
-
-	var subArgs []string
-	// We are going to only reexec ourself: only take options (without -run)
-	// and redirect coverage file
-	var hasExplicitTestAsRunArg bool
-	for i, arg := range os.Args {
-		if i != 0 && !strings.HasPrefix(arg, "-") {
-			continue
-		}
-		if strings.HasPrefix(arg, "-test.run=") {
-			if !strings.HasPrefix(arg, fmt.Sprintf("-test.run=%s", t.Name())) {
-				continue
-			}
-			hasExplicitTestAsRunArg = true
-		}
-		if strings.HasPrefix(arg, "-test.coverprofile=") {
-			continue
-		}
-		subArgs = append(subArgs, arg)
-	}
-	// Cover subprocess in a different file that we will merge when the test ends
-	if testCoverFile := testutils.TrackTestCoverage(t); testCoverFile != "" {
-		subArgs = append(subArgs, "-test.coverprofile="+testCoverFile)
-	}
-
-	if !hasExplicitTestAsRunArg {
-		subArgs = append(subArgs, fmt.Sprintf("-test.run=%s", t.Name()))
-	}
-
-	t.Log("Running subprocess with", subArgs)
-	// #nosec G204: this is only for tests, under controlled args
-	cmd := exec.Command(subArgs[0], subArgs[1:]...)
-
-	// Setup correct child environment, including LD_PRELOAD for wbclient mock
-	cmd.Env = append(os.Environ(),
-		"GO_WANT_HELPER_PROCESS=1",
-		// override system libwbclient
-		fmt.Sprintf("LD_PRELOAD=%s", mockLibPath),
-	)
-
-	cmd.Stderr = os.Stderr
-	cmd.Stdout = os.Stdout
-	if err := cmd.Run(); err != nil {
-		t.Fail() // The real failure will be written by the child test process
-	}
-
-	return true
 }
--- a/internal/ad/backends/winbind/winbind_test.go
+++ b/internal/ad/backends/winbind/winbind_test.go
@@ -5,9 +5,7 @@ import (
 	"flag"
 	"fmt"
 	"os"
-	"os/exec"
 	"path/filepath"
-	"strings"
 	"testing"
 
 	"github.com/sirupsen/logrus"
@@ -24,7 +22,7 @@ func TestWinbind(t *testing.T) {
 	}
 
 	// We setup and rerun in a subprocess because we need to preload the mock libwbclient
-	if setupSubprocessForTest(t, mockLibPath) {
+	if testutils.PreloadLibInSubprocess(t, mockLibPath) {
 		return
 	}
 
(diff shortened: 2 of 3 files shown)


internal/ad/krb5.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+	krb5cc := C.GoString(cKrb5cc)
+	if krb5cc == "" {
+		return "", errors.New(gotext.Get("path is empty"))
+	}
+
+	krb5ccPath, found := strings.CutPrefix(krb5cc, "FILE:")
+	if !found {
+		log.Warningf(ctx, gotext.Get("Krb5cc path %q does not start with FILE:", krb5cc))
+	}
```
denisonbarbosa, line 59: Does this matter from our POV or does everything still work if the ticket is not a file? Sincere question.

internal/ad/krb5_test.go
```diff
@@ -0,0 +1,131 @@
...
+	// Cover subprocess in a different file that we will merge when the test ends
+	if testCoverFile := testutils.TrackTestCoverage(t); testCoverFile != "" {
+		subArgs = append(subArgs, "-test.coverprofile="+testCoverFile)
+	}
+
+	if !hasExplicitTestAsRunArg {
+		subArgs = append(subArgs, fmt.Sprintf("-test.run=%s", t.Name()))
+	}
+
+	fmt.Println("Running subprocess with", subArgs)
```
denisonbarbosa, line 113: s/`fmt.Println`/`t.Log`

internal/ad/mock/libkrb5_mock.c
```diff
@@ -0,0 +1,31 @@
...
+    } else if (strcmp(behavior, "return_null_ccache") == 0) {
+        return NULL;
+    } else if (strstr(behavior, "return_ccache") != NULL) {
+        // split the string by the first colon
+        char *ccname = strtok(behavior, ":");
+        ccname = strtok(NULL, "");
+        return ccname;
+    } else if (strstr(behavior, "return_memory_ccache") != NULL) {
+        return "MEMORY:foo";
+    }
```
denisonbarbosa, line 27: nitpick: Since it's returning on every `if` block, we could remove the `else if` statements and turn this into 4 different `if`s. IMHO, it could improve readability.

denisonbarbosa (commented):
Looks good to me. Just a few suggestions...

internal/ad/krb5.go
denisonbarbosa: I don't think we should drop the check here, honestly. If the API returns something without the `FILE:` prefix, we should return an error instead of delegating the failure downstream, IMHO.

Looking at some docs (https://web.mit.edu/kerberos/krb5-1.12/doc/basic/ccache_def.html#ccache-types), there are quite some possible `ccache` types. Maybe we should cut the prefix and, if it's not `FILE` we return an error mentioning that prefix X is not supported by us?

denisonbarbosa: `access` would not fail for a dir, would it?

denisonbarbosa (approved):
LGTM. Well done!

internal/ad/krb5_test.go
```diff
@@ -0,0 +1,136 @@
...
+		ccacheIsDir  bool
+
+		wantErr bool
+	}{
+		"Lookup is successful":                 {krb5Behavior: "return_ccache:FILE:%s", createCCache: true},
+		"Allow ccache without FILE identifier": {krb5Behavior: "return_ccache:%s", createCCache: true},
+
+		"Error when ccache not present on disk": {krb5Behavior: "return_ccache:FILE:%s", wantErr: true},
+		"Error when ccache is a directory":      {krb5Behavior: "return_ccache:%s", createCCache: true, ccacheIsDir: true, wantErr: true},
+		"Error when initializing context":       {krb5Behavior: "error_initializing_context", wantErr: true},
```
didrocks, line 40: here, we don’t know if the error is due to `createCCache` not being true or the `krb5Behavior` (even if ok, the code is influenced by the krb5Behavior before the cache existence check). Shouldn’t we have a default to create a ccache (and so, we remove this repeated parameter everywhere) and then, have a noCreateCCache for the specific error case?

internal/ad/krb5_test.go
```diff
@@ -0,0 +1,136 @@
...
+
+			// Set up mock libwbclient behavior
+			t.Setenv("ADSYS_KRB5_BEHAVIOR", tc.krb5Behavior)
+
+			if tc.createCCache {
+				var err error
+				if tc.ccacheIsDir {
+					err = os.Mkdir(wantOut, 0700)
+				} else {
+					err = os.WriteFile(wantOut, []byte("dummy ticket data"), 0600)
```
didrocks, line 63: so, of course, that would mean that we have a default ccache path to return :)

internal/ad/mock/libkrb5_package_tests_mock.c
```diff
@@ -0,0 +1,14 @@
+#include <krb5.h>
+#include <stdio.h>
+#include <string.h>
+
+#include "libkrb5_mock.h"
+
+krb5_error_code KRB5_CALLCONV krb5_init_context(krb5_context *context) {
```
didrocks, line 7: don’t lie, you are getting used to write mocks in C :p

internal/testutils/backends.go
```diff
@@ -69,3 +69,29 @@ func BuildWinbindMock(t *testing.T, goPkgPath string) string {
...
 }
+
+// BuildKrb5Mock takes the path to the location of the ad internal package and
+// builds the libkrb5 mock for use with package or integration tests.
+func BuildKrb5Mock(t *testing.T, goPkgPath string) string {
+	t.Helper()
+
+	filesToCompile := []string{filepath.Join(goPkgPath, "mock", "libkrb5_mock.c")}
+
+	// We only compile the krb5_init_context mock for package tests, as on
```
didrocks, line 80: Waow, well done!

However, I am not a big fan of the detection relying on "." itself. Do you think we can lookup (easily, don’t die over it) if we link against the real krb5 lib in the current process? (we can’t rely on the symbol itself: `krb5_init_context` from the krb5 lib will be probably be loaded after us, as we `LD_PRELOAD`). Maybe it’s too much, but this sounds a little bit too black magic to me :)

cmd/adsysd/integration_tests/adsys_test.go
```diff
@@ -257,7 +264,9 @@ apparmor_dir: %s/apparmor.d/adsys
 apparmorfs_dir: %s/apparmorfs
 systemunit_dir: %s/systemd/system
 global_trust_dir: %s/share/ca-certificates
-`, args.adsysDir, args.adsysDir, args.adsysDir, args.adsysDir, args.backend, args.adsysDir, args.adsysDir, args.adsysDir, args.adsysDir, args.adsysDir, args.adsysDir, args.adsysDir, args.adsysDir))
+
+detect_cached_ticket: %t
+`, args.adsysDir, args.adsysDir, args.adsysDir, args.adsysDir, args.backend, args.adsysDir, args.adsysDir, args.adsysDir, args.adsysDir, args.adsysDir, args.adsysDir, args.adsysDir, args.adsysDir, args.detectCachedTicket))
```
didrocks, line 269: one more, one more! (yeah, this is starting to grow… uncontrolled, but I don’t have a good solution TBH)

.github/workflows/qa.yaml
```diff
@@ -10,7 +10,7 @@ on:
 env:
   apt_dependencies: >-
     ca-certificates curl dconf-cli gcc gettext git libnss-wrapper libsmbclient-dev
-    libwbclient-dev pkg-config python3-coverage samba sudo
+    libkrb5-dev libwbclient-dev pkg-config python3-coverage samba sudo
```
didrocks, line 13: shouldn’t we add it explicitely as a build-dep in debian/control too?

cmd/adsysd/integration_tests/adsysctl_policy_test.go
```diff
@@ -871,6 +901,27 @@ func TestPolicyUpdate(t *testing.T) {
 			},
 			wantErr: true,
 		},
+		// Krb5 library error cases
+		"Error when libkrb5 ccache not present on disk": {
+			initState:          "localhost-uptodate",
+			noExportKrb5cc:     true,
+			detectCachedTicket: true,
+			krb5MockBehavior:   "return_ccache:%s/not_present",
```
didrocks, line 909: so, here, you are using a special name on the mock and not an extra argument as in the other test. Maybe you should follow the same convention in the package tests?

didrocks (changes requested):
Excellent work as usual! I have very few things that I think we can enhance on this PR, but impressive digging and perfect execution as usual :)

.github/workflows/qa.yaml
didrocks: Yeah, no black magic from debhelper here, it needs to be explicit (well, it’s probably pulled by sss -dev package). But as we rely on it, let’s ensure this is explicit.

internal/ad/krb5.go
```diff
@@ -0,0 +1,64 @@
...
+)
+
+// TicketPath returns the path of the default kerberos ticket cache for the
+// current user.
+// It returns an error if the path is empty or does not exist on the disk.
+func TicketPath() (string, error) {
+	cKrb5cc, err := C.get_ticket_path()
+	defer C.free(unsafe.Pointer(cKrb5cc))
+	if err != nil {
+		return "", err
```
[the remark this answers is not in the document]
didrocks: Looks good to me!

internal/testutils/backends.go
didrocks: Ok, fine then, I think as this is a testhelper, we can survive this (thans to the comment you have too) :)

didrocks (approved):
On debian/control, thanks for sorting this out but still keep some relative importance ordering. Pure importance ordering is hard in general and subjective, I think your approach makes sense.

Everything’s fine, good to merge! Good work :)

PR #908, opened 2024-02-09
fix(test): Fix require outside of main goroutine

EduardGomezEscandell:
If a `require` check fails outside of the main goroutine, it causes a panic.

This test is a common pattern in our projects, so I went looking for it after having golangci-lint warn about it in Ubuntu Pro for WSL.

### DIFF
--- a/cmd/adsysd/daemon/daemon_test.go
+++ b/cmd/adsysd/daemon/daemon_test.go
@@ -307,16 +307,16 @@ func startDaemon(t *testing.T, setupEnv bool, args ...string) (app *daemon.App, 
 	a := daemon.New()
 	a.SetArgs(args...)
 
-	ch := make(chan error)
+	errCh := make(chan error)
 	go func() {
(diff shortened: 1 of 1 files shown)


cmd/adsysd/daemon/daemon_test.go
```diff
@@ -308,18 +307,16 @@ func startDaemon(t *testing.T, setupEnv bool, args ...string) (app *daemon.App,
 	a := daemon.New()
 	a.SetArgs(args...)
 
-	wg := sync.WaitGroup{}
-	wg.Add(1)
+	ch := make(chan error)
```
didrocks, line 310: errCh maybe? To signifie this is about the returned error?

didrocks (changes requested):
Small suggestion on the naming

didrocks (approved):

EduardGomezEscandell:
Feel free to merge, I think I don't have permissions.

PR #912, opened 2024-02-14
fix: allow sssd backend to work without ad_domain

### LINKED ISSUE #910: (Regression) sssd backend fails if ad_domain is not set

- [X] I have searched the existing issues and found none that matched mine

### Describe the issue

This is a regression from when we added support for multiple AD backends (see https://github.com/ubuntu/adsys/pull/467)

Previously adsys would use the first domain from `sssd.conf` and potentially override it if `ad_domain` is explicitly set for the domain, see: https://github.com/ubuntu/adsys/blob/32a830f2a8204cc8b896094bad512ed619fbf6b7/internal/adsysservice/adsysservice.go#L279-L280

The current implementation raises an error if we are not able to find an `ad_domain` setting in the domain section, even if we already have a domain (`sssdDomain`): https://github.com/ubuntu/adsys/blob/c68d2cc999d25b1cb408a9e31775a76d2af4c8c7/internal/ad/backends/sss/sss.go#L62-L65

Ideally we should set `domain` to `sssdDomain` if we cannot find a value for `ad_domain`, which will mimic the behavior previous to the refactor.

While by default joining a domain with `realm join` will set the appropriate configuration values in `sssd.conf` so this doesn't happen, this is a regression we should aim to fix.

### Steps to reproduce it

1. Join an AD domain with sssd (e.g. using `realm join`)
2. Install the latest version of adsys, run `adsysctl update -m -vv`, everything should work
3. Comment out the `ad_domain` line from `/etc/sssd/sssd.conf`
4. `adsysctl update -m -vv` now fails, and the adsysd service does not start anymore
5. (Optional) To confirm the functionality prior to the regression, re-attempt the steps above on Ubuntu 22.04 using the adsys version currently in the archive (0.9.2) -- adsys is able to correctly determine the domain even without the `ad_domain` setting.

_No response_

_No response_

### Additional information

_No response_

### Double check your logs

- [X] I have redacted any sensitive information from the logs

GabrielNagy:
This restores the functionality prior to the refactor in PR #467, where the case of having a domain section without the `ad_domain` setting resorted to using the domain from the `sssd.domains` setting. This is valid behavior supported and [suggested](https://sssd.io/docs/ad/ad-provider-manual.html#id4) by sssd.

In addition to that, avoid being too lenient and still raise an error if the domain section is empty or does not exist.

Fixes #910 / UDENG-2268

### DIFF
--- a/internal/ad/backends/sss/sss.go
+++ b/internal/ad/backends/sss/sss.go
@@ -59,9 +59,14 @@ func New(ctx context.Context, c Config, bus *dbus.Conn) (s SSS, err error) {
 	if sssdDomain == "" {
 		return SSS{}, errors.New(gotext.Get("failed to find default sssd domain in sssd.conf"))
 	}
-	domain := cfg.Section(fmt.Sprintf("domain/%s", sssdDomain)).Key("ad_domain").String()
+	domainSection := cfg.Section(fmt.Sprintf("domain/%s", sssdDomain))
+	if len(domainSection.KeyStrings()) == 0 {
+		return SSS{}, errors.New(gotext.Get("could not find AD domain section corresponding to %q, or the section is empty", sssdDomain))
+	}
+	domain := domainSection.Key("ad_domain").String()
 	if domain == "" {
-		return SSS{}, errors.New(gotext.Get("could not find AD domain name corresponding to %q", sssdDomain))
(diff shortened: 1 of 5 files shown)


didrocks (approved):
Perfect, nothing else to add :)

PR #914, opened 2024-02-16
feat: infer krb5 ccache from within the PAM module

GabrielNagy:
This is a follow-up to https://github.com/ubuntu/adsys/pull/903 where the functionality to get the default ticket path from `libkrb5` was added.

The simplest and safest way to use the functionality above fom the PAM module is to delegate most of the responsibility to a CLI command written in Go, that can return the ticket path of any given user. Because the PAM module runs as root, we need to de-escalate to the given user (using `setuid` since libkrb5 treats `setuid` and `seteuid` separately, so the latter is not enough).

The CLI command is specialized for usage from within the PAM module to minimize friction. It reads the configuration file to see if the `detect_cached_ticket` setting is enabled, and avoids returning an error if the ticket path is not present on disk. Thus the PAM module can just execute the command, read its output and silently return if the subprocess exited with 0 and no output was printed.

We block authentication if the subprocess returns an error, or if any buffer-allocation or system calls needed to run and read the output are failing.

On top of this being a fully backwards-compatible change on the account of the opt-in setting, the impact of not having a ticket on disk is minimized since adsys will then treat the user as a non-domain user. There is the risk of the ticket cache on disk being outdated, which will end up failing the `adsysctl update` operation that happens on login, but ensuring the validity of the existing ticket path pointed to by `libkrb5` is the responsibility of the process that generates/renews the ticket, not adsys.

Fixes UDENG-2191

### DIFF
--- a/cmd/adsysd/integration_tests/adsysctl_policy_test.go
+++ b/cmd/adsysd/integration_tests/adsysctl_policy_test.go
@@ -1248,6 +1248,77 @@ func TestPolicyDebugScriptDump(t *testing.T) {
 	}
 }
 
+func TestPolicyDebugTicketPath(t *testing.T) {
+	tests := map[string]struct {
+		username string
+
+		configDisabled bool
+		pathNotPresent bool
+		pathIsDir      bool
+
+		wantOut string
+		wantErr bool
+	}{
+		"Return path for current explicit user": {},
+		"Return path for current implicit user": {username: "-"},
+
+		// No-op cases (return no error and no output)
+		"No-op when path not present on disk":        {pathNotPresent: true},
+		"No-op when detect_cached_ticket is not set": {configDisabled: true},
+
+		// Error cases
+		"Error when passed an invalid user":   {username: "invaliduser", wantErr: true},
+		"Error if ticket path is a directory": {pathIsDir: true, wantErr: true},
+	}
+	for name, tc := range tests {
+		tc := tc
+		t.Run(name, func(t *testing.T) {
+			// Empty username means current user
+			if tc.username == "" {
+				u, err := user.Current()
+				require.NoError(t, err, "Setup: could not get current user")
+				tc.username = u.Username
+			}
+			// "-" username means empty argument, current user is inferred
+			if tc.username == "-" {
+				tc.username = ""
+			}
+
+			// Ensure we start and finish the test with a clean slate on disk
+			uid := os.Getuid()
+			krb5ccname := filepath.Join(os.TempDir(), fmt.Sprintf("krb5cc_%d", uid))
+			err := os.RemoveAll(krb5ccname)
+			require.NoError(t, err, "Setup: could not remove ticket path")
+			t.Cleanup(func() {
+				err := os.RemoveAll(krb5ccname)
+				require.NoError(t, err, "Teardown: could not remove ticket path")
+			})
+
+			if tc.pathIsDir {
+				err := os.MkdirAll(krb5ccname, 0700)
+				require.NoError(t, err, "Setup: could not create ticket directory")
+			} else if !tc.pathNotPresent {
+				err := os.WriteFile(krb5ccname, []byte("Some ticket content"), 0600)
+				require.NoError(t, err, "Setup: could not write ticket content")
+				tc.wantOut = krb5ccname + "\n"
+			}
+
+			if tc.configDisabled {
+				tc.wantOut = ""
+			}
+
+			conf := createConf(t, confDetectCachedTicket(!tc.configDisabled))
+			out, err := runClient(t, conf, "policy", "debug", "ticket-path", tc.username)
+			if tc.wantErr {
+				require.Error(t, err, "command should exit with an error")
+				return
+			}
+			require.NoError(t, err, "command should exit with no error")
+			require.Equal(t, tc.wantOut, out, "command output should match")
+		})
+	}
+}
+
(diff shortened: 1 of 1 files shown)


e2e/cmd/run_tests/05_test_pam_krb5cc/main.go
```diff
@@ -0,0 +1,109 @@
...
+Perform PAM krb5cc-related tests on the provisioned Ubuntu client.
+
+The runner must be connected to the ADSys E2E tests VPN.`, filepath.Base(os.Args[0]))
+
+	return cmd.Execute(context.Background())
+}
+
+func validate(_ context.Context, cmd *command.Command) (err error) {
+	sshKey, err = command.ValidateAndExpandPath(cmd.Inventory.SSHKeyPath, command.DefaultSSHKeyPath)
+	fmt.Println(sshKey)
```
didrocks, line 39: shouldn’t we print only after the error is not nil? I think you want to check that we don’t print anything if we return an error, but I think this is a more targetted test for it, wdyt?

internal/ad/krb5.go
```diff
@@ -54,7 +57,7 @@ func TicketPath() (string, error) {
 	krb5ccPath := strings.TrimPrefix(krb5cc, "FILE:")
 	fileInfo, err := os.Stat(krb5ccPath)
 	if err != nil {
-		return "", fmt.Errorf(gotext.Get("%q does not exist or is not accessible: %v", krb5ccPath, err))
+		return "", errors.Join(ErrTicketNotPresent, err)
```
didrocks, line 60: As this is now part of the API, shouldn’t we change the test to asserts on the error type?

cmd/adsysd/client/policy.go
```diff
@@ -300,6 +319,58 @@ func (a *App) dumpCertEnrollScript() error {
 	return os.WriteFile("cert-autoenroll", []byte(script), 0600)
 }
 
+// printTicketPath prints the path to the Kerberos ccache of the given (or current) user to stdout.
+// The function is a no-op if the detect_cached_ticket setting is not enabled.
+// No error is raised if the inferred ticket is not present on disk.
+func (a *App) printTicketPath(username string) (err error) {
```
didrocks, line 325: Do you think we can have an integration test here which is setting some environment/values so that we can mock libkbr5 ticket path and asserting on this command behaviour? It seems it’s not straightfoward but if we can achieve path mocking…

Then, the e2e one would be "only" PAM + CLI ingration.

e2e/scripts/first-run.sh
```diff
@@ -9,3 +9,7 @@ hostnamectl set-hostname "$hostname"
 
 echo "Adding hostname to hosts file..."
 echo "127.0.0.1 $hostname" >> /etc/hosts
+
+echo "Removing ssh configuration overrides..."
```
didrocks, line 13: hum? is it related to that PR? I think the echo isn’t descriptive enough why we are doing this.

e2e/scripts/provision.sh
```diff
@@ -14,7 +14,6 @@ echo "Configure PAM to register user sessions in the systemd control group hiera
 pam-auth-update --enable systemd
 
 echo "Enabling keyboard-interactive authentication for domain users..."
-rm -rf /etc/ssh/sshd_config.d # this contains overrides that conflict with our changes
```
didrocks, line 17: actually, the old comment was not descriptive enough either :p

pam/pam_adsys.c
```diff
@@ -340,7 +427,36 @@ PAM_EXTERN int pam_sm_open_session(pam_handle_t *pamh, int flags, int argc, cons
...
+        }
+
+        retval = pam_putenv(pamh, envvar);
+        krb5ccname = strdup(ticket_path);
+        _pam_drop(envvar);
+        free(ticket_path);
+        if (retval != PAM_SUCCESS) {
+            pam_syslog(pamh, LOG_ERR, "Failed to set KRB5CCNAME to %s", ticket_path);
+            return PAM_SYSTEM_ERR;
+        }
```
didrocks, line 459: We have a new C master now :sunglasses:

didrocks (changes requested):
The C part is perfect. I have a few suggestions and also some others on testing that may not be possible but still worth it IMHO.

This is really great to see, we now have a fallback, while still keeping the existing behaviour without the settings turned on. Well done! :)

pam/pam_adsys.c
didrocks: could be worse, could be C++ :p

internal/ad/krb5_test.go
```diff
@@ -65,6 +66,10 @@ func TestTicketPath(t *testing.T) {
 				require.Error(t, err, "TicketPath should have errored out")
 				return
 			}
+			if tc.wantErrType != nil {
```
didrocks, line 69: Just a nitpick: I think that could be only with wantErr, or maybe you want to if tc.wantErr || tc.wantErrType != " nil, then you check for an error, and readd a wantErrType check nested.

I’m afraid that we set `wantErr: true` and see the test passing (we return an error afterall), but then, don’t check the type because of the previous early return.

didrocks (changes requested):
Glad that I asked for integration tests, way easier than I envision too! Nice work :)

Just one small changes on the latest commits and then, we can merge!

didrocks (approved):
Perfect as usual! Be warned, I’m getting used to it :)

PR #919, opened 2024-02-21
maint: fix last revive occurrence

GabrielNagy:
Introduced after the PR that handled all these.

Fixes UDENG-2287


didrocks (approved):
A self-inflicted regression, right? :)

PR #921, opened 2024-02-22
maint(go): Upgrade to Go 1.22.0

didrocks:
Upgrade module and package to Go 1.22.0
Remove loop closure variable redeclaration

Note: once the end to end work is done to run on PR, we need to ensure we include either a ppa with go 1.22.0 or get it backported already to the targetted distro.

### DIFF
--- a/cmd/adsysd/integration_tests/adsys_test.go
+++ b/cmd/adsysd/integration_tests/adsys_test.go
@@ -121,7 +121,6 @@ socket: %s`, socket)), 0600)
 		"version":                     {args: []string{"version"}},
 	}
 	for name, tc := range tests {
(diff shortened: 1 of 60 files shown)


denisonbarbosa (approved):
I can't even describe how satisfactory this feels! Nicely done!

didrocks:
I think it’s something we will need regularly anyway (some stuff waiting to backport and so on). I propose that for the e2e tests running on older distros, we include a ppa and upload there the needed bits, wdyt?

didrocks:
(the goal to get it merged was to release it before FF and not have this cleanup has a following cleanup as we want to have a clean base for 24.04)

PR #926, opened 2024-02-28
tests: increase gpolist command timeout for AD tests

GabrielNagy:
As the AD object is initialized multiple times in tests but only once in production I opted for the less intuitive approach of defaulting to the longer timeout and overriding with the production one instead of the other way around.

This should fix the ad tests we've seen time out in armhf autopkgtests.


didrocks (approved):
Makes sense. Let’s hope that fixes the biggest hurdles when testing.

PR #927, opened 2024-03-01
tests(e2e): enable e2e tests in GitHub Actions

GabrielNagy:
This PR fixes the remaining issues with e2e tests and enables them to run on pushes to main. Workflow dispatch is also available where developers can trigger ad-hoc tests with custom adsys forks or refs.

Tests on Noble are currently disabled because of issues with the Azure templates. There hasn't been a new template for almost 1 month (strange) so I will come back to this once we have something new to work with.

Passing runs of the workflow on the other codenames can be seen here:
- triggered on push to main: https://github.com/GabrielNagy/adsys/actions/runs/8110510671
- triggered ad-hoc with custom parameters: https://github.com/GabrielNagy/adsys/actions/runs/8110696948
  + test only on Mantic
  + set repository to `GabrielNagy/adsys`
  + set branch to `e2e-tests-ci`

NB: For the `ADSYS_PRO_TOKEN` I'm using my personal Pro token, I don't know if we can get a bespoke token to use specifically for E2E tests.

Fixes UDENG-1536

### DIFF
--- a/e2e/internal/remote/remote.go
+++ b/e2e/internal/remote/remote.go
@@ -147,14 +147,17 @@ func (c Client) Run(ctx context.Context, cmd string) ([]byte, error) {
 		}
 		return nil, fmt.Errorf("command cancelled: %w", ctx.Err())
 	case err := <-waitDone:
+		elapsedTime := time.Since(startTime)
 		wg.Wait() // wait for scanners to finish
-		log.Infof("Command %q finished in %s", cmd, time.Since(startTime).String())
 		mu.Lock()
 		defer mu.Unlock()
+
 		out := []byte(strings.Join(combinedOutput, "\n"))
 		if err != nil {
+			log.Warningf("Command %q failed in %s", cmd, elapsedTime)
 			return out, fmt.Errorf("command failed: %w", err)
 		}
+		log.Infof("Command %q finished in %s", cmd, elapsedTime)
 
 		return out, nil
 	}


denisonbarbosa (approved):
Nothing to say besides nice work!!!!

e2e/cmd/run_tests/01_provision_client/main.go
```diff
@@ -190,7 +190,7 @@ func action(ctx context.Context, cmd *command.Command) error {
 
 	// TODO: remove this once the packages installed below are MIRed and installed by default with adsys
 	log.Infof("Installing universe packages required for some policy managers...")
-	_, err = client.Run(ctx, "DEBIAN_FRONTEND=noninteractive apt-get install -y ubuntu-proxy-manager python3-cepces")
+	_, _ = client.Run(ctx, "DEBIAN_FRONTEND=noninteractive apt-get install -y ubuntu-proxy-manager python3-cepces")
```
didrocks, line 193: hum, shouldn’t we fail quickly if we can’t install?

e2e/scripts/Dockerfile.build
```diff
@@ -1,4 +1,4 @@
-ARG CODENAME=mantic
+ARG CODENAME
```
didrocks, line 1: can’t set the default to "latest" (will be easier in case of a manual run?

e2e/scripts/patches/mantic.patch
```diff
@@ -0,0 +1 @@
+jammy.patch
```
didrocks, line 1: ahah, I won’t be tricked this time that this is a symlink!

didrocks (commented):
I’m just commenting on some considerations we might want to take before giving the final approval! Those are all opened questions :)

didrocks:
(oh and of course, fixing the conflicts in the yaml file :))

e2e/cmd/run_tests/01_provision_client/main.go
didrocks: shouldn’t we log them at least, so that if some policy fails where we expected to have them working, we can see that was because of the package failing to install for $reasons?

didrocks: Ah, that’s perfect then! :)

PR #933, opened 2024-03-05
tests(e2e): add AD cleanup script and run it at deprovision

GabrielNagy:
Just after leaving the domain, upload and run a PowerShell script to the domain controller that does the following:
- recursively delete the OU for the client (this has the side-effect of deleting the users we created for the test)
- delete the GPOs created for the test
- removes stale OUs, GPOs and users, more specifically touching only e2e resources that were created over 7 days ago

I've also made a minor change to the `prepare-ad.ps1` script to prefix GPOs with `e2e` for easier cleanup, and to create test OUs under `OU=e2e` as opposed to the root of the DC. This helps us be more confident in our cleanup process.

Passing test run: https://github.com/ubuntu/adsys/actions/runs/8163777979

Fixes UDENG-1539


didrocks (approved):
So much powershell scripts!

I didn’t see any early return error handling in it, but I must say that I am not native to that language and I know from Edu that this kind of handling is hard. So, I think we can merge it. Good job! :)

PR #934, opened 2024-03-06
docs(e2e): add developer documentation for running e2e tests

GabrielNagy:
I've opted to add this as a `README` inside the `e2e` directory as opposed to something public facing in `docs/`, as this is strictly relevant to developers and doesn't add value if we mirror it to Read the Docs.

I tried to find a compromise between being too terse and overwhelming the reader with details, as this also hurts maintainability of the documentation. However, I think this is a case where we need to be specific since the E2E tests involve many systems interacting and one can easily get lost in the process without guidance. If I've missed anything of note, please let me know.

Fixes UDENG-1538

### DIFF
--- a/e2e/README.md
+++ b/e2e/README.md
@@ -1,0 +1,72 @@
+# End-to-end tests
+
+Due to the nature of the ADSys project, there is only so much validation we can get from unit tests and integration tests. Because these do not interact with a real AD environment, certain scenarios, interactions and edge cases will remain untested.
+
+To bridge this gap we have developed an end-to-end tests pipeline that mimics real-world usage of ADSys as close as possible. To accomplish this we provision disposable Ubuntu VMs on Microsoft Azure which we connect to a real Active Directory domain, exercising ADSys functionality through various triggers like the CLI, systemd or PAM.
+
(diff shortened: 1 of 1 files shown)


e2e/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+The process is automated in GitHub Actions and consists of 2 separate scenarios:
+- building VM templates (preparing base images for the Ubuntu clients)
+- running the tests (asserting ADSys functionality on VMs created from the templates above)
+
+### Building VM templates
+
+Because Canonical doesn't provide Azure Cloud images for Ubuntu Desktop, we try our best to come up with something that closely resembles a desktop experience. This is simply a case of starting from an [Ubuntu Minimal](https://azuremarketplace.microsoft.com/en-us/marketplace/apps/canonical.0001-com-ubuntu-minimal-jammy?tab=overview) image on which we install `ubuntu-desktop` and other packages required for ADSys policy managers to work.
+
+A [provisioning script](https://github.com/ubuntu/adsys/blob/main/e2e/scripts/provision.sh) is then staged and ran on the VM template which enables various PAM modules and prepares the network configuration.
```
denisonbarbosa, line 19:
```suggestion
A [provisioning script](https://github.com/ubuntu/adsys/blob/main/e2e/scripts/provision.sh) is then staged and executed on the VM template which enables various PAM modules and prepares the network configuration.
```

denisonbarbosa (approved):
One very minor nitpick/suggestion. Well done!!

jibel (approved):
I didn't try it but it is well documentation and detailed enough for a developer to get started with the test suite.

PR #941, opened 2024-03-07
tests(e2e): minor e2e test fixups 

GabrielNagy:
Some low-hanging fruit I noticed as we got some activity on the E2E test workflow:

### Deprovisioning can happen from multiple initial states

Our current model only allows for one initial state and one target state. Some commands such as the deprovisioning one, can work incrementally for the client/AD. For example, if we manage to provision the client but fail to provision the AD, the script should still release any client resources. As of now, it only works if the AD was successfully provisioned.

There are multiple ways to handle this, but I've opted for the less intrusive one of making states variadic arguments, where all but the last are considered the "initial states", and the last is the state that is being transitioned to. We still sacrifice some readability but existing code will work, and I've made sure to document the new intricacies of the `WithStateTransition` function.

### Add simple retry mechanism for SSH connections

As seen in some failing CI runs, sometimes we fail to establish SSH connection to either a client or the AD controller. I believe this is a transient failure that could be mitigated if we retry the operation multiple times.

This implements a simple retry mechanism that for now attempts to connect a maximum of 5 times at an interval of 3 seconds.

### Update apt package list before realm join

It seems that realm join installs some packages on its own, but doesn't run `apt update` first, resulting in errors such as:

    * Installing necessary packages: sssd-tools adcli sssd libnss-sss libpam-sss
    Password for localadmin:
    ! E: http://azure.archive.ubuntu.com/ubuntu focal-updates/main amd64 libc-ares2 amd64 1.15.0-1ubuntu0.4 is not (yet) available (404  Not Found [IP: 20.54.144.51 80])

We already run `apt update` just after, before installing adsys, so this is a case of moving the update bit above `realm join`. In addition, upgrade packages to avoid Ubuntu running unattended-upgr on its own and breaking our use of `apt`.

### Move realm join after installing all packages

There's a pesky unattended upgrade that hits us often on focal... let's group all apt commands before joining the realm so there's less chance of it starting between our apt runs.

If it becomes a bigger problem we can stop and disable the unattended upgrade service.

### Add a sleep before checking machine scripts

Once a while we seem to have issues asserting the existence of the startup script. Let's see if we can rule this out with a sleep after reboot.

### Increase max timeout for client reboot

1 minute may not be enough, as per https://github.com/ubuntu/adsys/actions/runs/8195833180/job/22414854516

-------------

Passing CI run: https://github.com/ubuntu/adsys/actions/runs/8201760903

### DIFF
--- a/e2e/internal/remote/remote.go
+++ b/e2e/internal/remote/remote.go
@@ -231,7 +231,7 @@ func (c Client) Upload(localPath string, remotePath string) error {
 // Reboot reboots the remote host and waits for it to come back online, then
 // reestablishes the SSH connection.
 // It first waits for the host to go offline, then returns an error if the host
-// does not come back online within 1 minute.
+// does not come back online within 3 minutes.
 func (c *Client) Reboot() error {
 	log.Infof("Rebooting host %q", c.client.RemoteAddr().String())
 	_, _ = c.Run(context.Background(), "reboot")
@@ -251,7 +251,7 @@ func (c *Client) Reboot() error {
 		return fmt.Errorf("host did not go offline in time")
 	}
 
-	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
+	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Minute)
 	defer cancel()
 
 	// Wait for the host to come back online


didrocks (approved):
Nothing to spot, everything looks great. Thanks for the detailed description that helps on the context!

PR #942, opened 2024-03-08
tests(e2e): collect and upload e2e test logs on failure

GabrielNagy:
In case we get an error during the tests, attempt to collect some potentially relevant logs from the remote client and download them to `logs/$HOSTNAME`. 

We currently collect apport logs, journalctl, and an archive of `/var/log` (excluding `/var/log/journal`). They are uploaded as GH actions artifacts only if tests fail.

Example: https://github.com/GabrielNagy/adsys/actions/runs/8205179603

Fixes UDENG-1537

### DIFF
--- a/e2e/cmd/run_tests/04_test_pro_managers/main.go
+++ b/e2e/cmd/run_tests/04_test_pro_managers/main.go
@@ -70,11 +70,14 @@ func validate(_ context.Context, cmd *command.Command) (err error) {
 	return nil
 }
 
-func action(ctx context.Context, cmd *command.Command) error {
+func action(ctx context.Context, cmd *command.Command) (err error) {
 	rootClient, err := remote.NewClient(cmd.Inventory.IP, "root", sshKey)
 	if err != nil {
 		return fmt.Errorf("failed to connect to VM: %w", err)
 	}
+
+	//nolint:errcheck // This is a best effort to collect logs
+	defer rootClient.CollectLogsOnFailure(ctx, &err, cmd.Inventory.Hostname)
 
 	defer func() {
 		if _, err := rootClient.Run(ctx, "adsysctl policy purge --all -v"); err != nil {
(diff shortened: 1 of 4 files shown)


e2e/cmd/run_tests/04_test_pro_managers/main.go
```diff
@@ -101,6 +101,9 @@ func action(ctx context.Context, cmd *command.Command) error {
 		if _, err := client.Run(ctx, "rm -f /home/*/created-by-adsys-admin-logon-script /home/*/created-by-adsys-admin-logoff-script /home/*/created-by-adsys-user-logon-script"); err != nil {
 			log.Errorf("Teardown: Failed to remove user scripts: %v", err)
 		}
+
+		//nolint:errcheck // This is a best effort to collect logs
+		client.CollectLogs(ctx, cmd.Inventory.Hostname)
```
didrocks, line 106: shouldn’t we follow the same pattern of `defer()` to collect the logs so that if we add more tests/teardown, we always try to do this?

e2e/internal/remote/remote.go
```diff
@@ -272,3 +302,58 @@ func (c *Client) Reboot() error {
...
+		}
+	}
+
+	// Run ubuntu-bug to collect logs
+	_, err = c.Run(ctx, "APPORT_DISABLE_DISTRO_CHECK=1 ubuntu-bug --save=/root/bug adsys")
+	if err != nil {
+		return fmt.Errorf("failed to collect logs: %w", err)
+	}
+	// Save journalctl logs
+	_, err = c.Run(ctx, "journalctl --no-pager --output=short-precise --no-hostname > /root/journal")
```
didrocks, line 337: I suggest that we rather ensure that our apport hook collect the journal then for us as it means we will likely need it for accurate bug reports.

e2e/internal/remote/remote.go
```diff
@@ -272,3 +302,58 @@ func (c *Client) Reboot() error {
...
+	if err != nil {
+		return fmt.Errorf("failed to read logs: %w", err)
+	}
+
+	// Archive and download /var/log
+	if _, err := c.Run(ctx, "tar --exclude=/var/log/journal -czf /root/varlog.tar.gz /var/log"); err != nil {
```
didrocks, line 343: maybe we should collect that too in the apport hook?

didrocks (changes requested):
I have few suggestions and an opened question on this Friday! :)

denisonbarbosa (approved):
Looks good to me (besides what Didier already pointed out)! Nice job!

didrocks (approved):
With the few things already discussed, +1. Well done!

PR #946, opened 2024-03-19
ci: report coverage in Cobertura XML format

GabrielNagy:
The simplest way to report XML coverage in our multi-language project is to separate coverage generation per language, using bespoke tools to generate/convert to XML (we use `python3-coverage` for Python, `gocov` to convert Go coverage to an intermediate JSON format, and `gocov-xml` to convert the JSON to Cobertura XML). Additionally, `reportgenerator` is used to combine the resulting XML files into a single report, which is then uploaded to Codecov.

The way it works under the hood is:
- skip Python XML generation if `reportgenerator`, `gocov`, or `gocov-xml` do not exist in the `PATH`
- if the executables are present, Python coverage is converted to XML and copied to `$PROJECTROOT/coverage`
- as part of the QA action, the Go-only coverage is converted to XML
- the Python and Go XML files are combined to a single report using `reportgenerator`

It's important to note that existing coverage generation is not affected by these changes, thus everything here is additive.

Fixes UDENG-2034

### DIFF
--- a/.github/workflows/qa.yaml
+++ b/.github/workflows/qa.yaml
@@ -86,6 +86,11 @@ jobs:
           grep -hv -e "adsys-gpolist" -e "cert-autoenroll" "${combined_cov_file}" > "${go_only_cov_file}"
           gocov convert "${go_only_cov_file}" | gocov-xml > "${coverage_dir}/coverage.xml"
           reportgenerator -reports:"${coverage_dir}/*.xml" -targetdir:"${cod_cov_dir}" -reporttypes:Cobertura
+      - name: Upload XML coverage report as artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: coverage.xml
+          path: ./coverage/codecov/Cobertura.xml
       - name: Run tests (with race detector)
         run: |
           go test -race ./...


denisonbarbosa (approved):
Amazing work!

.github/workflows/qa.yaml
```diff
@@ -69,9 +74,23 @@ jobs:
...
-          grep -hv -e "testutils" -e "pb.go:" -e "/e2e/" /tmp/coverage.out /tmp/coverage.sudo.out > "${cod_cov_dir}/coverage.combined.out"
+          combined_cov_file="${coverage_dir}/coverage.combined.out"
+          go_only_cov_file="${coverage_dir}/coverage.go-only.out"
+          echo "mode: set" > "${combined_cov_file}"
+          grep -hv -e "testutils" -e "pb.go:" -e "/e2e/" -e "mode: set" /tmp/coverage.out /tmp/coverage.sudo.out >> "${combined_cov_file}"
+
+          # Prepare XML coverage report
+          grep -hv -e "adsys-gpolist" -e "cert-autoenroll" "${combined_cov_file}" > "${go_only_cov_file}"
+          gocov convert "${go_only_cov_file}" | gocov-xml > "${coverage_dir}/coverage.xml"
+          reportgenerator -reports:"${coverage_dir}/*.xml" -targetdir:"${cod_cov_dir}" -reporttypes:Cobertura
```
didrocks, line 88: It’s more manual that I was initially hoping for, but I think this is what it is. We will notice when adding any non go file or files to ignore that the conversion will fail anyway.

didrocks (approved):
Excellent work! Thanks for bumping up the coverage :)

PR #948, opened 2024-03-20
maint: use nolint instead of nosec for gosec

GabrielNagy:
The `nosec` directive [is broken](https://github.com/securego/gosec/issues/1105) as of gosec 2.19 -- we can work around this by using the `nolint` directive from golangci-lint. The main drawback is that this disables the entire linter, not just one rule, but I still kept the rule name for reference.

After this is merged we can bump golangci-lint to ~1.56.2~ 1.57.0.

### DIFF
--- a/.golangci-ci.yaml
+++ b/.golangci-ci.yaml
@@ -17,7 +17,6 @@ linters:
     - godot
     - gofmt
     - gosec
-    - ifshort
     - misspell
     - revive
     - thelper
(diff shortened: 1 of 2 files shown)


denisonbarbosa (approved):
Yeah, I bumped into this in `authd` as well, but it looks like it's only broken for `G30X` though, because we use the `nosec` directive for `exec.Command` and it works :thinking:...
Anyway, nice change!

PR #953, opened 2024-03-26
feat(coverage): ship Corbertura assets as a zip for our partner

didrocks:
They are requiring a zip file instead of the xml directly. Ship both of it so that we still link the xml directly from README.

UDENG-2555


didrocks:
The failing e2e tests are due to the machines not being able to provision. It seems that our automatically updated machine are not available on any releases. It’s something to look with Gabriel and ensure we have docs over this. I think this is skippable for now as we don’t do a real release (and don’t have logical change compared to previous one).

jibel (approved):
👍 THe zip file is properly generated

PR #972, opened 2024-04-24
fix: remove XML declaration from glib schemas

### LINKED ISSUE #955: Update supported policies to ADMX/L and docs failed
Failed to update ADMX/L and docs files for ADSys. Please take a look at the run output in https://github.com/ubuntu/adsys/actions/runs/8475969670.

GabrielNagy:
This is a workaround so [we don't fail](https://github.com/ubuntu/adsys/actions/runs/8809678726/job/24180895123) when parsing schemas with invalid XML declarations, such as [`org.gnome.Extensions.gschema.xml`](https://gitlab.gnome.org/GNOME/gnome-shell/-/blob/41b94ccf11a66c5b315f834061cdec82a622f4ec/subprojects/extensions-app/data/org.gnome.Extensions.gschema.xml). Interestingly this doesn't seem to bother `glib-compile-schemas` but fails with most XML parsers (including Go's parser).

While I've already fixed this [upstream](https://gitlab.gnome.org/GNOME/gnome-shell/-/merge_requests/3281), it'll take a while until the updated version feeds into Ubuntu so we need a workaround within adsys too, otherwise we are blocked from refreshing the ADMX/ADML files.

For simplicity I've opted to remove the XML declaration line if it occurs, this way we don't concern ourselves checking for specific syntax errors.

Fixes #955 / UDENG-2730


didrocks (approved):
Thanks a lot for the separate commits, including the yaml parser update consequence :)

The change and the tests looks good, let’s get that in! Thanks for looking into it!

PR #973, opened 2024-04-25
tests(e2e): enable e2e tests on Noble

GabrielNagy:
Now that we have functional VM image templates for Noble we can stop excluding it from test runs.

Passing run on fork: https://github.com/GabrielNagy/adsys/actions/runs/8832525087

Fixes UDENG-2528


didrocks (approved):
As it’s passing, I won’t dare blocking you running more tests! :D

PR #991, opened 2024-05-10
tests(e2e): add script to destroy old Azure image versions

GabrielNagy:
While checking on our Azure costs I noticed that Gallery costs were slightly increasing over time. This happens because we are charged per VM image version and up until now we had no policy to clean up unused versions.

To mitigate this I've devised a script intended to be run just after the image finishes building (as part of the weekly image build workflow).

The script destroys the oldest N-2 image versions of the specified codename by default, with the number of versions to keep being configurable via CLI argument in case we ever need to run it by hand.

A future commit will integrate this step as part of the scheduled workflow.

Fixes UDENG-2786

-------------

I've tested this script locally and it's currently running on my fork for `mantic`: https://github.com/GabrielNagy/adsys/actions/runs/9031928826/job/24819142947

### DIFF
--- a/.github/workflows/e2e-build-images.yaml
+++ b/.github/workflows/e2e-build-images.yaml
@@ -105,3 +105,6 @@ jobs:
         if: steps.check-vm-template.outputs.image-version != ''
         run: |
           go run ./e2e/cmd/build_base_image/02_create_vm_template
+      - name: Purge old template versions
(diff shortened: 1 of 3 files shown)


didrocks (approved):
I didn’t notice anything particular! Thanks a lot :)

PR #992, opened 2024-05-10
Change coverage publish to nightly prerelease

GabrielNagy:
In short, this saves us from an additional test suite run in our forthcoming TiCS scheduled workflow.

Update the coverage report to be uploaded as a pre-release artifact on every 'main' commit -- the consumer can then simply perform a request like the following to download the file:

https://github.com/ubuntu/adsys/releases/download/nightly/Cobertura.xml

A caveat of this approach is that the tag has to exist, and the ref the tag is pointing to will not change. Unfortunately there's no one size fits all solution for this that:
- handles creating the tag
- keeps the ref up to date
- replaces the coverage asset with the latest

I think these are reasonable drawbacks since tag creation is a one-time action, and we don't (yet) need the tag to point to the latest ref.

Part of UDENG-2756

### DIFF
--- a/.github/workflows/qa.yaml
+++ b/.github/workflows/qa.yaml
@@ -103,16 +103,15 @@ jobs:
           token: ${{ secrets.CODECOV_TOKEN }}
 
       # For quality indicator model partner measure.
-      - name: Compress coverage for release asset
-        if: startsWith(github.ref, 'refs/tags/')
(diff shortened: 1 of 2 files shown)


didrocks:
TBH, I would even not push it and attach as a release artefacts at all in the end as it will be reported during the nightly job to TiCS.

Can we upload as a QA job artefacts so that we can retrieve the one from the last run on "main"?

PR #993, opened 2024-05-10
ci: add TiCS cron job workflow

GabrielNagy:
Because the TiCS suite takes quite a long time to install and run, our best bet is to run it on a nightly basis as a scheduled job. I've selected 4 AM UTC as the start time, this should be a quiet as no one is supposed to be working at this hour on any of our team's timezones (I hope), so no risk of changes coming through while the workflow is running.

To set up TiCS itself, we follow TIOBE's guidelines, namely:
- place the coverage in the expected location
- install TiCS
- compile all of our binaries
- run TiCS
- archive and upload logs as artifacts

Fixes UDENG-2756

-------------

You can see a passing run of this workflow here: https://github.com/GabrielNagy/adsys/actions/runs/9034819300/job/24828572639

### DIFF
--- a/.github/workflows/cron-jobs.yaml
+++ b/.github/workflows/cron-jobs.yaml
@@ -1,0 +1,51 @@
+name: Code quality nightly scan
+
+on:
+  schedule:
+    - cron: '0 4 * * *'
+
+env:
+  apt_dependencies: >-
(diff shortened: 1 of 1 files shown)


didrocks (approved):
Excellent! I’ll ping TIOBE about the new coverage location.

PR #994, opened 2024-05-10
ci: stop publishing coverage as a release artifact

GabrielNagy:
Now that we've decided to rely on workflow artifacts for coverage handling, there's no need to publish these on releases anymore.

I've removed the code responsible for this, together with the README badge.

--------------

Because of the minor change to the artifact extension this has to go in before #993.


didrocks (approved):
Easy! :)

PR #995, opened 2024-05-13
ci: exclude Samba vendor and mock files from TiCS scan

GabrielNagy:
These are either vendored files or mocks used in tests, they should not impact the TiCS coverage report.

Part of UDENG-2759 (still need to handle the Samba auto-update workflow)


internal/policies/certificate/python/vendor_samba/gp/gp_cert_auto_enroll_ext.py
```diff
@@ -1,3 +1,5 @@
+# TiCS: disabled # samba vendored code
```
didrocks, line 1: I would have put generated by the CI job, wdyt?

didrocks (approved):
Thanks a lot!

PR #996, opened 2024-05-13
ci: automate magic comment in Samba patch workflow

GabrielNagy:
We already updated the vendored files in #995, however these are generated by our weekly Samba auto-patch workflow. The simplest way to ensure these comments aren't lost on the next workflow run is to add a patch responsible for adding them.

A manual run of the workflow steps reports no changes to the files in `internal/policies/` which is what we want.

Fixes UDENG-2759


didrocks (approved):
easy rather than hacking the merging tool. Let’s do this!

PR #1007, opened 2024-05-27
tests: cover adsysctl policy shell completion code

GabrielNagy:
This (fairly complex) logic was not previously covered by tests. With a bit of effort we can cover all shell completion code paths and bump the coverage of `adsysctl_policy_test.go` to over 91%.

Fixes UDENG-2763


didrocks (approved):
Excellent work :) Nice to see those covered and the bump in coverage!

PR #1012, opened 2024-06-03
fix(ci): get coverage artifact from last successful job

GabrielNagy:
Previously we would get it from the last _completed_ job which was not necessarily passing so it didn't guarantee that our coverage artifact got uploaded.

To fix this, go for the last successful job instead.

Example failing run: https://github.com/ubuntu/adsys/actions/runs/9344222405/job/25715014478


didrocks (approved):
Thanks for looking at this!

PR #1015, opened 2024-06-03
ci: install staticcheck in TiCS scanner job

GabrielNagy:
The scanner penalizes us because staticcheck is not installed. Let's install it to get some extra points.


didrocks (approved):
Trusting your test, thanks! :)

PR #1029, opened 2024-06-13
ci: switch to team PPA for Azure VPN

GabrielNagy:
I've repackaged `ppp` and `sstp-client` in the team PPA for both Jammy and Noble, so that we are not affected by when the `ubuntu-latest` runner transitions to Noble on GitHub Actions.

Here's a passing run with `ubuntu-latest` (aka 22.04): https://github.com/GabrielNagy/adsys/actions/runs/9502739724
And with `ubuntu-24.04`: https://github.com/GabrielNagy/adsys/actions/runs/9506065559

Fixes UDENG-3048


didrocks (approved):
Thanks for spotting this!

PR #1030, opened 2024-06-14
ci: fix failing TICS check

GabrielNagy:
The TICS job was failing because it couldn't build the PAM module.
Indeed, the required dependency was not installed. To improve upon this a bit, install it and make sure we build the PAM module as part of the sanity checks too.

I've also included some small improvements in the TICS workflow such as archiving and uploading the logs regardless of job status.

Fixes UDENG-3155

Passing TICS run: https://github.com/GabrielNagy/adsys/actions/runs/9513279430/job/26223035380


didrocks (approved):
Ack, that makes total sense why it started to fail! Many thanks!

PR #1032, opened 2024-06-19
tests(e2e): collect and report PAM module coverage in e2e tests

GabrielNagy:
Add two new scripts, one that recompiles the PAM module to add coverage support, and one that collects the coverage at the end of the suite, before deprovisioning. For simplicity we only do these additional steps on a single runner, namely the Noble one. It's not worth it attempting to pool the coverage from all runners since it is most likely identical.

The collected coverage is uploaded as an artifact on the workflow, with a deterministic name, so it can be easily downloaded and reused in the TICS check workflow.

-------------------------

Another way of doing this, with less duplication in the workflows, would have been to download the E2E coverage in the QA workflow and combine it there. This would have given us the updated coverage in Codecov as well, with the side effect of adding more indirection because the coverage "journey" would become:

```
e2e workflow -> QA workflow -> TICS workflow
```
as opposed to

```
e2e workflow | -> TICS workflow
QA workflow  |
```

While it may make sense from a code duplication point of view (i.e. only use `reportgenerator` in the QA workflow and keep downloading a single coverage report in the TICS check), there's no functional argument for coupling the QA and e2e workflows like this, so I went with the more obvious approach.

Fixes UDENG-3049

### DIFF
--- a/.github/workflows/tics-report-daily.yaml
+++ b/.github/workflows/tics-report-daily.yaml
@@ -25,6 +25,7 @@ jobs:
           sudo apt-get update
           sudo DEBIAN_FRONTEND=noninteractive apt-get install -y ${{ env.apt_dependencies }}
           go install honnef.co/go/tools/cmd/staticcheck@latest
+          dotnet tool install -g dotnet-reportgenerator-globaltool
       - name: TiCS scan
         env:
           TICSAUTHTOKEN: ${{ secrets.TICSAUTHTOKEN }}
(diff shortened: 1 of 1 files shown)


didrocks (approved):
I agree with the approach, this is the more obvious one of the two.

I didn’t spot anything special, thanks a lot for working on this!

e2e/cmd/run_tests/03_pam_coverage_support/main.go
```diff
@@ -0,0 +1,72 @@
...
+		return fmt.Errorf("failed to install coverage dependencies: %w", err)
+	}
+
+	// Upload PAM module source code to a persistent location
+	if err := client.Upload(filepath.Join(adsysRootDir, "pam", "pam_adsys.c"), "/root/pam/pam_adsys.c"); err != nil {
+		return fmt.Errorf("failed to upload PAM module source code: %w", err)
+	}
+
+	// Rebuild PAM module in-place with coverage support
+	if _, err := client.Run(ctx, fmt.Sprintf("gcc -fprofile-arcs -ftest-coverage -O0 -shared -Wl,-soname,libpam_adsys.so -o %s/pam_adsys.so pam/pam_adsys.c -lpam", remote.PAMModuleDirectory)); err != nil {
```
3v1n0, line 67: I think you can just use `--coverage`. In authd we've also `-fprofile-abs-path` since it helps in figuring out the source path, but maybe you don't need here.

PR #1034, opened 2024-06-20
tests(e2e): re-enable e2e tests on Oracular

GabrielNagy:
We discussed this yesterday and came to the conclusion that it's better to run the Oracular cell even if an image template doesn't exist yet. It will fail the job, but this way we will notice when a template is finally built (the job will turn green), and no action will be required of the team.


didrocks (approved):
As discussed, thanks again a lot on all this work!

PR #1035, opened 2024-06-21
tests(e2e): filter out releases with no VM images

GabrielNagy:
Rather than have the e2e tests failing constantly until an image becomes available, we can be smarter about it and dynamically exclude the codename until the Azure CLI reports at least one existing image for it.

Of course, this doesn't directly map to the existence of a customized image built by our "E2E - Build image templates" workflow -- but it gives a clue as to when it becomes possible to build such an image.

-----------

This came as a result of the TICS job failing because it couldn't find a passing E2E job from which to download the PAM coverage, and with us enabling Oracular tests none of the jobs would actually be passing anymore (until a Oracular VM template becomes available). This is a band-aid for that, and hopefully for future occurrences as well.

I've tested this locally (oracular is excluded): https://github.com/GabrielNagy/adsys/actions/runs/9611677172

### DIFF
--- a/e2e/scripts/provision.sh
+++ b/e2e/scripts/provision.sh
@@ -25,6 +25,9 @@ if [[ ! "$(lsmod)" =~ cifs ]]; then
     echo "cifs" >> /etc/modules
 fi
 
+echo "Disabling unattended-upgrades to avoid unexpected dpkg frontend locks..."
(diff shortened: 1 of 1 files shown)


didrocks (approved):
Sounds like a reasonable enough approach! Let’s hope next run will be tone one!

PR #1039, opened 2024-07-03
tests(e2e): support new Azure image versioning scheme

GabrielNagy:
The Canonical CPC team recently changed the versioning scheme for Ubuntu Azure VM images from:

```json
{
    "architecture": "x64",
    "offer": "0001-com-ubuntu-minimal-mantic",
    "publisher": "Canonical",
    "sku": "minimal-23_10-gen2",
    "urn": "Canonical:0001-com-ubuntu-minimal-mantic:minimal-23_10-gen2:23.10.202406260",
    "version": "23.10.202406260"
}
```

To something like:

```json
{
    "architecture": "x64",
    "offer": "ubuntu-24_04-lts",
    "publisher": "Canonical",
    "sku": "minimal",
    "urn": "Canonical:ubuntu-24_04-lts:minimal:24.04.202404230",
    "version": "24.04.202404230"
}
```
To make matters worse, previous supported Ubuntu versions (Focal, Jammy, Mantic) still use the old version and there’s no plan to migrate them as well, thus we have to maintain both versioning schemes.

Update the relevant script to differentiate between old and new codenames according to the specification above. We now require an additional `version` CLI parameter since the new versioning scheme does not use the codename at all. This is a hard requirement just for the sake of consistency, even if the previous versioning scheme does not need a version.

The other main difference is that with the new versioning scheme images are Gen2 by default, while Gen1 images are explicitly noted in the SKU.

Fixes UDENG-3379

------------------

The additional commits reverse some codename-specific logic in order to add support for Oracular.
Here's a run on my fork with all cells passing (including Oracular): https://github.com/GabrielNagy/adsys/actions/runs/9766528441

### DIFF
--- a/.github/workflows/e2e-build-images.yaml
+++ b/.github/workflows/e2e-build-images.yaml
@@ -31,35 +31,30 @@ jobs:
         run: |
           set -eu
 
-          all="$(distro-info --supported-esm --fullname)\n$(distro-info --supported --fullname)\n"
-          all="$(printf "$all" | sort -u)"
+          codenames="$(distro-info --supported-esm)\n$(distro-info --supported)\n"
+          versions="$(distro-info --supported-esm --release)\n$(distro-info --supported --release)\n"
+
+          # Paste the codenames and versions together, sort them, and remove duplicates
+          codenames_with_versions="$(paste <(printf "$versions" | cut -d' ' -f1) <(printf "$codenames") | sort -u)"
 
           releases=""
           versions=""
-
-          while IFS= read -r release; do
-              # Extract the version and codename
-              version=$(echo "$release" | awk '{print $2}')
-              r=$(echo "$release" | sed -E 's/.*"([^"]+)".*/\1/' | awk '{print tolower($1)}')
-
-              # Remove the LTS bit from the version if present
-              version=$(echo "$version" | sed 's/ LTS//')
-
+          while IFS=$'\t' read -r version codename; do
               # Filter out unsupported LTS releases
-              if [ "${r}" = "trusty" -o "${r}" = "xenial" -o "${r}" = "bionic" ]; then
+              if [[ "${codename}" =~ trusty|xenial|bionic ]]; then
                   continue
               fi
 
               if [ -n "${releases}" ]; then
                   releases="${releases}, "
               fi
-              releases="${releases}'${r}'"
+              releases="${releases}'${codename}'"
 
               if [ -n "${versions}" ]; then
                   versions="${versions}, "
               fi
-              versions="${versions}\"${r}\": \"${version}\""
-          done <<< "$all"
+              versions="${versions}\"${codename}\": \"${version}\""
+          done <<< "$codenames_with_versions"
 
           echo versions="{${versions}}" >> $GITHUB_OUTPUT
           echo matrix="${releases}" >> $GITHUB_OUTPUT
--- a/.github/workflows/e2e-tests.yaml
+++ b/.github/workflows/e2e-tests.yaml
@@ -48,26 +48,21 @@ jobs:
         run: |
           set -eu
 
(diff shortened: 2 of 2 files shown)


.github/workflows/e2e-build-images.yaml
```diff
@@ -30,12 +31,20 @@ jobs:
...
           releases=""
+          versions=""
+
+          while IFS= read -r release; do
+              # Extract the version and codename
+              version=$(echo "$release" | awk '{print $2}')
+              r=$(echo "$release" | sed -E 's/.*"([^"]+)".*/\1/' | awk '{print tolower($1)}')
+
+              # Remove the LTS bit from the version if present
+              version=$(echo "$version" | sed 's/ LTS//')
```
didrocks, line 46: We unfortunately have similar code in WSL image building :D

.github/workflows/e2e-build-images.yaml
```diff
@@ -30,12 +31,20 @@ jobs:
...
+          all="$(distro-info --supported-esm --fullname)\n$(distro-info --supported --fullname)\n"
+          all="$(printf "$all" | sort -u)"
 
           releases=""
+          versions=""
+
+          while IFS= read -r release; do
+              # Extract the version and codename
+              version=$(echo "$release" | awk '{print $2}')
+              r=$(echo "$release" | sed -E 's/.*"([^"]+)".*/\1/' | awk '{print tolower($1)}')
```
didrocks, line 43: couldn’t we use distroinfo with codename to feed that argument? I know that will force it to run twice to extract version and codename (because it seems options don’t combine :/), but we can assume they are in the same order?

e2e/internal/az/image.go
```diff
@@ -15,6 +16,10 @@ import (
 // NullImageVersion is the version returned when no image version is found.
 const NullImageVersion = "0.0.0"
 
+// oldImageVersionCodenames is a list of Ubuntu codenames for which Azure images
+// are built using the previous image versioning format.
+var oldImageVersionCodenames = []string{"focal", "jammy", "mantic"}
```
didrocks, line 21: nitpick: the new will become old at some port, but as they didn’t version the format… :p

e2e/scripts/Dockerfile.build
```diff
@@ -11,7 +11,7 @@ RUN set -ex \
     && apt-get upgrade -y --no-install-recommends
 
 ARG CODENAME=latest
-RUN if [ "$CODENAME" != "noble" ]; then \
+RUN if echo "focal" | grep -q "$CODENAME"; then \
```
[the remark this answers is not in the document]
didrocks: mind adding a comment for it?

didrocks (changes requested):
Excellent work! but yeah, quite convoluted :/

I only have a couple of small suggestions on it…

.github/workflows/e2e-build-images.yaml
```diff
@@ -31,35 +31,30 @@ jobs:
...
-              version=$(echo "$release" | awk '{print $2}')
-              r=$(echo "$release" | sed -E 's/.*"([^"]+)".*/\1/' | awk '{print tolower($1)}')
-
-              # Remove the LTS bit from the version if present
-              version=$(echo "$version" | sed 's/ LTS//')
-
+          while IFS=$'\t' read -r version codename; do
               # Filter out unsupported LTS releases
-              if [ "${r}" = "trusty" -o "${r}" = "xenial" -o "${r}" = "bionic" ]; then
+              if [[ "${codename}" =~ trusty|xenial|bionic ]]; then
```
didrocks, line 44: assuming bash forever I see :)

didrocks (approved):
Great! This really simplifies it over the long run.

I generally forget about the existence of the paste command too. Too many options :D

Well done!

PR #1050, opened 2024-07-16
maint: remove TICS log uploads

didrocks:
This was leaking the GITHUB_TOKEN which was active for a short period of time, as shown by researcher.
Disabling the upload from their advice.


didrocks (commented):
Tests pass, security issue, merging.

PR #1051, opened 2024-07-18
docs: Change to GB spelling

edibotopic:
This changes the default spelling in the docs from US to GB.

Spelling was fixed in text elements, such as headers and paragraphs, but was not changed when a US-spelling referred to paths, keywords or GUI element that are defined in the code/app.

Where, for example, a path or keyword needed to include a US-spelling, it was surrounded in backticks to exclude it from the spellcheck. These changes were then applied to other similar elements for consistency, even when they did not contain US-spelling.

UDENG-3525

### DIFF
--- a/docs/.sphinx/spellingcheck.yaml
+++ b/docs/.sphinx/spellingcheck.yaml
@@ -2,7 +2,7 @@ matrix:
 - name: rST files
   aspell:
     lang: en
-    d: en_US
+    d: en_GB
   dictionary:
     wordlists:
     - .wordlist.txt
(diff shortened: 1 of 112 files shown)


didrocks (approved):
It's still in draft, but I am assuming it’s ready and the CI pass :)

"enrol*", seriously? I would have never guessed and remember this single `l`!

Appro*u*ving, oh no! :)

edibotopic:
Hey @didrocks --- this is nearly ready. It was in draft on purpose.

I wanted to restore the checks on internal refs because the errors being raised were associated with a US spelling in a dir name, which then had side-effects. I'm restoring the checks because it's the default behaviour and it works fine but it was noisy during editing.

This involve a lot of busy work :smile: 

I will merge once checks pass again. Thanks!!!

PR #1065, opened 2024-08-08
Misc fixes related to Noble updates

denisonbarbosa:
To update the docker image, that we use for the integration tests,  to Noble we had to fix some behaviors that changed in the new release.


debian/control
```diff
@@ -18,6 +18,7 @@ Build-Depends: debhelper-compat (= 13),
                python3-samba,
                samba,
                samba-dsdb-modules,
+               polkitd-pkla,
```
didrocks, line 21: This is in universe, we should either MIR it or downgrade it to suggests until we support the "new" format.

didrocks (commented):
Just one note once the tests pass, do not forget about this. :)

.github/workflows/qa.yaml
```diff
@@ -43,7 +43,7 @@ jobs:
 
   tests:
     name: Tests
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
```
didrocks, line 46: A small, nice TODO to revert to -latest once GH points 24.04 to it? :)

denisonbarbosa: What now that we are using variables? :joy:

didrocks (approved):
\o/

PR #1069, opened 2024-08-14
docs: fix spelling checks in auto-generated material

edibotopic:
The reference section in the docs mostly consists of auto-generated material.
Some of the material is not under our control (dconf) as it comes from upstream packages.
As the main docs use GB spelling and the upstream US spelling, the latter fail our spellcheck.

This PR removes reference material that of type dconf from our automated spellcheck.
It also adds a note to the CONTRIBUTING.md about the autogenerated docs along with
some other minor fixes.

UDENG-3685
UDENG-3686
UDENG-3687


denisonbarbosa (approved):
LGTM! Just a small nitpick/question on the formatting.

CONTRIBUTING.md
```diff
@@ -63,12 +63,24 @@ Once merged to the main branch, `po` files, `README.md` with the command line re
 
 You can also contribute to the documentation. It uses [GitHub Markdown Format](https://docs.github.com/en/github/writing-on-github/getting-started-with-writing-and-formatting-on-github).
 
-You can propose modifications in 2 ways:
+To contribute to the documentation edit the files in the `/docs/` directory after forking and cloning the repo.
+Each page on the documentation [website](https://canonical-adsys.readthedocs-hosted.com/en/stable/) also includes
```
denisonbarbosa, line 67: Small question: wouldn't it be better if we extended the link to match the entirety of `[documentation website]` instead of just `[website]`?

edibotopic: > Small question: wouldn't it be better if we extended the link to match the entirety of `[documentation website]` instead of just `[website]`? 

Agreed! These contributing guidelines will also need some other improvements in a future pulse. Thanks.

docs/.sphinx/spellingcheck.yaml
```diff
@@ -9,7 +9,13 @@ matrix:
...
     output: .sphinx/.wordlist.dic
   sources:
-  - _build/**/*.html
+      # check all files except dconf type in the technical reference
+  - _build/tutorial/**/*.html
+  - _build/explanation/**/*.html
+  - _build/how-to/**/*.html
+  - _build/reference/policies/Computer Policies/Ubuntu/Client management/Computer Scripts/**/*.html|!_build/reference/policies/Computer Policies/Ubuntu/Client management/Power Management/**/*.html
+  - _build/reference/policies/User Policies/Ubuntu/Session management/**/*.html
+  - _build/reference/policies/User Policies/Ubuntu/Peripherals/**/*.html
```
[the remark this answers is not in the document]
edibotopic: Thanks @jibel -- your one-liner was what I originally tested locally to resolve the spelling errors.
I changed it so that specific files in `/policies/` (that we control) could be included; that was something specified in the Jira card and that I originally tried to resolve with regex (not pretty!).

I agree that the one-liner alternative would be easier to maintain and maybe this is a compelling reason to use it.

Could you clarify how/if we want to capture the Scripts, Session and Peripherals files?
If I read you correctly, you are suggesting we just ignore those too for now??
I wouldn't have a problem with that, especially if there is a maintainability argument.
I also don't expect there to be show-stopping spelling errors going from the .go and .yaml files in the near future.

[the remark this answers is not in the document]
jibel: Right, I suggest that we ignore those.
These strings come from upstream (from schemas in `/usr/share/glib-2.0/schemas/`) and if we want to track spelling mistakes it should be tracked at the source. adsys is just a consumer and there isn't much we can do at this level.

CONTRIBUTING.md
edibotopic: fixed in [56da448](https://github.com/ubuntu/adsys/pull/1069/commits/56da44816225a7d595eeb09d0af8a5d9cd7d6521)

jibel (approved):
LGTM. Thank you!

PR #1070, opened 2024-08-14
Fix DCONF policy manager removing userdb on empty policy 

### LINKED ISSUE #1002: Issue: Get error: unable to open named profile (user): using the null configuration.

- [X] I have searched the existing issues and found none that matched mine

### Describe the issue

After installing adsys I get error **unable to open named profile (user@domain): using the null configuration** while logging. I tried to join the pc to the domain using sssd and winbind and in both cases it works the same.

### Steps to reproduce it

Using Ubuntu Desktop 24.04, and Windows Server 2012r2

_No response_

### Environment
* adsys version: please run `adsysctl version`
* Distribution: (**NAME** in `/etc/os-release`)
* Distribution version: (**VERSION_ID** on `/etc/os-release`):

### Log files
Please redact/remove sensitive information:
```raw
adsys service logs can be acquired by running `adsysctl service cat -v`.
You can increase the amount of information displayed by increasing the verbosity level (-v) to -vv or -vvv.
```

### Application settings
Please redact/remove sensitive information:
```raw
Paste the contents of your adsys.yaml file here, if you created one.
```

### Additional information

_No response_

### Double check your logs

- [X] I have redacted any sensitive information from the logs

denisonbarbosa:
We decided to go for this approach to support the `purge` command. Since it's not yet fully functional and this change is breaking user sessions that do not have at least one DCONF-related policy applied, we should revert this change for now and consider a better way of removing the leftover files in the future. 

Closes #1002.

UDENG-3732.


internal/policies/dconf/dconf_test.go
```diff
@@ -57,8 +55,8 @@ func TestApplyPolicy(t *testing.T) {
 		"Update machine disabled key with value": {entries: []entry.Entry{
 			{Key: "com/ubuntu/category/key-s", Value: "'onekey-s-othervalue'", Meta: "s"}},
 			isComputer: true, existingDconfDir: "machine-with-disabled-value"},
-		"Machine empty state": {entries: []entry.Entry{}, isComputer: true, existingDconfDir: "-"},
 
+		"No policy still generates a valid db": {entries: nil},
```
didrocks, line 59: Maybe add a comment now here to explain why we need it, to ensure if we go back that route, we don’t recreate the same issue :)

didrocks (changes requested):
Feeling sorry to see this revert from Gabriel's code, but yeah, necessary, I just have a suggestion on adding some comments in the tests, to ensure this time we don’t drop it without a second thought…

internal/policies/dconf/dconf_test.go
```diff
@@ -57,7 +55,10 @@ func TestApplyPolicy(t *testing.T) {
 		"Update machine disabled key with value": {entries: []entry.Entry{
 			{Key: "com/ubuntu/category/key-s", Value: "'onekey-s-othervalue'", Meta: "s"}},
 			isComputer: true, existingDconfDir: "machine-with-disabled-value"},
-		"Machine empty state": {entries: []entry.Entry{}, isComputer: true, existingDconfDir: "-"},
+
+		// We still need to create the database even if there is no policy, otherwise DCONF will switch to using
```
didrocks, line 59:
```suggestion
		// We still need to create an empty database even if there is no policy, otherwise DCONF will block any writes
```

internal/policies/dconf/dconf_test.go
```diff
@@ -57,7 +55,10 @@ func TestApplyPolicy(t *testing.T) {
...
+		// the default database, which can not be modified by the user.
```
didrocks, line 60:
```suggestion
		// due to missing database profile stack file.
```

didrocks (changes requested):
I adjusted the comments in the proposal to be more exact :)

didrocks (approved):
let’s restore dconf \o/

PR #1080, opened 2024-08-28
Update disable-user-switching.md to make gender neutral.

Lovatoty:
Replaced "his" with "their" in disable-user-switching.md to make the page more gender neutral.


didrocks:
Thanks for your bug report and helping making adsys better!

We are directly importing and auto-refreshing those description from the GNOME desktop schemas projects, which is where the string is.

It means that any update here will be overwritten automatically by next GNOME release. I thus suggest that you send your patch to https://gitlab.gnome.org/GNOME/gsettings-desktop-schemas, more precisely against https://gitlab.gnome.org/GNOME/gsettings-desktop-schemas/-/blob/master/schemas/org.gnome.desktop.lockdown.gschema.xml.in?ref_type=heads#L41.

Thanks a lot! I’m thus closing this.

PR #1081, opened 2024-08-29
Fix username parsing in PAM's slash_to_at_username

denisonbarbosa:
This function is supposed to transform a domain.com\user username to the @ format (i.e. user@domain.com). However, it was leaving the parsed username with bad characters in the end (maybe an off behavior from strncpy or something wrong in our string manipulation). Switching to `strncat` ensures we concatenate only what we want at the end of the string.

UDENG-4314

### DIFF
--- a/pam/pam_adsys.c
+++ b/pam/pam_adsys.c
@@ -264,7 +264,7 @@ static char *slash_to_at_username(const char *username) {
         char *ret = malloc((strlen(username) + 1) * sizeof(char));
         strcpy(ret, backslash + 1);
         strcat(ret, "@");
-        strncpy(ret + strlen(ret), username, backslash - username);
+        strncat(ret, username, backslash - username);
         return ret;
     }
     return strdup(username);


didrocks (changes requested):
isn’t the issue that we just didn’t append `\0` after the second `strncpy`? Like you do with `ret[index] = '\0'`, just determining the right index…

denisonbarbosa:
> isn’t the issue that we just didn’t append `\0` after the second `strncpy`? Like you do with `ret[index] = '\0'`, just determining the right index…

Nope, it wasn't. `strncpy` added more than just what we wanted to the target string, so adding a `\0` wasn't enough. That's why I replaced the function call.

didrocks:
> Nope, it wasn't. strncpy added more than just what we wanted to the target string, so adding a \0 wasn't enough. That's why I replaced the function call.

Maybe we should thus pair, I’m pretty sure it’s doable this way and will be more intelligible

didrocks (approved):
way better!

PR #1085, opened 2024-09-02
Fix version based tests on released version

didrocks:
The release version has a real version number and not dev. Just ensure something is printed in the version string.


denisonbarbosa (approved):
Looks good to me! Thanks for addressing it quickly!

PR #1096, opened 2024-09-11
Add dconf settings for usb drives

falencastro:
Allows for usbguard configuration and mounting removable media as read only.


falencastro:
Blocking thumb drives or mounting them r/o is a requirement from our security team to prevent data exfiltration.

didrocks (approved):
Thanks for proposing this patch! CI pass and the changes and sections looks good. Let’s merge.

Note that you will have the admx/adml generated automatically once merged to main, and you can deploy those to your AD server. You don’t need to wait on any change on the client side for the new dconf keys to be picked up and applied.

Thanks again for the PR and your contribution to adsys.

Issue #1098, opened 2024-09-11
Policy files in all lower case are not applied.


falencastro:
### Is there an existing issue for this?

- [X] I have searched the existing issues and found none that matched mine

### Describe the issue

Our domain controllers sends some GPO's in a file named Registry.pol[1] which is what adsys expects[2], while other policies come as registry.pol[1] which adsys doesn't read[3]. A similar bug[4] was fixed in order to allow interchangeability between User<->USER and Machine<->MACHINE. 

If I create a symlink from registry.pol to Registry.pol the GPO is applied fine.: 

root@deskun:/var/cache/adsys/sysvol/Policies/{1E6D0BAC-10B0-4B4A-8E40-48083AEBE4F6}/User# ls -l
total 12 
lrwxrwxrwx 1 root root 12 Sep 5 19:48 Registry.pol -> registry.pol 
-rw------- 1 root root 552 Sep 5 19:36 comment.cmtx 
-rw------- 1 root root 4960 Sep 5 19:36 registry.pol 

[1] 
root@deskun:/var/cache/adsys/sysvol/Policies# find . -iname Registry.pol 
 ./{1E6D0BAC-10B0-4B4A-8E40-48083AEBE4F6}/User/Registry.pol 
 ./{1E6D0BAC-10B0-4B4A-8E40-48083AEBE4F6}/User/registry.pol 
 ./{DE7D317F-CFDD-4F18-A732-9E45E0B7832C}/Machine/Registry.pol 
 ./{EAE12EC9-EAE8-432F-8CA6-7F7834C377D8}/Machine/Registry.pol 
 ./{8CA9B898-D518-4BCB-9C04-EBEADFA399EF}/Machine/Registry.pol 
 ./{B93FB2AF-0E84-480D-9F79-EB55AA6C2A7F}/Machine/Registry.pol 
 ./{D2722F99-D35C-4034-961C-2706337FB81F}/Machine/Registry.pol 
 ./{63B01851-B60E-45DF-8397-54B6A384673E}/User/registry.pol 
 ./{C51B2AA4-A8F5-4F19-89CD-DE7DBA0F0DDB}/Machine/registry.pol 
 ./{B782125A-E093-49BF-93EC-F24D3E0A17CD}/Machine/Registry.pol 
 ./{C65E81F4-793D-4D84-8DF5-21C74923FF15}/User/Registry.pol 
[...]

PR #1102, opened 2024-09-12
[Fixed] Parse GPO ignoring when registry policy name is not Registry.pol

dangerousplay:
Hello, I hope you are good.

Currently, if the registry policy name file is not exactly `Registry.pol` the AD parse GPO does not find it, ignoring it.
However, there are cases where the registry file name is `registry.pol` for a GPO as reported in issue #1098.
In this pull request the implementation lists the files in the policy class directory and looks for the `registry.pol` file ignoring the case. If the file is found it's parsed, otherwise the policy is ignored.

Solves the issue:
Policy files in all lower case are not applied. #1098

Have a good day ^^


internal/ad/ad.go
```diff
@@ -491,116 +491,155 @@ func (ad *AD) parseGPOs(ctx context.Context, gpos []gpo, objectClass ObjectClass
...
+	var f *os.File
+	for _, class := range classes {
+		var e error
+		var files []os.DirEntry
+
+		policyDir := filepath.Join(ad.sysvolCacheDir, "Policies", filepath.Base(url), class)
+		files, e = os.ReadDir(policyDir)
+
+		if e != nil {
+			if errors.Is(e, fs.ErrNotExist) {
```
didrocks, line 527: I think we should can bypass those nested if / else by only one, and skip directly to the next class if we couldn’t read a directory.
However, if the error is different, it’s better to stop right away.

Something like what we had:
```go
if errors.Is(e, fs.ErrNotExist) {
    log.Debugf(ctx, "Policy directory %q not found", policyDir)
    continue
} else if e != nil {
  return e
}
```

I would drop then the next debug that we found something (we will already print this once try to look for a registry file).

internal/ad/ad.go
```diff
@@ -491,116 +491,155 @@ func (ad *AD) parseGPOs(ctx context.Context, gpos []gpo, objectClass ObjectClass
...
+				log.Debugf(ctx, "Not found policy directory %q", policyDir)
+			} else {
+				log.Warningf(ctx, "Failed to read policy directory %q: %s", policyDir, e)
 			}
+		} else {
+			log.Debugf(ctx, "Success reading policy directory %q: found %d files", policyDir, len(files))
+		}
+
+		foundPolicy := false
```
didrocks, line 536: Maybe put that before the loop?

We are trying to align left, meaning that the happy path (the directory path is found, we list the dir, a filename matching it is found and we can parse it) is on the left, then any error is either a "let’s skip to next directory/potential match" or "an early return".

We still need we ensure we need to find one element then, so foundPolicy can be before the for loop if you see what I mean?

internal/ad/ad.go
```diff
@@ -491,116 +491,155 @@ func (ad *AD) parseGPOs(ctx context.Context, gpos []gpo, objectClass ObjectClass
...
-
-				// We only care about the first error which is caused by opening
-				// the capitalized version of the class, instead of the
-				// uppercase version which is less common and more of an edge case.
-				if e != nil && err == nil {
-					err = e
-				} else if e == nil {
-					err = nil
-					break
+		//The filesystem might be case-sensitive, and if the registry policy is created as registry.pol
```
didrocks, line 538:
```suggestion
		// Registry.pol can have different cases, ensure we can find it whatever its case is
```

internal/ad/ad.go
```diff
@@ -491,116 +491,155 @@ func (ad *AD) parseGPOs(ctx context.Context, gpos []gpo, objectClass ObjectClass
...
+		//it will not be found.
+		for _, file := range files {
+			if strings.EqualFold(file.Name(), "Registry.pol") {
```
didrocks, line 541: We can easily remove one level of nested with:
```
if !strings.EqualFold(file.Name(), "Registry.pol") {
    continue
}
<… the rest of the code :)>
```

internal/ad/ad.go
```diff
@@ -491,116 +491,155 @@ func (ad *AD) parseGPOs(ctx context.Context, gpos []gpo, objectClass ObjectClass
...
+				policyPath := filepath.Join(policyDir, file.Name())
+				log.Debugf(ctx, "Found registry policy file %q", policyPath)
+
+				f, e = os.Open(policyPath)
+
+				if e != nil {
```
didrocks, line 547: See my previous comment on early return on error. Also here, as we can’t have multiple "classes" where we could read the directory, I guess there are 2 kinds of errors:
- the file doesn’t exist -> ok, do next
- any other kind of return -> early return, the policy file has an issue (do not ignore it silently and try to continue).

Something like what we have below:
```go
if errors.Is(err, fs.ErrNotExist) {
	log.Debugf(ctx, "Policy %q doesn't have any policy for class %q %s", name, objectClass, e)
	return nil
} else if e != nil {
	return e
}
```

didrocks (changes requested):
Thanks a lot for this other excellent PR! I’m quite impressed that you were able to dive into the code and get there quite easily, well done!

I have some suggestions that I think will enhance the code from a maintainability perspectice. Some parts of my reviews may be wrong (because the github diff due to the identation changed as you rightly refactor some things into its own function) made it hard to read. I had to rely then on `git diff` directly to be able to have a better sense, ignoring those, but I may still have made some mistakes in my review!

Also, we do have most of the code under tests, and I think this needs a testcase at least in the ad package. Maybe even two (one to parse the a `registry.pol` and one for a mixed case like `regiSTRY.pol`).
I don’t think this necessitate integration tests though as this is more a property to handle some edge cases on `internal/ad`. I would suggest then to only add it here (you can easily run the tests with `go test .` inside that directory).

If the requested changes and/or the tests are too much for you, we can take it from there and finish the work you started, just tell us! Thanks again for your contribution! :)

internal/ad/ad.go
dangerousplay: Yes, It was a bug, changed it to be before the loop.

internal/ad/ad.go
dangerousplay: Suggestion applied.

internal/ad/ad.go
dangerousplay: Simplified the code to fail fast.

dangerousplay:
Thank you for the review.
The code was simplified to fail fast when an error occurs during the class directory list or `Registry.pol` read.

I would appreciate a help to create an automated test for this scenario. I'm trying to figure out how to change the test data to have different cases for the `Registry.pol`.

didrocks (commented):
Excellent! As you can see, I only have 2 very small nitpicks, but the code itself is completely what I was envisionning for. It’s so much more readabale with the early returns, thanks for tackling this!

On the tests, now, as you can see, existing tests still pass, so it means that there is no regression :) 
I think we should add some new cases with a case of `registry.pol` and testing that it’s parsed.

In `internal/ad/ad_test.go`, as you can see, we have multiple tests doing parametric (table) testing. Look at `TestGetPolicies` and see what we have a section for uppercase/lowercase in the class (USER/Machine…): `// Policy class directory spelling cases`

This is the section where you can add new cases (probably changing the comment above), like the following:
```
"Lowercase registry file is parsed":  {
			gpoListArgs: []string{"gpoonly.com", "bob:lowercase-registry"},
			want:        policies.Policies{GPOs: []policies.GPO{standardUserGPO("standard")}},
			},
		},
"Mixed case registry file is parsed":  {
			gpoListArgs: []string{"gpoonly.com", "bob:mixedcase-registry"},
			want:        policies.Policies{GPOs: []policies.GPO{standardUserGPO("standard")}},
			},
		},
```
And the finale missing part, will be to create those 2 `lowercase-registry` and `mixedcase-registry` policies for the `gpoonly.com` domain!

We mock an AD file directory under `internal/ad/testdata/AD/SYSVOL/gpoonly.com/Policies/`. You should copy `standard` to both `lowercase-registry` and `mixedcase-registry`. That way, you have your GPO fixture used in tests with some policies inside.
Now, rename In the first our ` User/Registry.pol` to `User/registry.pol`, and on the second one to something like `User/reGIStry.pol`. I would suggest remove in both the `Machine/` directory itself as it’s of no use.

I know this is a little bit involved, but that should be all what is needed to implement those 2 additional tests and ensuring the code is then covered! How does that sound? If you need any help, I’m happy again to take it over to the finish line, your work and code is already excellent, Thanks again!

internal/ad/ad.go
```diff
@@ -491,116 +491,136 @@ func (ad *AD) parseGPOs(ctx context.Context, gpos []gpo, objectClass ObjectClass
...
+	// We need to consider the uppercase version of the name as well,
+	// which could occur in some of the default GPOs such as Default
+	// Domain Policy.
+	classes := []string{"User", "USER"}
+	if objectClass == ComputerObject {
+		classes = []string{"Machine", "MACHINE"}
+	}
+
+	var f *os.File
+loop:
```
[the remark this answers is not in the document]
dangerousplay: Sure, changed to classLoop

internal/ad/ad.go
```diff
@@ -491,116 +491,136 @@ func (ad *AD) parseGPOs(ctx context.Context, gpos []gpo, objectClass ObjectClass
...
+	for _, class := range classes {
+		var err error
+		var files []os.DirEntry
+
+		policyDir := filepath.Join(ad.sysvolCacheDir, "Policies", filepath.Base(url), class)
+		files, err = os.ReadDir(policyDir)
+
+		if errors.Is(err, fs.ErrNotExist) {
```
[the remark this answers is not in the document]
dangerousplay: Removed the blank line.

internal/ad/ad.go
```diff
@@ -491,116 +491,136 @@ func (ad *AD) parseGPOs(ctx context.Context, gpos []gpo, objectClass ObjectClass
...
 			}
-			defer decorate.LogFuncOnErrorContext(ctx, f.Close)
 
-			// Decode and apply policies in gpo order. First win
-			pols, err := registry.DecodePolicy(f)
+			policyPath := filepath.Join(policyDir, file.Name())
+			log.Debugf(ctx, "Found registry policy file %q", policyPath)
+
+			f, err = os.Open(policyPath)
+
```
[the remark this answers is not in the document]
dangerousplay: Removed the blank line.

dangerousplay:
I highly appreciate your detailed explanation. Thank you for the awesome support ^^

I removed the blank lines and added new tests covering the lower case and mixed case for both computer and user policies.

didrocks (approved):
This looks excellent! Thanks a lot for adding the tests and fixing the remaining comments. The new code is really aligned with the current style and code of adsys, well done to figure this out!

I’m going to merge your excellent contribution. Please note that we are already backporting one version of authd currently to 22.04 and 24.04 LTS. This will be in the next one and we will probably publish this in a ppa in between.

Thanks again, approving and merging now!

Note for myself: for some reasons, even on approved workflow, the environments variables (like the runners we are running the tests against) aren’t populated and this is why I had to run the tests manually. But I’m happy to confirm they all pass (and I did run code quality check manually too).

PR #1103, opened 2024-09-17
Fix minor linting issue

didrocks:
As the CI couldn’t run on forked repo due to env variable capture, fix now this linting issue.


didrocks:
Self merging as it address the CI linting failing.

PR #1147, opened 2024-11-27
Support for new Polkit (versions >= 124)

denisonbarbosa:
The new Polkit introduced quite some changes, even syntax ones. Since then, users had to install an additional compatibility package to make the privilege escalation policy work.

This PR adds support for the new Polkit versions whilst still supporting clients that have the older versions and properly handling migration between versions as well.

UDENG-5318


internal/policies/privilege/privilege.go
```diff
@@ -243,10 +299,94 @@ func splitAndNormalizeUsersAndGroups(ctx context.Context, v string) []string {
...
+	if oldPolkit {
+		return polkitAdminIdentitiesFromConf(ctx, policyKitDir)
+	}
+	return polkitAdminIdentitiesFromRules(ctx, []string{
+		policyKitDir,
+		m.policyKitSystemDir,
+	})
+}
+
+func polkitAdminIdentitiesFromRules(ctx context.Context, rulesDirPaths []string) (adminIdentities string, err error) {
```
didrocks, line 312: No comment either :) Also, it can be beneficial to explain the logic: we take the first file in ascii order matching blablabla, independent of them being in /etc or /usr.

internal/policies/privilege/privilege.go
```diff
@@ -279,3 +419,36 @@ func getSystemPolkitAdminIdentities(ctx context.Context, policyKitDir string) (a
...
+
+// isOldPolkit checks current polkit-1 configuration to determine if the current version < 124.
+//
+// To determine the version, we follow the steps:
+//  1. If the old configuration directory does not exist or is empty -> version < 124.
+//  2. If the old configuration directory only contains the adsys generated file -> version < 124.
+//
+// If the previous checks are valid, we still need to check if the new configuration file exists as the user
+// could have installed the compatibility package (polkitd-pkla), which adds old configuration files even if
+// the polkit version is >= 124.
```
didrocks, line 431: This comment is great!

internal/policies/privilege/privilege.go
```diff
@@ -279,3 +419,36 @@ func getSystemPolkitAdminIdentities(ctx context.Context, policyKitDir string) (a
...
+func isOldPolkit(policyKitDir, policyKitReservedDir string) bool {
+	dirEntries, err := os.ReadDir(filepath.Join(policyKitDir, "localauthority.conf.d"))
+	nEntries := len(dirEntries)
+	if err != nil || nEntries == 0 {
+		return false
+	}
+
+	if nEntries == 1 && dirEntries[0].Name() == adsysOldPolkitName+".conf" {
```
didrocks, line 439: maybe add some comment "empty" (apart from adsys generated file) directory?

internal/policies/privilege/privilege_test.go
```diff
@@ -131,3 +143,30 @@ func TestApplyPolicy(t *testing.T) {
 		})
 	}
 }
+
+// Creates a temporary directory with the default rules file for Polkit.
+//
+// This is needed to avoid the huge absolute paths that come from using t.TempDir() and can
+// eventually go over the maximum limit for paths.
+func createPolkitReservedPath(t *testing.T) string {
```
didrocks, line 151: Shouldn’t those directories and files be shipped as part of the test fixtures in some directories?

internal/policies/privilege/privilege.go
```diff
@@ -279,3 +419,36 @@ func getSystemPolkitAdminIdentities(ctx context.Context, policyKitDir string) (a
...
+		return false
+	}
+
+	if _, err := os.Stat(filepath.Join(policyKitReservedDir, "rules.d/49-ubuntu-admin.rules")); err != nil {
+		return true
+	}
+	return false
+}
+
+const (
```
didrocks, line 449: I would put those consts at the top as they are used in `ApplyPolicy`, no?

didrocks (changes requested):
Great work!

It’s hard to check every single golden files, I did check most of them but I assume you did check all of them one by one :)

I have some nitpicks, but nothing really core to the changes. However, I see that in the policy manager tests `testdata/existing-files` don’t contains pre-existings `rules.d` content which should be ready to be overriden, shouldn’t that be the case or does the github diff is fooling me? I see also that you generate some files that should be as test fixtures IMHO. Finally, some idea on how to improve IMHO the code readability and logic.

But overall this looks really sweet, nice!

internal/policies/privilege/privilege.go
```diff
@@ -43,24 +45,49 @@ import (
...
 	We are modifying 2 files:
 	- one for sudo, named 99-adsys-privilege-enforcement in sudoers.d
-	- one under 99-adsys-privilege-enforcement.conf for policykit
+	- one under 00-adsys-privilege-enforcement.rules for policykit
 
 	Both are installed under respective /etc directories.
 */
 
-const adsysBaseConfName = "99-adsys-privilege-enforcement"
+const adsysBaseSudoersName = "99-adsys-privilege-enforcement"
```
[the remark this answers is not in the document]
denisonbarbosa: I purposedly left it separated since it relates to the sudoers file and all of the below refer to Polkit. But I can join them if you think it makes more sense.

internal/policies/privilege/privilege.go
```diff
@@ -243,10 +299,94 @@ func splitAndNormalizeUsersAndGroups(ctx context.Context, v string) []string {
...
+			rules := string(b)
+
+			// Check if the file contains the rule we are looking for
+			if !strings.Contains(rules, "polkit.addAdminRule") {
+				continue
+			}
+
+			// Extract the return value of the polkit.addAdminRule function
+			// Matches: polkit.addAdminRule(function(action, subject){(.*)});
+			adminRulesRegex, err := regexp.Compile(`polkit\.addAdminRule\s*\(\s*function\s*\(\s*action\s*\,\s*subject\s*\)\s*{\s*[^\}]*}\s*\)\s*\;`)
```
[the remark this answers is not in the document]
denisonbarbosa: Yeah, I realized they didn't need to be inside the loop, but then I had already requested for review so I decided not to touch it (maybe should have commented it before 😅)

internal/policies/privilege/privilege.go
denisonbarbosa: They are, but these template-like strings are kind of "ugly" to look at, so I didn't want to throw them at someone's face for no reason 🤣

internal/policies/privilege/privilege.go
```diff
@@ -243,10 +299,94 @@ func splitAndNormalizeUsersAndGroups(ctx context.Context, v string) []string {
...
+	for _, rulesDirPath := range rulesDirPaths {
+		ruleFiles, err := filepath.Glob(filepath.Join(rulesDirPath, "rules.d", "*.rules"))
+		if err != nil {
+			return "", err
+		}
+
+		slices.Sort(ruleFiles)
+	ruleFilesRange:
+		for _, path := range ruleFiles {
+			if filepath.Base(path) >= currentRulesFile {
```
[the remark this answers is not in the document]
denisonbarbosa: It will be dropped by the new logic, which I like better as well.

internal/policies/privilege/privilege.go
```diff
@@ -243,10 +299,94 @@ func splitAndNormalizeUsersAndGroups(ctx context.Context, v string) []string {
...
+	}
+	return polkitAdminIdentitiesFromRules(ctx, []string{
+		policyKitDir,
+		m.policyKitSystemDir,
+	})
+}
+
+func polkitAdminIdentitiesFromRules(ctx context.Context, rulesDirPaths []string) (adminIdentities string, err error) {
+	currentRulesFile := "99"
+	for _, rulesDirPath := range rulesDirPaths {
```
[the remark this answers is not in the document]
denisonbarbosa: Makes sense, I'll refactor it.

internal/policies/privilege/privilege_test.go
didrocks: I think maybe thus we should fix the test to not call `t.Tempdir` but using something like you do `os.MkdirTemp()` with a shorter path and commenting that this is to prevent overflowing there? (In your comment, I didn’t realize at first the fix was that you are doing `os.MkdirTemp` instead of it).

That way, we still have all fixtures at the same place.

internal/policies/privilege/privilege.go
```diff
@@ -77,12 +104,19 @@ func (m *Manager) ApplyPolicy(ctx context.Context, objectName string, isComputer
...
 	policyKitDir := m.policyKitDir
 	if policyKitDir == "" {
 		policyKitDir = consts.DefaultPolicyKitDir
 	}
-	sudoersConf := filepath.Join(sudoersDir, adsysBaseConfName)
-	policyKitConf := filepath.Join(policyKitDir, "localauthority.conf.d", adsysBaseConfName+".conf")
+	policyKitConf := filepath.Join(policyKitDir, "rules.d", adsysBasePolkitName+".rules")
+
+	// Polkit versions before 124 use a different directory for admin configuration and a different file extension and syntax
+	var oldPolkit bool
```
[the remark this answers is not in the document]
didrocks: I’m ok to give it a pass, I just think it’s a mistake in the long term and some people reading the code will mock us in the future :) I’ll point them at this comment on that PR! :p

internal/policies/privilege/privilege.go
```diff
@@ -243,10 +307,113 @@ func splitAndNormalizeUsersAndGroups(ctx context.Context, v string) []string {
...
+			pathErr := &os.PathError{}
+			if errors.As(err, &pathErr) && pathErr.Op == "open" {
+				// This means that we couldn't open the file for reading, likely due to permission errors.
+				// If so, we can not ensure that we will match the expected admin identities from the system
+				// and we should return an error.
+				return "", err
+			}
+			// If we get an error when reading the file, it's likely due to it being a directory.
+			// This case we can ignore and continue to the next file.
+			log.Debugf(ctx, gotext.Get("Ignoring %s: %v", path, err))
```
didrocks, line 389:
```suggestion
			log.Debugf(ctx, gotext.Get("Ignoring %s: %v"), path, err)
```

internal/policies/privilege/privilege.go
```diff
@@ -243,10 +307,113 @@ func splitAndNormalizeUsersAndGroups(ctx context.Context, v string) []string {
...
+			continue
+		}
+
+		for _, adminRule := range adminRulesRegex.FindAllString(rules, -1) {
+			returnStmt := returnRegex.FindString(adminRule)
+			if returnStmt == "" {
+				continue
+			}
+
+			log.Debugf(ctx, gotext.Get("Using polkit admin identities from %q", path))
```
didrocks, line 405:
```suggestion
			log.Debugf(ctx, gotext.Get("Using polkit admin identities from %q"), path)
```

internal/policies/privilege/privilege.go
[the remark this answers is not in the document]
didrocks: This new logic looks way better and more left-aligned :)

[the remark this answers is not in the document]
denisonbarbosa: I definitely agree!

internal/policies/privilege/privilege_test.go
denisonbarbosa: Hmm... :thinking:
It can be done, sure, but I didn't want to add too much to the test itself and we only have a single test case that requires this directory to be different. Also, the fixtures are made to resemble the structure in `/etc` and these assets don't belong there. If we want the file in `testdata/`, I think we should also switch the structure to resemble `/` instead. But is it worth it?

didrocks: I think it does, having things coming "from 2 places" looks weird to me, no?

denisonbarbosa: It depends... This is more of a system state than a fixture (which, to me, implies it's something we can control in some way), so I think it's fine to have it set in a different way/place. But I can update the assets to match the `/` structure we need if you think it's beneficial.

didrocks: `/etc` is a state, but as we are talking about `/usr` files here, we are more on mimicking the OS FS. I think it’s beneficial to take the time to update the structure so that we ship (as in other tests in adsys) in a single place the current emulated machine FS.

denisonbarbosa: Ack. Will do!

didrocks (approved):
We do have everything we need now. Well done!

PR #1151, opened 2024-11-29
Run certificate autoenroll with debug enabled

denisonbarbosa:
When something went wrong with some parts of autoenrollment, the user would need to download the autoenroll script and then run it manually, instead of getting the expected output when running update policies in debug mode. To avoid this cumbersome process, we now always run the script in debug mode and log the error message.


didrocks (approved):
As we discussed, that makes total sense!

PR #1152, opened 2024-12-04
Sync Go versions declared in the project

denisonbarbosa:
We have a mismatch between the versions listed in the project root and the tools/ directory. Let's keep it synced.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/ubuntu/adsys
 
-go 1.22.0
+go 1.22.2
 
 toolchain go1.22.7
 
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -1,8 +1,6 @@
 module github.com/ubuntu/adsys
 
-go 1.22.0
-
-toolchain go1.22.7
+go 1.22.7
```
didrocks, line 3: Do we really rely on 1.22.7 from a feature or security vulnerability perspective?

denisonbarbosa: As discussed in call: Although we do rely on 1.22.7 to bump some dependencies, the current available Go version in Jammy is still 1.22.2, so we can't go (see what I did here? :joy:) beyond that.

didrocks (approved):
Thanks!

PR #1175, opened 2025-01-15
Bump Go to 1.23

denisonbarbosa:
We now rely on Go >= 1.22.7, but Jammy and Noble versions (for the `golang-1.22-go` package) only go up to 1.22.2, which means we would be unable to build for those distributions. Since those versions now also have `golang-1.23-go` available, let's use that package instead (this part of the change will only be noticeable on the SRUs) and bump the version here to ensure we are covered by our CI.


didrocks (changes requested):
debian/changelog still refer to                golang-go (>= 2:1.22~). Shouldn’t it be updated too?

didrocks (approved):
LGTM now! :)

PR #1177, opened 2025-01-22
docs: ADSys arch explanation

edibotopic:
Adds an architecture explanation for ADSys to the documentation, including a discussion of SSSD's role.

Mermaid is used for rendering the diagrams, which required an extension.

Other minor changes:

- Slightly rewritten landing page for Explanations
- Formatting, spelling

UDENG-5808

### DIFF
--- a/docs/how-to/set-up-ad.md
+++ b/docs/how-to/set-up-ad.md
@@ -10,7 +10,7 @@ As a rule of thumb, we generally separate the Ubuntu configuration from the Wind
 
 ## Ubuntu administrative template generations
 
-**ADSys** ships with pre-built Active Directory administrative templates that you can install on your Active Directory server. You will find two flavors of them:
+**ADSys** ships with pre-built Active Directory administrative templates that you can install on your Active Directory server. You will find two flavours of them:
 
 * One listing only Long Term Supported (LTS) Ubuntu versions.
 * One listing all currently supported Ubuntu versions, including non LTS.
--- a/docs/custom_conf.py
+++ b/docs/custom_conf.py
@@ -185,9 +185,7 @@ custom_extensions = [
 custom_required_modules = ["sphinxcontrib-mermaid"]
 
 # Add files or directories that should be excluded from processing.
-custom_excludes = [
-    'doc-cheat-sheet*',
-    ]
+custom_excludes = ["doc-cheat-sheet*", "diagrams/readme.md"]
 
 # Add CSS files (located in .sphinx/_static/)
 custom_html_css_files = []
--- a/docs/custom_conf.py
+++ b/docs/custom_conf.py
@@ -173,6 +173,7 @@ custom_extensions = [
     'canonical.terminal-output',
     'notfound.extension'
     ]
+    "sphinxcontrib.mermaid",
 
 # Add custom required Python modules that must be added to the
 # .sphinx/requirements.txt file.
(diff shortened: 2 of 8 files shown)


docs/how-to/set-up-ad.md
```diff
@@ -10,7 +10,7 @@ As a rule of thumb, we generally separate the Ubuntu configuration from the Wind
 
 ## Ubuntu administrative template generations
 
-**ADSys** ships with pre-built Active Directory administrative templates that you can install on your Active Directory server. You will find two flavors of them:
+**ADSys** ships with pre-built Active Directory administrative templates that you can install on your Active Directory server. You will find two flavours of them:
```
didrocks, line 13: isn’t that en_UK and we wanted the doc to be en_US or was it the opposite?

didrocks (approved):
Thanks for the detailed, separated, commits!

I only have one question, nothing on the rest. Ofc, I couldn’t check the mermaid diagrams as there is no rendering here, but we can always iterate over them.

docs/how-to/set-up-ad.md
edibotopic: Unfortunately, the Mermaid diagrams are not generated as image files in the repo, and the syntax for including them is not compatible with the native GitHub renderer -- annoying! Another way of doing this would be to run the mermaid CLI in an action.

Indeed. We can iterate on this as needed. We should also fix the RTD preview on PR for ADSys though, which would help for things like this too.

Lastly, the spelling is whatever it is currently set to. It needs to change to US but I will do that separately (eventually).

edibotopic: image

edibotopic: image

didrocks: Thanks for the diagrams! Just a note that stroke me when reading it: on the first diagram, you mention "Active Directory Client" (which I find hard to comprehend), while on the second one, it’s simply "Ubuntu machine" which is way more concrete. Any reasons to use different terms?

edibotopic: Indeed! I guess the client is always an Ubuntu machine, so we could change it to the more concrete version. I prefer the more concrete option if it's sufficiently accurate.

didrocks: Definitively!

denisonbarbosa: @edibotopic, maybe it's also a good idea to use the same term for fetching policies? I see that, on the first diagram, we use "Fetch GPOs" but then we switch to "relevant policies applied". Maybe it's better to narrow it down to a single definition?

denisonbarbosa (changes requested):
Looking good! Just a small nitpick...

PS.: Sorry, I added the nitpick on [this conversation](https://github.com/ubuntu/adsys/pull/1177/files#r1925349798)

edibotopic:
@denisonbarbosa -- just to clarify, are you suggesting replacing all instances of "policies" with "GPOs"? For example, we would have: "fetch GPOs" and "relevant GPOs applied". I would agree with that change.

denisonbarbosa:
> @denisonbarbosa -- just to clarify, are you suggesting replacing all instances of "policies" with "GPOs"? For example, we would have: "fetch GPOs" and "relevant GPOs applied". I would agree with that change.

Exactly. Either that or the other way around (replacing GPOs w/ policies or group policies). Whatever you think is going to be clearer.

edibotopic:
@didrocks -- I addressed your comment in https://github.com/ubuntu/adsys/pull/1177/commits/10458bb84ce4fa6c5ce6a95c7a2a3568db21f797

@denisonbarbosa -- your suggestion was applied in https://github.com/ubuntu/adsys/pull/1177/commits/2388ee43e615443ff995aa47053bf795545b7f7e

denisonbarbosa (approved):
Thanks for the changes. Everything looks great now!

docs/how-to/set-up-ad.md
edibotopic: Marking as resolved -- hope that's OK.

didrocks: ofc!

PR #1183, opened 2025-01-28
Bump Go toolchain version to 1.23.5

denisonbarbosa:
The crypto/x509 package has a vulnerability issue in Go 1.23.1, as can be seen in: https://github.com/ubuntu/adsys/actions/runs/13009232065.

To avoid it, we should bump the toolchain version.

### DIFF
--- a/e2e/scripts/patches/focal.patch
+++ b/e2e/scripts/patches/focal.patch
@@ -20,20 +20,22 @@ index ccf213e0..40a2aa9f 100644
                 dbus,
                 libdbus-1-dev,
 diff --git a/debian/rules b/debian/rules
-index 43646c6a..0708aa3d 100755
+index f9705a3d..3513e53f 100755
 --- a/debian/rules
 +++ b/debian/rules
-@@ -25,6 +25,9 @@ export DH_GOLANG_INSTALL_ALL := 1
+@@ -25,9 +25,8 @@ export DH_GOLANG_INSTALL_ALL := 1
  # Tests needing sudo will be skipped automatically
  export ADSYS_SKIP_INTEGRATION_TESTS=1
 
+-# We want to take whatever ubuntu propose to us (as it won’t download a newer version),
+-# as long as it matches the go.mod go stenza which is the language requirement.
+-export GOTOOLCHAIN := local
 +# Run with Go 1.22
 +export PATH := /usr/lib/go-1.22/bin/:$(PATH)
-+
+
  %:
         dh $@ --buildsystem=golang --with=golang,apport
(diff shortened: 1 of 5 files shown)


e2e/scripts/patches/focal.patch
```diff
@@ -20,20 +20,22 @@ index ccf213e0..40a2aa9f 100644
...
 --- a/debian/rules
 +++ b/debian/rules
-@@ -25,6 +25,9 @@ export DH_GOLANG_INSTALL_ALL := 1
+@@ -25,9 +25,8 @@ export DH_GOLANG_INSTALL_ALL := 1
  # Tests needing sudo will be skipped automatically
  export ADSYS_SKIP_INTEGRATION_TESTS=1
 
+-# We want to take whatever ubuntu propose to us (as it won’t download a newer version),
+-# as long as it matches the go.mod go stenza which is the language requirement.
+-export GOTOOLCHAIN := local
```
didrocks, line 32: you still want to keep GOTOOLCHAIN = local in those diff, or they will be forceably upgraded.

didrocks (changes requested):

e2e/scripts/patches/focal.patch
denisonbarbosa: That was on purpose. If we don't upgrade the Go version, we can't build the package in Focal.

didrocks: Right, but even here, we want to take whatever is available locally, or it will try to bump to align with the go and toolchain stenza in `go.mod`.

adombeck:
Not sure if it's relevant here, but the [analogous PR for authd](https://github.com/ubuntu/authd/pull/762) is blocked because Go 1.23.5 is not in any noble archives. I didn't understand yet why that matters (except that the CI can't build the Debian package, and for some reason we must not install Go from another source).

didrocks:
@adombeck: I think the deb CI build try to mimick launchpad where we don’t have access to the Internet from the builder for obvious security reason.

As long as debian/control set a version requirement (so without the last `.5`) which is in the distro, this is fine with `GOTOOLCHAIN=local` which prevents trying to upgrade it to match the `toolchain` version requested. However, in normal CI (for QA sanity check, build, tests), this variable is not set, and so the go tool will upgrade to match `toolchain`, which is set to the security fixed version.

I’m happy to expand that tomorrow in our standup, it took me some time to find how we should articulate this, while staying compatible with the distro constraints.

didrocks (approved):
Let’s go with this now! Thanks :)

PR #1223, opened 2025-03-18
Fix URL for getting root certificates and some bad code 

denisonbarbosa:
The last fix needed to get everything working on the certificate's autoenrollment end.


didrocks (approved):
Looking good on our vendored code, thanks!

PR #1230, opened 2025-03-21
Increase buffer token size as needed

### LINKED ISSUE #1191: Issue: go bufio.scanner GPO size too large

### Describe the issue

There is a GPO object that is being applied to about ~1000 windows vms, its size is ~476KB in size and we see an error with adsysctl 
after running `adsysctl update -m`
```
ERROR Error from server: error while updating policy: can't get policies for "hostname redacted": one or more error while parsing downloaded elements: /var/cache/adsys/sysvol/Policies/{UUID redacted}/Machine/Registry.pol: can't parse policy: invalid policy: bufio.Scanner: token too long
```

### Steps to reproduce it

In a ubuntu vm use `adsysctl update -m -vvv` on a large GPO object in windows AD.

[report.txt](https://github.com/user-attachments/files/18680024/report.txt)

### Environment
* adsys version: 0.16.0ubuntu1`
* Distribution: plucky
* Distribution version: 25.04

### Log files
Please redact/remove sensitive information:
```raw
adsysctl service cat -vvv
INFO github.com/ubuntu/adsys/internal/config/config.go:79 Init() Using configuration file: /etc/adsys.yaml 
DEBUG Connecting as [[2924:773482]]                
DEBUG github.com/ubuntu/adsys/internal/grpc/logconnections/logconnections.go:28 (*Service).RegisterGRPCServer.StreamServerInterceptor.func3() New request /service/Cat 
DEBUG github.com/ubuntu/adsys/internal/grpc/logconnections/logconnections.go:61 loggedServerStream.RecvMsg() Requesting with parameters:  
DEBUG github.com/ubuntu/adsys/internal/authorizer/authorizer.go:112 Authorizer.IsAllowedFromContext() Check if grpc request peer is authorized 
DEBUG github.com/ubuntu/adsys/internal/authorizer/authorizer.go:203 Authorizer.isAllowed() Polkit call result, authorized: true 

[adsysctl.txt](https://github.com/user-attachments/files/18680013/adsysctl.txt)

```

### Application settings
Please redact/remove sensitive information:
```raw
Paste the contents of your adsys.yaml file here, if you created one.
```

### Additional information

_No response_

### Double check your logs

denisonbarbosa:
Some users had issues with GPOs with large entries overflowing the buffer and causing the update to fail. This fixes that by progressively increasing the expected readable token size until the maximum we can (the `bufio` package limits this).

UDENG-6434
Closes #1191


internal/ad/registry/registry.go
```diff
@@ -51,10 +52,10 @@ type meta struct {
 }
 
 // DecodePolicy parses a policy stream in registry file format and returns a slice of entries.
-func DecodePolicy(r io.Reader) (entries []entry.Entry, err error) {
+func DecodePolicy(f *os.File) (entries []entry.Entry, err error) {
```
adombeck, line 60: > In order to be able to resize the buffer later, we need to have control
> over the file cursor otherwise it would move forward and we would not be
> able to retry reading the large entry.
> 
> Sadly, this implies on removing the fuzz test as we can't use os.File as
> argument.

Could we do the `f.Seek(0, 0)` conditionally if the `r io.Reader` is of type `*os.File`, so that we're able to keep the fuzz test?

adombeck: alternatively, the fuzz test could write the fuzz data to a file (which should be in-memory for performance) and pass that to `DecodePolicy`

denisonbarbosa: I thought about those cases, but I decided against them due to the fact that the fuzz test had just a single test case and it was a bit old. I could be wrong, but it looked like just a POC test so I thought it would be ok to drop it and avoid more complexity.

adombeck: That reasoning is not reflected in the git commit message.

> the fuzz test had just a single test case

It fuzzes the `DecodePolicy` function. IIUC, the policy is user-controlled, so it seems very useful to me to fuzz it to reduce the risk of vulnerabilities.

adombeck: > it looked like just a POC test

It looks like a useful test to me. What makes you think it's just a POC?

adombeck: Ok! Let me know if anything is still unclear, then I can try to explain (I have some expertise with fuzzing).

denisonbarbosa: Ok, I updated it. Let me know what you think!

adombeck: That's creating and writing an actual file in the filesystem, which is a slow operation. When writing fuzz tests, it's important to optimize for efficiency, so that you have a better chance that the fuzz test finds something in a given amount of time. On my system, the change results in a slowdown of factor 2 (which is not that bad, but IMO still worth spending a little bit of time to optimize).

There are two things we can optimize here:
1. Move the creation of the temporary file out of the fuzz target (the function executed by `f.Fuzz`) and instead call `f.Truncate(0)` at the end of the fuzz target.
2. Ensure that the temporary file is not created on a physical disk

For 2., we could use https://pkg.go.dev/github.com/dsnet/golib/memfile to create an in-memory file and change the type of the argument passed to `DecodePolicy` to `io.ReadSeeker`.

internal/ad/registry/registry_test.go
```diff
@@ -427,9 +426,25 @@ func FuzzDecodePolicy(f *testing.F) {
...
-		r := bytes.NewReader(d)
-		_, _ = registry.DecodePolicy(r)
+	// Setup a temporary directory to store the fuzzed files
+	fuzzDir, err := os.MkdirTemp("", f.Name())
+	if err != nil {
+		f.Fail()
+	}
+	f.Cleanup(func() { os.RemoveAll(fuzzDir) })
+
+	f.Fuzz(func(t *testing.T, d []byte) {
```
adombeck, line 436: After running this fuzz test for 1m20s, it found an input which resulted in an error:

```
fuzz: elapsed: 1m20s, execs: 1429611 (0/sec), new interesting: 5 (total: 81)
--- FAIL: FuzzDecodePolicy (79.89s)
    fuzzing process hung or terminated unexpectedly while minimizing: EOF
    Failing input written to testdata/fuzz/FuzzDecodePolicy/44ef800113434b04
    To re-run:
    go test -run=FuzzDecodePolicy/44ef800113434b04
```

[Here](https://github.com/user-attachments/files/19434212/44ef800113434b04.zip) is the failing input which you can use to reproduce and debug the failure. To reproduce, put the input in `internal/ad/registry/testdata/fuzz/FuzzDecodePolicy` and run `go test -fuzz FuzzDecodePolicy ./internal/ad/registry/...`.

internal/ad/registry/registry.go
denisonbarbosa: Done. Nice call!

adombeck (approved):

adombeck:
I've been running the fuzz test on my machine for 30 minutes now and it didn't crash yet, so it seems like your fix worked.

PR #1232, opened 2025-03-25
FuzzDecodePolicy: Add generated corpus

adombeck:
I ran the fuzz test on branch `increase-gpo-buffer` (commit 293c7fe7d23c586486e9f518913d6a7c08b08938) for 38 minutes today.

This commit copies the generated corpus from that run, so that subsequent runs on other machines will make use of the interesting inputs that were found.

A generated corpus can grow quite large, so checking it in with git is not always a good idea, but I think here we're still good with a corpus size of 4 MiB.

Command to copy the generated corpus:

    cp -r "${GOCACHE-$HOME/.cache/go-build}/fuzz/github.com/ubuntu/adsys/internal/ad/registry/FuzzDecodePolicy/" internal/ad/registry/testdata/fuzz/


didrocks (changes requested):
Just to note what was discussed during the standup: let’s try to separate those (either renaming or in a subdirectory with then manual seeding) from the non regression one(s). But that’s a great suggestion!

didrocks (approved):
Excellent! Thanks a lot :)

PR #1242, opened 2025-04-11
Do not fail on commands without exit status

denisonbarbosa:
Some commands are not returning an exit status anymore, so let's not fail on those. If they failed to run properly, it's very likely that other commands down the chain will also fail, so we don't need to worry about it too much.


denisonbarbosa:
This is no longer needed after #1470

PR #1246, opened 2025-04-14
docs: add security overview explanation

edibotopic:
A security overview includes information about what relationship ADSys has to security, what factors influence its secure usage and what security documentation can be found to support its use.

* adds a security overview page
* updates explanation index
* updates wordlist

UDENG-6576


denisonbarbosa (approved):
Precisely as we discussed. Thanks for addressing this! Nice work!

PR #1275, opened 2025-08-01
Allow searching GPO list by userPrincipalName

### LINKED ISSUE #1273: Issue: SAMAccountName assumption

### Describe the issue

Adsys user lookup assumes that a partial UserPrincipalName will equal to SAMaccountName.

For example, if a user has UPN as "user@domain.com" and a SAMaccountName as "u" in Active Directory, the lookup will fail. If SAMaccountName was "user" then adsys would behave correctly.

https://github.com/ubuntu/adsys/blob/fee11c164a8b284fdcdceb4b7575da0dc9b040ee/internal/ad/adsys-gpolist#L236-L238

https://github.com/ubuntu/adsys/blob/fee11c164a8b284fdcdceb4b7575da0dc9b040ee/internal/ad/adsys-gpolist#L86-L94

The failure message is:
```
Searching for account failed with: Failed to find account user
```

### Steps to reproduce it

1. Enroll 24.04 Ubuntu machine into Active Directory and Adsys
2. Create a user in Active Directory with values similar to "user@domain.com" as UPN and "DOMAIN\u" as SAM
3. On Ubuntu host login as "user@domain.com": `sudo login user@domain.com`

Stock Ubuntu 24.04 Desktop image.

Irrelevant.

### Additional information

Collateral GDM behavior:

1. On login screen, use "not listed" button and use user@domain.com to login
2. Authentication with PAM modules will succeed but login to desktop will fail due to adsys behavior described
3. GDM will add user to login screen
4. Using added user will result in correct authentication with PAM and a correct SAMaccountName sent to adsys, successfully logging in user to desktop.

### Double check your logs

drencrom:
Users normally login using their user principal name and also usually the sAMAccountName is the same user removing the @ and the domain. Currently the GPO search is done by removing the @ and the rest from the login name and searching that value as sAMAccountName. This fails for users where their sAMAccountName is not a suffix of their userPrincipalName. This patch adds a search using the userPrincipalName in case the search for sAMAccountName fails to support users with that configuration.

Fixes #1273

### DIFF
--- a/internal/ad/adsys-gpolist
+++ b/internal/ad/adsys-gpolist
@@ -86,9 +86,20 @@ def connectLDAP(url):
 def get_entity(samdb, accountname, objectClass):
     ''' Returns the entity for a given accountname and objectclass '''
 
-    msg = samdb.search(expression='(&(|(samAccountName=%s)(samAccountName=%s$))(objectClass=%s))' %
-                       (ldb.binary_encode(accountname), ldb.binary_encode(accountname), ldb.binary_encode(objectClass)),
-                       attrs=['objectClass', 'objectSid'])
+    if objectClass == ObjectClass.user:
+         samaccountname = accountname.split('@')[0]
+         msg = samdb.search(expression='(&(|(samAccountName=%s)(samAccountName=%s$))(objectClass=%s))' %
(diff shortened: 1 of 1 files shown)


denisonbarbosa (approved):
LGTM. Thanks for spotting the issue and dedicating some time to fixing it. We appreciate it!

PR #1278, opened 2025-08-11
docs: improve presentation of Pro features

edibotopic:
In the ADSys docs, there is currently no one place where the standard and Pro features are summarised, which can make it difficult for users to determine what features are available to them.

This information is summarised in older blog articles, which are susceptible to becoming outdated and that we have less control over maintaining, in comparison with the documentation.

Here, a reference page for "Features" is created, which includes a table of Standard and Pro features.
In addition, a custom Pro block is used to clearly indicate a Pro feature on any page focused on Pro-gated functionality, such as scripts execution.

UDENG-7556


denisonbarbosa (approved):
LGTM!

PR #1288, opened 2025-09-04
Update golangci-lint to v2 and migrate its config

adombeck:
Update golangci-lint to v2, migrate its config and fix issues found by the new linter version.

UDENG-7800


tools/go.mod
```diff
@@ -1,80 +1,92 @@
 module github.com/ubuntu/adsys/tools
 
-go 1.23.0
-
-toolchain go1.23.6
+go 1.24.0
```
denisonbarbosa, line 3: Is the bump to 1.24 mandatory?

adombeck: Yes. Quoting the commit message of adc8d84f55904ba9ade250a8514e8968f1e9fd35:

> golangci-lint v2 requires Go 1.24.0

PR #1299, opened 2025-10-06
docs: add Getting Started ADSys tutorial

edibotopic:
Adds a three-part getting started with ADSys tutorial, for:

1. Setting up Active Directory
2. Setting up and joining an Ubuntu client
3. Applying GPOs with ADSys

Intended to provide an entry point to ADSys and the documentation, which is currently absent.
Each part is relatively self-contained, so readers can skip based on their setup.

UDENG-7175


denisonbarbosa (approved):
Amazing work!

PR #1316, opened 2025-10-28
Fix E2E distro patches

denisonbarbosa:
We now require Go 1.24 to build the package, so this means we can support only until Jammy. This updates the list of supported releases and removes the unnecessary patch and filtering.

### DIFF
--- a/e2e/scripts/patches/jammy.patch
+++ b/e2e/scripts/patches/jammy.patch
@@ -6,8 +6,8 @@ index ccf213e0..2d34a328 100644
  Build-Depends: debhelper-compat (= 13),
                 dh-apport,
                 dh-golang,
--               golang-go (>= 2:1.23~),
-+               golang-1.23-go,
+-               golang-go (>= 2:1.24~),
++               golang-1.24-go,
                 apparmor,
                 dbus,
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e-build-images.yaml
```diff
@@ -42,7 +42,7 @@ jobs:
           versions=""
           while IFS=$'\t' read -r version codename; do
               # Filter out unsupported LTS releases
-              if [[ "${codename}" =~ trusty|xenial|bionic ]]; then
+              if [[ "${codename}" =~ xenial|bionic|focal ]]; then
```
adombeck: Trusty still has "Legacy support" (not sure what exactly that is tbh): https://ubuntu.com/about/release-cycle

Should we keep testing it?

denisonbarbosa: The "legacy support" relates to the 2 years of additional support that we used to provide for some LTS'es before Ubuntu Pro was a thing.

We have stopped releasing adsys for versions older than Jammy for a while now, so it's not worth to test those anymore (and the package won't even build unless we backport golang-1.24 to those releases)

adombeck (approved):

PR #1381, opened 2026-03-13
Properly disable Kerberos usage in AD tests

denisonbarbosa:
We can't use Kerberos in the tests without a proper Kerberos server. However, we were not propagating the actual disablement to the code, which was resulting in Kerberos being enabled on all tests.

UDENG-9401


panlinux:
This is the samba change: https://github.com/samba-team/samba/commit/98ba05559b25be0e0a6e15838a489a7fb0b34813

Related to: https://bugzilla.samba.org/show_bug.cgi?id=15789

adombeck (approved):

Issue #1358, opened 2026-01-12
Issue: Adsys fails to get user policy list when joined to Child Domain but user is member of Universal Group in Parent Domain


fme-scottyno:
### Is there an existing issue for this?

- [x] I have searched the existing issues and found none that matched mine

### Describe the issue

I Built New AD 2019 Domains in a lab (FFL 2016).  td.corp (Forest Root)  and child.td.corp (Child). Both running dns, td.corp is a ca and dhcp server as well.  Created 2 users, child-admin, child-user in the child domain.   The child-admin is member of enterprise admins (located in the td.corp domain) and child-user only has 'domain users' membership in the child domain.   The child-user account works and pulls policies without an issue.   the child-admin@child.td.corp login fails.   Removed the enterprise admins group and it works.   Created a new universal group in parent domain called parentgroup to make sure it wasn't anything specific to enterprise admins group and added the child-admin user to it.... and it fails again.   
also using ldaps but referrals are trying the unencrypted port of the GC during lookups.  both 3268 and 3269 work with telnet.  Both domain servers are GC in the Forest. 

### Steps to reproduce it

1. create Parent Domain
2. create child domain
3. add universal group in the parent domain
4. add user in the child domain with membership to the universal group of parent domain
5. join child domain using realm / sssd using domain admin account of child domain
6. getent passwrd child-user@child.td.corp works
7. groups child-user@child.parent.domain enumerates parent universal group properly
8. sudo login child-admin@child.parent.domain to produce failure.  
9. remove parent universal group from child account and try login again.   No failure.  

### Ubuntu users: System information

Architecture: amd64
CasperMD5CheckResult: pass
CurrentDesktop: ubuntu:GNOME
Date: Sun Jan 11 17:54:16 2026
Dependencies:
 adduser 3.137ubuntu1
 apparmor 4.0.1really4.0.1-0ubuntu0.24.04.5
 apt 2.8.3
 apt-utils 2.8.3
 base-passwd 3.6.3build1
 bind9-dnsutils 1:9.18.39-0ubuntu0.24.04.2
 bind9-host 1:9.18.39-0ubuntu0.24.04.2
 bind9-libs 1:9.18.39-0ubuntu0.24.04.2
 ca-certificates 20240203
 cifs-utils 2:7.0-2ubuntu0.2
 cracklib-runtime 2.9.6-5.1build2
 dbus 1.14.10-4ubuntu4.1
 dbus-bin 1.14.10-4ubuntu4.1
[...]

Issue #1256, opened 2025-05-16
Issue: adsys selects Global Catalog from wrong domain when --objectclass is user in multi-domain environment


ToastKiste21:
### Is there an existing issue for this?

- [x] I have searched the existing issues and found none that matched mine

### Describe the issue

Hello, I'm out of ideas. We have a forest and two domains: one for the users and one for the computers. Each has its own GPOs.
I want to use adsys to apply the GPOs, but when a user tries to log in, the client fetches the client GPOs but not the user GPOs. I get the following error:
` ERROR Error from server: error while updating policy: can't get policies for "user@user.domain": failed to retrieve the list of GPO (exited with 1): exit status 1`
I edited the adsys-gpolist (adsys) to manually set the FQDN for `--objectclass user`:
```
    if args.objectclass == ObjectClass.user: 
       fqdn = '<*RODC*.user.domain>
```
With this, I can retrieve the user GPOs (and also the client GPOs).
From what I understand, adsys gets the Global Catalog (GC) server from sssd. But the issue is: when I run sssctl domain-status computer.domain, I get all of our AD servers, because all of them are GC servers:

```
Online status: Offline

Active servers:
AD Global Catalog: *RODC*.computer.domain
AD Domain Controller: *RODC*.computer.domain

Discovered AD Global Catalog servers:
- *RODC*.computer.domain
- *RODC*.computer.domain
- *RODC*.user.domain
- *RODC*.user.domain
- *dc*.computer.domain
- *dc*.computer.domain
- *dc*.user.domain
- *dc*.user.domain
```

And when I run `sssctl domain-status user.domain`, I get the same servers – with one difference: the AD Domain Controller is from the correct domain (user.domain), but the Global Catalog is still from computer.domain.

```
Online status: Online

Active servers:
AD Domain Controller: *RODC*.user.domain
AD Global Catalog: *RODC*.computer.domain

Discovered AD Global Catalog servers:
- *RODC*.computer.domain
- *RODC*.computer.domain
- *RODC*.user.domain
- *RODC*.user.domain
- *dc*.computer.domain
- *dc*.computer.domain
- *dc*.user.domain
- *dc*.user.domain
```

So adsys is picking the wrong GC when --objectclass is user.

I tried different sssd.conf setups, using only one domain or setup user.domain also, but nothing changed the error – and I'm out of ideas. I need help.
This is my current sssd.conf:

```
[sssd]
services            = nss, pam, pac
domains             = computer.domain
[...]

PR #1421, opened 2026-06-19
AD: Query Global Catalog when fetching GPOs

denisonbarbosa:
In a multi-domain forest, a user can belong to a universal or foreign-domain group defined in another domain (for example, Enterprise Admins in the forest root). Building the security token with samba.auth.user_session() forces Samba to expand that membership over a single domain controller connection; the DC answers with an LDAP referral to the Global Catalog that Samba does not follow, so user_session() raises NTSTATUSError and the script crashes with an unhandled traceback. This left child-domain users unable to fetch their policies (https://github.com/ubuntu/adsys/issues/1358, https://github.com/ubuntu/adsys/issues/1256).

Query the tokenGroups constructed attribute against the Global Catalog instead: its partial replica of the whole forest yields the complete, transitive set of group SIDs without any referral chasing. The token is then assembled locally from those SIDs, removing the dependency on user_session() entirely. The token building is guarded, so any remaining environment-specific failure degrades into a clear GPO_FAILED return code rather than an opaque crash.

UDENG-10734

### DIFF
--- a/internal/ad/ad.go
+++ b/internal/ad/ad.go
@@ -21,6 +21,7 @@ import (
 	"time"
 
 	"github.com/leonelquinteros/gotext"
+	"github.com/sirupsen/logrus"
 	"github.com/ubuntu/adsys/internal/ad/backends"
 	adcommon "github.com/ubuntu/adsys/internal/ad/common"
 	"github.com/ubuntu/adsys/internal/ad/registry"
@@ -266,6 +267,9 @@ func (ad *AD) GetPolicies(ctx context.Context, objectName string, objectClass Ob
 	// Otherwise, try fetching the GPO list from LDAP
 	args := append([]string{}, ad.gpoListCmd...) // Copy gpoListCmd to prevent data race
 	scriptArgs := []string{"--objectclass", string(objectClass), adServerFQDN, objectName}
+	if logrus.GetLevel() >= logrus.DebugLevel {
+		scriptArgs = append(scriptArgs, "--debug")
+	}
 	cmdArgs := append(args, scriptArgs...)
 	cmdCtx, cancel := context.WithTimeout(ctx, ad.gpoListTimeout)
 	defer cancel()
--- a/internal/ad/ad.go
+++ b/internal/ad/ad.go
@@ -48,6 +48,17 @@ const (
 	// policyServerPrefix is the GPO prefix containing keys that configure
 	// policy servers for certificate enrollment.
 	policyServersPrefix string = "Software/Policies/Microsoft/Cryptography/PolicyServers/"
+
+	// The following constants mirror the ReturnCode values returned by the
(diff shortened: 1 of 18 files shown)


internal/ad/ad.go
```diff
@@ -48,6 +48,17 @@ const (
...
+	// The following constants mirror the ReturnCode values returned by the
+	// adsys-gpolist script, so that distinct failures can be reported with
+	// actionable errors instead of an opaque non-zero exit code.
+
+	// gpoListNotFound is returned when the requested account can't be found in AD.
+	gpoListNotFound int = 1
+	// gpoListConnectionFailed is returned when the connection to the AD server failed.
+	gpoListConnectionFailed int = 2
+	// gpoListGPOFailed is returned when the GPO list couldn't be computed for the account.
+	gpoListGPOFailed int = 3
```
adombeck, line 63: Where is the code of the adsys-gpolist script that exits with these exit codes?

denisonbarbosa: It's a Python script. Here: https://github.com/ubuntu/adsys/blob/main/internal/ad/adsys-gpolist

internal/ad/adsys-gpolist
```diff
@@ -161,7 +219,86 @@ def get_token(samdb, dn):
...
+    group_sids = {}
+    dc_failure = None
+    gc_failure = None
+
+    try:
+        dc_group_sids = get_group_sids(samdb, dn)
+        for sid in dc_group_sids:
+            group_sids[str(sid)] = sid
+        if debug:
+            print("DEBUG domain controller tokenGroups (%d): %s" % (len(dc_group_sids), ", ".join(str(s) for s in dc_group_sids)), file=sys.stderr)
```
adombeck: Shouldn't the debug messages have colon after the "DEBUG" level, for consistency with existing warning messages like https://github.com/ubuntu/adsys/blob/6cc9a59e024c6843428fa2d39574c5c77a43d4ff/internal/ad/adsys-gpolist#L449

```suggestion
            print("DEBUG: domain controller tokenGroups (%d): %s" % (len(dc_group_sids), ", ".join(str(s) for s in dc_group_sids)), file=sys.stderr)
```

adombeck: we could also consider extracting  `log_debug` and `log_warning` functions

adombeck (approved):

internal/ad/adsys-gpolist
denisonbarbosa: I don't think we would gain much from it, tbh.

adombeck: ok

denisonbarbosa: Ideally, we would drop this Python script altogether, but we should work on that later

PR #1424, opened 2026-06-23
policies: Allow configuring dynamic values

### LINKED ISSUE #1419: Feature: Support dynamic user variables (e.g. ${USER}) in mount paths
### Is there an existing request for this feature?

### Describe the feature

Currently, when configuring remote storage mounts via adsys and GPO - the paths are static. This makes it difficult to configure generic, per-user network shares (like Active Directory home directories or personal SMB shares) at a global policy level. Administrators have to either create policies per user or use logon scripts to evaluate the environment and mount the appropriate share.

### Describe the ideal solution

It would be nice if `adsys` can support variables inside the mount path configuration. Specifically beneficial would be to have the ability to insert the username of the logged-in user.

For example the `User mounts` policy setting can have something like:
```
smb://hostname/homes/${USER}
```
or
```
[krb5]smb://hostname/homes/${USER}
```

Then a the path on the server would be automatically resolved depending on the user.

### Alternatives and current workarounds

1. Using logon user scripts.
2. Creating separate policies per user.

Ubuntu 24.04 LTS

_No response_

### Additional information

/ 

### Double check your logs

denisonbarbosa:
Currently, all policy values have to be hardcoded in the AD server. This creates a maintenance hassle when trying to set up policies that are the same but have specific targets for each user/machine.

Adding support for specific variables that are well-known or easy for adsys to discover should give users more flexibility when configuring their GPOs.

Closes #1419 
UDENG-10725


internal/policies/dynamicvalues/dynamicvalues.go
```diff
@@ -0,0 +1,110 @@
...
+
+	"github.com/leonelquinteros/gotext"
+)
+
+// Variable names recognized by the engine. Lookup is case-insensitive.
+const (
+	// VarUser is the login name without the domain (sAMAccountName), e.g. "bob".
+	VarUser = "USER"
+	// VarFQDNUser is the fully-qualified user, e.g. "bob@dom.com".
+	VarFQDNUser = "FQDN_USER"
```
adombeck: FQDN is the abbreviation for fully qualified domain name. This, however, is a user name, not a domain name. We should find a better name for the variable.

internal/policies/dynamicvalues/dynamicvalues.go
```diff
@@ -0,0 +1,110 @@
...
+	// VarHostname is the short machine hostname, e.g. "workstation01".
+	VarHostname = "HOSTNAME"
+	// VarFQDNHostname is the fully-qualified machine name, e.g. "workstation01.dom.com".
+	VarFQDNHostname = "FQDN_HOSTNAME"
```
adombeck: Similarly here, I don't think we should call the hostname a "domain name".

internal/policies/dynamicvalues/dynamicvalues.go
adombeck: maybe "FULL_USER" and "FULL_HOSTNAME"

denisonbarbosa: I agree that the abbreviation isn't really precise, but it's something that AD users can quickly wrap their minds around. SSSD uses `fully_qualified_names` in its config for users, so maybe we can go for something like that? i.e. FQN_USER

adombeck: "FQN" is not a common abbreviation. some alternatives:
* `FULLY_QUALIFIED_USER`
* `QUALIFIED_USER`
* `FULL_USER`

adombeck: and ideally whatever we pick should also work for the hostname setting, for consistency

adombeck: or, if the actual username *is* the fully qualified one, then we could make `$USER` expand to that and use something like `$SHORT_USER` for the short username

docs/explanation/dynamic-values.md
```diff
@@ -0,0 +1,72 @@
...
+
+- an unknown or misspelled placeholder is used (for example `${USR}`);
+- a placeholder is malformed (for example `${USER` or `${}`);
+- a user-only placeholder (`${USER}` or `${FQDN_USER}`) is used in a machine policy.
+
+## Examples
+
+- Per-user network share: `smb://server/homes/${USER}`
+- Per-machine system mount: `nfs://server/exports/${HOSTNAME}`
+- Per-user logon script (relative to `SYSVOL/ubuntu/scripts/`): `${USER}/logon.sh`
```
adombeck, line 69: Should we ask Shane for a review of the new docs page?

adombeck (approved):

internal/policies/dynamicvalues/dynamicvalues.go
denisonbarbosa: Also, it looks like [Microsoft](https://support.microsoft.com/en-us/topic/-netbios-domain-name-username-format-cannot-be-used-with-the-kerberos-referral-mechanism-to-log-on-to-a-computer-in-a-cross-forest-environment-8a0486be-4a71-e022-8954-be2e7dfaa4ac) refers to those cases as "FQDN username/domain name"

adombeck: It's not clear to me what "FQDN domain name\username" refers to in that context. The page also mentions the UPN format which is how Microsoft usually calls usernames in an email address format. What are the "FQDN domain name\username" and "NetBIOS domain name\username" formats they mention, and how do they differ from each other and from the UPN format? Do you know?

adombeck: I suspect that with "FQDN domain name\username" they mean something like "workstation01.example.com\denison". In which case the "FQDN" only refers to the domain name, not the username.

docs/explanation/dynamic-values.md
```diff
@@ -0,0 +1,72 @@
...
+    description: "Use dynamic value placeholders such as ${USER} and ${HOSTNAME} in Active Directory policies to template per-user and per-machine configuration."
+---
+
+(exp::dynamic-values)=
+# Dynamic values
+
+```{include} ../pro_content_notice.txt
+    :start-after: <!-- Include start pro -->
+    :end-before: <!-- Include end pro -->
+```
```
edibotopic: If this is a Pro feature we should add it to the feature table in `docs/reference/features/`.

docs/explanation/dynamic-values.md
```diff
@@ -0,0 +1,72 @@
...
+Dynamic values are expanded for every policy manager (GSettings/dconf, privileges,
+scripts, network shares, AppArmor, proxy and certificates). They are expanded only in the
+policy value itself: the contents of files referenced by a policy — such as a script body
+or an AppArmor profile — are **not** considered.
+
+## Error handling
+
+ADSys is strict about placeholders so that mistakes surface immediately rather than
+silently applying a broken value. Policy application fails, blocking the affected
+authentication, when:
```
edibotopic:
```suggestion
ADSys is strict about placeholders and will surface a mistake immediately rather than
silently applying a broken value.

Policy application fails, blocking the affected authentication, when:
```

docs/explanation/dynamic-values.md
```diff
@@ -0,0 +1,72 @@
...
+
+- an unknown or misspelled placeholder is used (for example `${USR}`);
+- a placeholder is malformed (for example `${USER` or `${}`);
+- a user-only placeholder (`${USER}` or `${FULL_USER}`) is used in a machine policy.
```
edibotopic:
```suggestion
- An unknown or misspelled placeholder is used. For example: `${USR}`;
- A placeholder is malformed. For example: `${USER` or `${}`;
- A user-only placeholder, such as `${USER}` or `${FULL_USER}`, is used in a machine policy.
```

Where using enough brackets to describe the syntax so let's avoid adding more if possible.

docs/explanation/dynamic-values.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Examples
```
edibotopic:
```suggestion
## Example uses of dynamic values
```

Just more descriptive.

docs/explanation/dynamic-values.md
```diff
@@ -0,0 +1,72 @@
...
+
+| Placeholder | Expands to | Example | User policies | Machine policies |
+| --- | --- | --- | --- | --- |
+| `${USER}` | Login name without the domain | `bob` | yes | no |
+| `${FULL_USER}` | Fully-qualified user name | `bob@example.com` | yes | no |
+| `${HOSTNAME}` | Short machine hostname | `workstation01` | yes | yes |
+| `${FULL_HOSTNAME}` | Fully-qualified machine name | `workstation01.example.com` | yes | yes |
+| `${DOMAIN}` | Active Directory domain | `example.com` | yes | yes |
+
+Notes:
```
edibotopic:
```suggestion
### Notes on placeholders
```

docs/explanation/dynamic-values.md
```diff
@@ -0,0 +1,72 @@
...
+
+- In a user policy, `${DOMAIN}` is the user's domain, which keeps it correct in
+  multi-domain forests. `${FULL_HOSTNAME}` always uses the machine's domain.
+- Placeholder names are matched case-insensitively, so `${user}` and `${USER}` are
+  equivalent.
```
edibotopic, line 37:
```suggestion
- In a user policy, `${DOMAIN}` is the user's domain, which keeps it correct in
  multi-domain forests. 
- `${FULL_HOSTNAME}` always uses the machine's domain.
- All placeholder names are matched case-insensitively, so `${user}` and `${USER}` are
  equivalent.
```

edibotopic (changes requested):
Minor suggestions only.

docs/explanation/dynamic-values.md
denisonbarbosa: It's only a Pro feature on policies that are already Pro-locked. Does it make a difference?

edibotopic: Oh. In that case, it doesn't make sense to add it to the table.

But, in the new Dynamic Values page, we probably also shouldn't put the Pro notice on top, right? Because that suggests it is a Pro feature for all policies, I would think.

edibotopic (approved):
Signing off on the docs side. Approved.

PR #1444, opened 2026-07-17
Add AGENTS.md instructions

denisonbarbosa:
Document the project-specific architecture, build and generation workflows, test environments, and security constraints so coding agents follow the same conventions as contributors and CI.


AGENTS.md
```diff
@@ -0,0 +1,229 @@
...
+ADSys writes system configuration and runs as a privileged daemon. Treat paths,
+file modes, ownership, symlinks, downloaded assets, command execution, and
+Kerberos material as security boundaries. Reuse existing safe path and
+filesystem helpers, and do not weaken validation to make a test pass.
+
+Policy application is serialized per target object. Preserve the locking and
+cache invariants in `internal/ad/` and `internal/policies/`; avoid holding broad
+locks across network access, parsing, or external commands.
+
+### Documentation and policy definitions
```
edibotopic, line 172: The documentation section could be more comprehensive; see an example from the Ubuntu server docs: https://github.com/canonical/ubuntu-server-documentation/blob/main/AGENTS.md, and cover general style guide, docs structure, and build instructions. Perhaps it could also include ADSys-specific detail, like policy defs and `adsysctl doc`.

But it might be more appropriate to have a dedicated `docs/AGENTS.md` for some of that. What do you think? If so, it could be a separate, future PR.

denisonbarbosa: It makes sense to have a specific one for documentation, especially since it would indeed benefit from a more detailed explanation.

denisonbarbosa: Added a docs/AGENTS.md as part of the same commit; I thought about using a different PR, but I also think it makes sense to add it in this one

docs/AGENTS.md
```diff
@@ -0,0 +1,188 @@
...
+Run documentation commands from the repository root:
+
+```text
+make -C docs install
+make -C docs run
+make -C docs html
+make -C docs lint-md
+make -C docs vale
+make -C docs spelling
+make -C docs woke
```
edibotopic: I think this is deprecated, or should be soon, in favour of vale.

```suggestion
```

edibotopic (approved):
The docs/AGENTS.md looks good to me, @denisonbarbosa .

Seems to capture most of the general docs style points, as well as some ADSys specifics.

We can look at it again later in the cycle, but, for now, this is enough.

PR #1450, opened 2026-08-10
certificate: expose enrolled certificates over gRPC and adsysctl

denisonbarbosa:
Machines enrolled through the native LDAP backend had no way to report what they hold. certmonger answered that question for the CEPCES path via `getcert`, but the native path registers nothing with certmonger, leaving administrators to inspect files by hand.

> **This is part 3 of 4**, split out of #1401.
> `#1448` → `#1449 (engine)` → **this PR** → `packaging + e2e`.
> Targets `split/ldap-enrollment-core`; will be retargeted to `main` once #1449 lands.

## What this contains

Purely the **exposure layer** on top of the engine from #1449 — no enrollment logic changes.

New `adsysctl certificate` command group: `list`, `status`, `renew`, `remove`, `verify` (with `--online` CRL checking), `cas`, `templates` — backed by seven RPCs and a `CertHealth` enum (`healthy`, `due_renewal`, `expired`, `missing`, `key_mismatch`, `unparseable`, `not_yet_valid`), each with its own exit code so scripts can branch on state.

## Notes for reviewers

- **~1,050 of the ~2,850 lines are generated protobuf** (`adsys.pb.go`, `adsys_grpc.pb.go`). Review `adsys.proto` and skim the rest.
- `internal/policies/manager.go` gains only thin pass-throughs to the certificate manager.
- The `templates` RPC holds no global lock, and the requested server is only a selector over CAs already discovered in the domain — it is never dialed, so an arbitrary hostname cannot be turned into an outbound connection.
- `verify --online` only trusts a CRL signed by the certificate's issuer chain and inside its validity window; anything else is reported as indeterminate rather than clean.

UDENG-11276

PR #1451, opened 2026-08-10
debian: default fresh installs to native certificate enrollment

denisonbarbosa:
Final part of the LDAP certificate enrollment work: flip fresh installs to the native backend and make the e2e suite prove which backend it actually tested.

> **This is part 4 of 4**, split out of #1401.
> `#1448` → `#1449 (engine)` → `#1450 (API/CLI)` → **this PR**.
> Targets `split/certificate-management`; will be retargeted to `main` as its parents land.

## What this contains

**Packaging**
- `/etc/adsys.yaml` is written with `certificate_enrollment: ldap` **on first install only**, and removed on purge. Upgrades and reinstalls keep whatever is already there, so no existing machine changes how it enrolls.
- `python3-cepces` moves from `Recommends` to `Suggests` — adsys is in `main` and `python3-cepces` is still in `universe`, so recommending it would pull a universe package into a main package's default install set.
- `ca-certificates` added to `Depends` (the native path installs discovered CA chains).
- Go build dependency raised to the version `go.mod` already declares.
- New packaging smoke autopkgtest covering install / upgrade / remove / purge.

**e2e**
- The certificate assertion previously called `getcert`, so it only ever proved the CEPCES path. It now pins the backend and asserts what that backend should produce: for the native method the health, key match, chain and trust-store presence, plus renewal; for CEPCES the existing certmonger tracking; and the migration between the two.
- The Jammy/Noble backport patches stopped applying after the Go bump, which only surfaced inside a Docker build. Regenerated, plus a test that replays `build-deb.sh`'s patch invocation so packaging drift fails fast and locally.

## Note for reviewers

`/etc/adsys.yaml` is created by `postinst` rather than shipped as a conffile. That is deliberate: it is a first-install default, not a package-owned configuration, so dpkg must not prompt about it or restore it after an administrator removes it. The reasoning is recorded inline in the maintainer script.

UDENG-11277

### DIFF
--- a/conf.example/adsys.yaml
+++ b/conf.example/adsys.yaml
@@ -18,7 +18,8 @@ global_trust_dir: /usr/local/share/ca-certificates
 #ad_backend: sssd
 
 # Certificate enrollment method: ldap (native Go, LDAP/RPC) or cepces (legacy Python/CEPCES).
-# Defaults to cepces for backward compatibility with existing installations.
+# If unset, defaults to cepces for backwards compatibility.
+# New installations default to ldap.
 #certificate_enrollment: cepces
 
 # SSSd configuration
--- a/debian/control
+++ b/debian/control
@@ -32,6 +32,7 @@ Architecture: any
 Built-Using: ${misc:Built-Using},
 Depends: ${shlibs:Depends},
          ${misc:Depends},
+         ca-certificates,
          python3,
          python3-samba,
          samba-dsdb-modules,
@@ -45,8 +46,8 @@ Depends: ${shlibs:Depends},
 Recommends: ${misc:Recommends},
             ubuntu-advantage-desktop-daemon,
 Suggests: curlftpfs,
-          ubuntu-proxy-manager,
           python3-cepces,
+          ubuntu-proxy-manager,
 Description: ${source:Synopsis}
  ${source:Extended-Description}
 
--- a/debian/copyright
+++ b/debian/copyright
@@ -47,6 +47,30 @@ Files: vendor/charm.land/*
 Copyright: 2020-2026 Charmbracelet, Inc.
 License: Expat
 
+Files: vendor/github.com/Azure/*
+Copyright: 2016 Microsoft
+License: Expat
+
+Files: vendor/github.com/Azure/go-ntlmssp/.golangci.yml
+       vendor/github.com/Azure/go-ntlmssp/authenticate_message.go
+       vendor/github.com/Azure/go-ntlmssp/authheader.go
+       vendor/github.com/Azure/go-ntlmssp/avids.go
+       vendor/github.com/Azure/go-ntlmssp/challenge_message.go
+       vendor/github.com/Azure/go-ntlmssp/messageheader.go
+       vendor/github.com/Azure/go-ntlmssp/negotiate_flags.go
+       vendor/github.com/Azure/go-ntlmssp/negotiate_message.go
+       vendor/github.com/Azure/go-ntlmssp/negotiator.go
+       vendor/github.com/Azure/go-ntlmssp/nlmp.go
+       vendor/github.com/Azure/go-ntlmssp/unicode.go
+       vendor/github.com/Azure/go-ntlmssp/varfield.go
+       vendor/github.com/Azure/go-ntlmssp/version.go
+Copyright: Microsoft Corporation.
+License: Expat
+
+Files: vendor/github.com/Azure/go-ntlmssp/internal/*
+Copyright: 2009 The Go Authors.
+License: Expat
+
 Files: vendor/github.com/alecthomas/*
 Copyright: 2017 Alec Thomas
 License: Expat
@@ -130,6 +154,20 @@ Copyright: 2012 The Go Authors.
            fsnotify Authors.
 License: BSD-3-clause
 
+Files: vendor/github.com/geoffgarside/*
+Copyright: 2009 The Go Authors.
+License: BSD-3-clause
+
+Files: vendor/github.com/go-asn1-ber/*
+Copyright: 2011-2015 Michael Mitton () <mmitton@gmail.com>
+           2015-2016 go-asn1-ber Authors
+License: Expat
+
+Files: vendor/github.com/go-ldap/*
+Copyright: 2011-2015 Michael Mitton () <mmitton@gmail.com>
+           2015-2024 go-ldap Authors
+License: Expat
+
 Files: vendor/github.com/go-viper/*
 Copyright: 2013 Mitchell Hashimoto
 License: Expat
@@ -154,8 +192,27 @@ Files: vendor/github.com/gorilla/*
 Copyright: 2012-2023 The Gorilla Authors.
 License: BSD-3-clause
 
+Files: vendor/github.com/hashicorp/*
+Copyright: 2015-2022 HashiCorp, Inc.
+License: UNKNOWN
+
 Files: vendor/github.com/inconshreveable/*
 License: Apache-2.0
+
+Files: vendor/github.com/indece-official/*
+Copyright: 2019 indece
+License: Expat
+
+Files: vendor/github.com/indece-official/go-ebcdic/ebcdic.go
+Copyright: 2019 indece UG (haftungsbeschränkt)
+License: Expat
+
+Files: vendor/github.com/jcmturner/*
+License: Apache-2.0
+
+Files: vendor/github.com/jcmturner/gofork/*
+Copyright: 2009-2012 The Go Authors.
+License: BSD-3-clause
 
 Files: vendor/github.com/kardianos/*
 Copyright: 2015-2019 Daniel Theophanes.
@@ -227,6 +284,29 @@ Files: vendor/github.com/mvo5/*
 Copyright: 2013-2019 Michael Vogt
 License: UNKNOWN
 
+Files: vendor/github.com/oiweiwei/*
+Copyright: 2026 oiweiwei
+License: Expat
+
+Files: vendor/github.com/oiweiwei/go-smb2.fork/*
+Copyright: 2016 Hiroshi Ioka.
+License: BSD-2-clause
+
+Files: vendor/github.com/oiweiwei/go-smb2.fork/all.go
+       vendor/github.com/oiweiwei/go-smb2.fork/filepath.go
+Copyright: 2009 The Go Authors.
+           2016 Hiroshi Ioka.
+License: BSD-3-clause
+
+Files: vendor/github.com/oiweiwei/go-smb2.fork/internal/*
+Copyright: 2009-2016 The Go Authors.
+           2016 Hiroshi Ioka.
+License: BSD-3-clause
+
+Files: vendor/github.com/oiweiwei/gokrb5.fork/*
+Copyright: 2026 oiweiwei
+License: Apache-2.0
+
 Files: vendor/github.com/pelletier/*
 Copyright: 2021-2023 Thomas Pelletier
 License: Expat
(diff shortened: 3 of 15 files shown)


adombeck (commented):
I'll wait until the Copilot comments were addressed before reviewing this. Please request a review again when that's done. Thanks!

e2e/cmd/run_tests/11_test_pro_managers/main.go
```diff
@@ -174,11 +174,70 @@ func action(ctx context.Context, cmd *command.Command) (err error) {
...
-		// Enrollment takes a few seconds, so no better way to do this than an arbitrary sleep :)
-		time.Sleep(5 * time.Second)
-		if err := rootClient.RequireContains(ctx, "getcert list -i warthogs-CA.Machine", "status: MONITORING"); err != nil {
+		// A fresh package installation defaults to the native LDAP enrollment
+		// method. Pin the selection anyway so the assertions do not depend on
+		// whether the base image already carried an adsys configuration.
+		if err := setCertificateEnrollment(ctx, rootClient, "ldap"); err != nil {
 			return err
 		}
+		if _, err := rootClient.Run(ctx, "systemctl restart adsys-gpo-refresh"); err != nil {
```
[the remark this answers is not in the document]
denisonbarbosa: Confirmed and fixed: the refresh unit only runs `adsysctl update --all` against the daemon that is already running, and the daemon builds its certificate manager once at startup. Every backend switch now goes through a helper that restarts `adsysd` before triggering the refresh, so the test actually exercises the selected backend.

e2e/cmd/run_tests/11_test_pro_managers/main.go
```diff
@@ -174,11 +174,70 @@ func action(ctx context.Context, cmd *command.Command) (err error) {
...
+		if err := setCertificateEnrollment(ctx, rootClient, "ldap"); err != nil {
+			return err
+		}
+		if _, err := rootClient.Run(ctx, "systemctl restart adsys-gpo-refresh"); err != nil {
+			return err
+		}
+		if err := requireEventuallyContains(ctx, rootClient, "adsysctl certificate list", "status: healthy"); err != nil {
+			return err
+		}
+		if _, err := rootClient.Run(ctx, "! getcert list -i warthogs-CA.Machine 2>/dev/null | grep -q 'status: MONITORING'"); err != nil {
```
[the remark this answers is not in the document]
denisonbarbosa: Fixed: the assertion now requires the request to be gone entirely rather than merely out of `MONITORING`.

e2e/cmd/run_tests/11_test_pro_managers/main.go
```diff
@@ -276,3 +335,37 @@ ftp_proxy="http://127.0.0.1:8080"`); err != nil {
...
+		case <-time.After(5 * time.Second):
+		}
+	}
+}
+
+// setCertificateEnrollment pins the certificate enrollment method in
+// /etc/adsys.yaml, replacing the key when present and appending it otherwise.
+func setCertificateEnrollment(ctx context.Context, client remote.Client, method string) error {
+	_, err := client.Run(ctx, fmt.Sprintf(
+		"sed -i -e 's/^certificate_enrollment:.*/certificate_enrollment: %s/' /etc/adsys.yaml && (grep -q '^certificate_enrollment:' /etc/adsys.yaml || echo 'certificate_enrollment: %s' >> /etc/adsys.yaml)",
```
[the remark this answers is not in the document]
denisonbarbosa: Fixed: the file is created before `sed` runs, so the append branch is reachable on an image without a configuration file.

denisonbarbosa:
All Copilot comments are addressed in `e2e: switch the enrollment backend the way a machine really does`:

- the backend switch now restarts `adsysd` before triggering the refresh, since `adsys-gpo-refresh` only asks the already-running daemon to update policy and the certificate manager is built once at startup;
- `/etc/adsys.yaml` is created before `sed` edits it, so pinning the backend works on an image without a configuration file;
- the migration assertion requires the certmonger request to be gone entirely rather than just out of `MONITORING`.

It also restores the documentation about new installations defaulting to the native backend, which moved here from #1449 because this is the change that makes it true.

@adombeck ready for your review.

debian/control
```diff
@@ -5,7 +5,7 @@ Maintainer: Ubuntu Developers <ubuntu-devel-discuss@lists.ubuntu.com>
 Build-Depends: debhelper-compat (= 13),
                dh-apport,
                dh-golang,
-               golang-go (>= 2:1.24~),
+               golang-go (>= 2:1.25~),
```
adombeck, line 8: is this related to the change which the PR is about?

debian/copyright
```diff
@@ -225,6 +225,61 @@ Files: vendor/github.com/ubuntu/decorate/*
...
+License: BSD-2
+
+Files: vendor/github.com/oiweiwei/gokrb5.fork/*
+Copyright: 2016 Jonathan Turner (original gokrb5 upstream)
+           oiweiwei (fork maintainer)
+License: Apache-2.0
+
+Files: vendor/github.com/rs/zerolog/*
+Copyright: 2017 Olivier Poitrey
+License: MIT
```
adombeck: How were these copyright entries created / generated? We might want to try using https://github.com/canonical/dep5-vendor-gen for that

e2e/cmd/run_tests/11_test_pro_managers/main.go
```diff
@@ -358,11 +351,34 @@ func requireEventuallyContains(ctx context.Context, client remote.Client, cmd, e
...
+}
+
 // setCertificateEnrollment pins the certificate enrollment method in
 // /etc/adsys.yaml, replacing the key when present and appending it otherwise.
+// The file is created first: an upgraded or base image may not ship one, and
+// sed would fail on the missing file.
 func setCertificateEnrollment(ctx context.Context, client remote.Client, method string) error {
 	_, err := client.Run(ctx, fmt.Sprintf(
-		"sed -i -e 's/^certificate_enrollment:.*/certificate_enrollment: %s/' /etc/adsys.yaml && (grep -q '^certificate_enrollment:' /etc/adsys.yaml || echo 'certificate_enrollment: %s' >> /etc/adsys.yaml)",
+		"touch /etc/adsys.yaml && sed -i -e 's/^certificate_enrollment:.*/certificate_enrollment: %s/' /etc/adsys.yaml && (grep -q '^certificate_enrollment:' /etc/adsys.yaml || echo 'certificate_enrollment: %s' >> /etc/adsys.yaml)",
```
adombeck, line 381: This commit seems to be a fixup of the previous (or previous two?) commits. Maybe squash the changes into the previous commits?

debian/control
denisonbarbosa: Not really. I bumped the Go version a while back and missed the needed update in ` d/control`. I'll open a separate PR to fix this issue.

denisonbarbosa: Here: https://github.com/ubuntu/adsys/pull/1453

debian/copyright
denisonbarbosa: Hmmm, this is a nice one and, apparently, our d/copyright had a lot of issues. I'll open another PR to update the copyright file with the script, and then we can update the one here in the PR to avoid a huge diff.

adombeck (approved):

debian/copyright
```diff
@@ -440,6 +440,109 @@ Files: vendor/gopkg.in/yaml.v3/decode.go
...
+
+Files: internal/policies/certificate/python/vendor_samba/*
+Copyright: 2021 David Mulder <dmulder@suse.com>
+License: GPL-3+
+
```
adombeck: I get a diff when I run `dep5-vendor-gen` on this branch

denisonbarbosa: hmmm, I don't see that diff when running the tool.

adombeck: the diff is a lot smaller now on commit 7112e13cf511335e726eb27ff0034f9cafbd9937 than it was on commit 22bdc48626f307cdad8e3d120456a7f3993ecd95. this is the remaining diff:

```diff
diff --git a/debian/copyright b/debian/copyright
index 250cc42f..13fe9a53 100644
--- a/debian/copyright
+++ b/debian/copyright
@@ -449,7 +449,6 @@ Files: vendor/go.yaml.in/yaml/v3/decode.go
        vendor/go.yaml.in/yaml/v3/resolve.go
        vendor/go.yaml.in/yaml/v3/sorter.go
        vendor/go.yaml.in/yaml/v3/yaml.go
-       vendor/go.yaml.in/yaml/v3/NOTICE
 Copyright: 2011-2019 Canonical Ltd
 License: Apache-2.0
 
@@ -520,7 +519,6 @@ Files: vendor/gopkg.in/yaml.v3/decode.go
        vendor/gopkg.in/yaml.v3/resolve.go
        vendor/gopkg.in/yaml.v3/sorter.go
        vendor/gopkg.in/yaml.v3/yaml.go
-       vendor/gopkg.in/yaml.v3/NOTICE
 Copyright: 2011-2019 Canonical Ltd
 License: Apache-2.0
```

denisonbarbosa: Yeah, this one I explicitly kept. Removing those lines would cause discrepancy with the output of `lrc`, since they would be added to the `Apache-2.0 or Expat` line and those files explicitly have an Apache-2.0 license.

adombeck: Ah, I think lrc is wrong about that. `NOTICE` is a license file, it specifies the license for other files in the repo but not for the file itself. You can tell lrc to ignore it, like we do for authd: https://github.com/canonical/authd/blob/c8848c2a121c07202efb23f243929c1fec1968c1/debian/lrc.config#L23

conf.example/adsys.yaml
```diff
@@ -19,6 +19,7 @@ global_trust_dir: /usr/local/share/ca-certificates
 
 # Certificate enrollment method: ldap (native Go, LDAP/RPC) or cepces (legacy Python/CEPCES).
 # Defaults to cepces for backward compatibility with existing installations.
+# New installations default to ldap via the package postinst.
```
adombeck: Is this a user-facing comment? If so, I would omit the "via package postinst", it's an implementation detail.

```suggestion
# If unset, defaults to cepces for backward compatibility.
# New installations default to ldap.
```

conf.example/adsys.yaml
```diff
@@ -19,6 +19,7 @@ global_trust_dir: /usr/local/share/ca-certificates
...
 #certificate_enrollment: cepces
```
adombeck, line 23: The example configuration file in docs/reference/adsys-daemon.md doesn't include this setting. Should we add it there too?

adombeck (approved):

debian/copyright
denisonbarbosa: Oh, I see... I'll update it then.

docs/explanation/certificates.md
```diff
@@ -38,7 +38,8 @@ Set the enrollment method in `/etc/adsys.yaml`:
 certificate_enrollment: ldap    # or "cepces"
 ```
 
-* **Existing and new installations**: The default is `cepces` for backward compatibility. To use the native LDAP enrollment, add the setting above to your configuration file.
+* **New installations**: The package creates `/etc/adsys.yaml` with `certificate_enrollment: ldap` when no configuration file exists yet.
+* **Existing installations**: The default is `cepces` for backward compatibility. To switch to the native LDAP enrollment, add the setting above to your configuration file.
```
edibotopic, line 47: Query: is switching from cepces to ldap as simple as changing this setting in the config? Or are there also follow-up actions required of the user?

If so, it's not to be resolved here, but maybe in a migration guide.

edibotopic (approved):
One small Q. Approving in advance.

docs/explanation/certificates.md
denisonbarbosa: It should be as simple as that, yes. The daemon should automatically clean up any certificates enrolled by the other method and keep only the ones that come from the selected method.

edibotopic: Great: thanks for clarifying.

PR #1470, opened 2026-08-25
e2e: get the end-to-end suite running again

denisonbarbosa:
The end-to-end suite has not had a passing run in roughly a year. This gets both running again across noble, resolute, and stonking.

Example of a successful run after the fixes: https://github.com/ubuntu/adsys/actions/runs/33166934209

UDENG-11315


adombeck:
Do you want to review this PR, @nooreldeenmansour?

nooreldeenmansour:
> Do you want to review this PR, @nooreldeenmansour?

Sure, I wouldn't mind!

I'm not familiar with the e2e testing in adsys, but I believe this would be a great opportunity to get to know about it, and to dive a bit deeper in Adsys in general. :)

e2e/cmd/build_base_image/02_create_vm_template/main.go
nooreldeenmansour: `vmDeleted` is set before `az.DeleteVM` succeeds, I think it would be safer to do so only when we are sure `deleteVM` succeeded?

internal/policies/certificate/python/vendor_samba/gp/gp_cert_auto_enroll_ext.py
```diff
@@ -210,7 +210,7 @@ def get_supported_templates(server):
     try:
         # Server calls could hang if the provider is slow or too busy.
         # A timeout is set to avoid this.
-        out, err = p.communicate(timeout=3)
+        out, err = p.communicate(timeout=30)
```
nooreldeenmansour, line 213: The timeout exists only in the generated vendored file. A future vendor update (I see there is already a workflow and there is PR #1237) can replace it with the upstream `p.communicate()` implementation because `.github/samba/_patches/` does not carry the change.

The same refresh has already dropped the pre-existing `cepces.conf` override, so this is an observed failure mode rather than a theoretical risk. Should this change be carried in the patch series and then regenerated into the vendored file?

e2e/scripts/prepare-ad.ps1
```diff
@@ -66,6 +66,131 @@ $password = ConvertTo-SecureString -String 'supersecretpassword' -AsPlainText -F
...
+
+    $suffix = [System.IO.Path]::GetRandomFileName()
+    $infPath = Join-Path $env:TEMP "cep_renewal_${hostname}_${suffix}.inf"
+    $reqPath = Join-Path $env:TEMP "cep_renewal_${hostname}_${suffix}.req"
+    $crtPath = Join-Path $env:TEMP "cep_renewal_${hostname}_${suffix}.crt"
+    Set-Content -Path $infPath -Value $infContent -Encoding ASCII
+
+    certreq -q -f -new $infPath $reqPath
+    certreq -q -f -submit -config $caConfig -attrib "CertificateTemplate:DomainController" $reqPath $crtPath
+    certreq -q -f -accept $crtPath
```
nooreldeenmansour: Could we check `$LASTEXITCODE` after each `certreq` call? A failed renewal can leave the script succeeding without a valid certificate.

.github/actions/azure-sstpc-vpn/action.yaml
```diff
@@ -45,16 +74,36 @@ runs:
...
         CN=$(sudo openssl x509 -noout -subject -nameopt multiline -in /vpn-certificates/cert.pem | sed -n 's/ *commonName *= //p')
+        if [ -z "${CN}" ]; then
+            echo "::error::Could not determine the common name of the client certificate"
+            exit 1
+        fi
+
+        # The plugin lives in a directory named after the exact pppd version,
+        # so resolve it instead of hardcoding a version that breaks whenever
+        # the PPA ships a new one.
+        plugin="$(ls /usr/lib/pppd/*/sstp-pppd-plugin.so 2>/dev/null | head -1)"
```
nooreldeenmansour: Could we guard this command substitution? With `bash -e -o pipefail`, a failed `ls` (and the Python command at line 190) exits before the diagnostic below runs.

nooreldeenmansour (approved):
Apologies for the review delay 😅 

Great job on getting these tests running again! I've left two suggestions to consider, but I am approving in advance.

internal/policies/certificate/python/vendor_samba/gp/gp_cert_auto_enroll_ext.py
denisonbarbosa: Well spotted!

PR #1471, opened 2026-08-25
certificate: use GSSAPI seal outside TLS during bootstrap

denisonbarbosa:
## Description

Fresh certificate enrollment used StartTLS and then requested a GSSAPI confidentiality layer when the domain controller certificate was not trusted yet. Active Directory does not apply SASL security layers to an LDAP connection already protected by TLS, so the next plaintext LDAP response was interpreted as a SASL frame and enrollment failed.

Use StartTLS to determine whether the DC certificate is trusted, then reconnect to the same candidate without TLS when first-use bootstrap requires GSSAPI confidentiality. Kerberos mutual authentication and sealing protect the bootstrap exchange. Trusted connections continue to use StartTLS and channel binding.

Regression coverage verifies both transport paths.

## Tests

- `go test ./internal/policies/certificate -count=1`
- `go test -race ./internal/policies/certificate -count=1`
- `go test ./internal/policies/... -count=1`
- `golangci-lint run ./internal/policies/certificate/...`

The Azure/VPN-backed end-to-end suite was not run locally.

### DIFF
--- a/internal/policies/certificate/gssapi.go
+++ b/internal/policies/certificate/gssapi.go
@@ -225,12 +225,13 @@ func (g *gssapiClient) processAPREP(token []byte) ([]byte, bool, error) {
 //  1. Unwrap the server's GSSAPI wrap token to extract the 4-byte SASL payload
 //     describing supported security layers and max buffer size.
 //  2. Build a response selecting auth-only on a verified TLS connection, or
-//     confidentiality during first-use bootstrap.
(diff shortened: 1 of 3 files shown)


adombeck (approved):
Did you test this with a real AD instance?

denisonbarbosa:
> Did you test this with a real AD instance?

Yeah, this is one of the fixes needed to get the e2e tests green again.

PR #1475, opened 2026-08-31
ci: pin staticcheck to v0.7.0

denisonbarbosa:
Newer staticcheck versions [require Go >= 1.26](https://github.com/ubuntu/adsys/actions/runs/33355093037/job/99375596297#step:5:659). Since we are bound to Go 1.25 due to archive availability, we need to pin the dependency version to the latest one that still supports Go 1.25.


denisonbarbosa:
This is not required anymore after #1479

PR #1493, opened 2026-09-16
docs: update tutorial to no longer mention uneeded packagekit-tools

### LINKED ISSUE #1490: docs: current ubuntu (26.04) no longer requires packagekit-tools
current ubuntu (26.04) no longer requires packagekit-tools
The install line in the docs still mentions it: 
sudo apt install realmd sssd sssd-tools adcli samba-common samba-common-bin krb5-user packagekit-tools -y

"2026-01-28 - Matthias Klumpp <mak@debian.org>
packagekit (1.3.4-1) unstable; urgency=medium
* New upstream version: 1.3.4
* Drop packagekit-tools package
- The pkcon/pkmon tools are obsolete, and the replacement `pkgcli` tool
is available in the main package. The old split into a separate -tools
package was for historical reasons only.
- Nothing in Debian uses pkcon, and there is plenty of time to adjust any
scripts while this change sits in unstable/testing. But if there is a
huge, justifiable need to reintroduce pkcon, we can add the -tools
package back (please file a bug in that case). The `pkgcli` utility
should be an adequate replacement for both pkcon & pkmon.
* Update symbols file" from https://ubuntu.pkgs.org/26.04/ubuntu-main-amd64/packagekit_1.3.4-3_amd64.deb.html

---
*Reported from: https://ubuntu.com/docs/adsys/stable/tutorial/getting-started/*

denisonbarbosa:
The package packagekit-tools was never needed for adsys to work. It was a no-op in previous Ubuntu versions, but it breaks the tutorial after 26.04 because the package no longer exists in the archive, so let's remove it.

Closes #1490


OrazioLucioTerranova (approved):

edibotopic (approved):
Agreed. Thanks for the fix.

PR #1494, opened 2026-09-16
tests: Improve stability on CI

denisonbarbosa:
Some of our tests were quite flaky when running on GitHub CI. This PR should fix the issues.
More details about each fix are explained in each commit message.


adombeck:
Sorry, Copilot has more comments which should be addressed

PR #1502, opened 2026-09-21
e2e: stop using sstp-client from PPA

denisonbarbosa:
Now that ubuntu-latest resolves to Resolute rather than Noble, we no longer need to use the PPA to get the required ppp version (>= 2.5.0).


denisonbarbosa:
The ubuntu-latest label will migrate to Ubuntu 26 beginning October 19, 2026, so let's keep this as a draft until it gets migrated.

